PR #159, opened 2023-07-03
Speed is key

tht7:
# Checklist
- [ ] If this PR alters the UI, I have attached pictures/videos

# Pull Request Information

## About this Pull Request
Cached Image With Nsfw Filter.swift:
 - conditionally removed the ZStack (which stalls a bit for layout)
 - Replaced CachedAsyncImage with the native AsyncImage (the profiler was happy, but I'm not sure why)
 - Gave the image set size which prevents LazyVStack needing to relayout everything, and prevents jumping feed

FiltersTracker Saved Account Tracker:
 - moved the loading of files to during the init, so the initial value is the already loaded one

Community View:
 - Moved the post filtering to a background thread
 - Replaced the ZStack + offset (which causes re-layouting) with VStack + opacity + frame to hide/show the searchResultsView
 - added an id: to ForEach so he doesn't have to guess ContentView:
 - now pre-sets the main account onAppear (which should only happen once per window anyway) Feed Root:
 - now going into the default feed less

FeedTracker:
 - moved most of the functions to work on a background thread by default and only store the results using the main thread (add, merge and, deduped I'm looking at you)
 - added a way to attach a filtering fnuction to the dataset, that should take place the a background thread
 - preallocated an IDs set, so the allocator won't have to move it around as much

App State & Feed Root:
 - made sure the toasts don't mess up the SwiftUI state

## Screenshots and Videos
**In case this PR changes something in the UI, please include screenshots or videos of this new feature**

## Additional Context
**Any additional context you'd like to add to help us review this PR**


WestonHanners:
This PR is doing too much. That's why it's not getting approved. Can you cherry pick some of the individual commits into several more focused PR's?

tht7:
this PR is a lot of tiny changes to optimise all around, individually those changes seem like ineffective and "changes for the sake of changes", so it's kinda scary and time consuming,

But if you think that'll help I'll get to it (it will take a bit more time tho) 😞 

(but also looking at the changes in GitHub it looks like it displays a lot of changes that were "rebased" from master, and were not included in this PR originally so it leads me to believe I've done something to piss off git, even tho the app still builds and still receives the speed benefits when built locally)

EricBAndrews (approved):
Needs merge conflicts resolved but otherwise LGTM

westongpt (approved):

EricBAndrews:
At this point I'm not entirely sure which changes here have been cherry-picked, which are stale, and which are still viable--is it possible to nix this PR altogether, scrape together the changes that haven't been applied yet, and get them into a smaller PR against `dev`?

tht7:
This is totally stale, I’ll check what’s still relevant and pick this into tiny PRs to dev!

PR #203, opened 2023-07-04
fix #199 - ensure access token is not written to insecure disk location

### LINKED ISSUE #199: Why are we saving access tokens to JSON? Potential security issue?
Currently, Mlem saves account access tokens in a JSON file (`Saved Accounts.json` under Mlem's application support folder) in addition to in the keychain. Doesn't that defeat the whole point of using the keychain? It seems like a security issue to me.

It's pretty easy to just exclude the `accessToken` attribute from the `SavedAccount` codable, which would fix this issue. https://stackoverflow.com/questions/44655562/how-to-exclude-properties-from-swift-codable

mormaer:
# Checklist
      - #199

# Pull Request Information

## About this Pull Request
This addresses the concern raised in #199

As we're now storing the access token in the keychain (and retrieving it during start-up) we should not be storing it in JSON. This change ensures we write `"redacted"` into the file on disk instead of the real token.

## Additional context

The way we read/write to disk at start-up could do with some TLC, but for now this addresses the immediate concern.

### DIFF
--- a/Mlem/Models/Saved Account.swift
+++ b/Mlem/Models/Saved Account.swift
@@ -8,11 +8,17 @@
 import Foundation
 
 struct SavedAccount: Identifiable, Codable, Equatable, Hashable {
-    var id: Int
-
+    let id: Int
     let instanceLink: URL
-
-    var accessToken: String
-
+    let accessToken: String
     let username: String
+    
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        
+        try container.encode(self.id, forKey: .id)
+        try container.encode(self.instanceLink, forKey: .instanceLink)
+        try container.encode("redacted", forKey: .accessToken)
+        try container.encode(self.username, forKey: .username)
+    }
 }


EricBAndrews (approved):

WestonHanners (approved):

ShadowJonathan (commented):
Does this survive an application restart?

I'm not entirely knowledged on how encoding works, here, to me it looks like it just overwrites it with "redacted", which is then what the application reads on open.

tht7:
Do you think we can read the keychain from a background process? (like in the background app refresh if we might want to deliver notifications to the user, we will probably want to read the keychain and query the server)

mormaer:
> Does this survive an application restart?
> 
> I'm not entirely knowledged on how encoding works, here, to me it looks like it just overwrites it with "redacted", which is then what the application reads on open.

Sorry, didn't see the additional comments till I'd hit merge, the change ensures that the object we write out onto the disk in an insecure location will no longer store the token, when the application restarts we read that file back in and then use those to create a set of accounts with the tokens injected from the keychain.

You can see that bit happening here: https://github.com/mlemgroup/mlem/blob/master/Mlem/MlemApp.swift#L65

mormaer:
> Do you think we can read the keychain from a background process? (like in the background app refresh if we might want to deliver notifications to the user, we will probably want to read the keychain and query the server)

I'm not sure off the top of my head, but if the app is _alive_, even if in the background the accounts we have stored in memory will still have the token available, the change here is only to avoid writing the token to disk, as we were storing it in both the keychain _and_ as a plain text JSON in the application directory.

I think this was just an oversight when David had moved the tokens to the keychain, we'd never noticed it was still being written to disk in the same process 🙈

PR #222, opened 2023-07-06
Speed Pt.2 - Trackers Glory

tht7:
removed the annoying slow startup

# Checklist
- [ ] 
# The How
Made all the trackers load from disk on init, that way need need to publish less changes to swiftUI making their data available much earlier

Changed the startup sequence to load a community view much earlier

### DIFF
--- a/Mlem/MlemApp.swift
+++ b/Mlem/MlemApp.swift
@@ -6,6 +6,7 @@
 //
 
 import SwiftUI
+import UIKit
 
 @main
 struct MlemApp: App {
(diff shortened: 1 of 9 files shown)


EricBAndrews (approved):
Some prints in Filters Tracker can probably be removed but LGTM

Mlem/MlemApp.swift
```diff
@@ -21,76 +22,28 @@ struct MlemApp: App {
...
-        }
-        .onChange(of: accountsTracker.savedAccounts) { newValue in
-            do {
-                let encodedSavedAccounts: Data = try encodeForSaving(object: newValue)
 
-                do {
-                    try writeDataToFile(data: encodedSavedAccounts, fileURL: AppConstants.savedAccountsFilePath)
-                } catch let writingError {
-                    print("Failed while saving data to file: \(writingError)")
+                    let appearance = UITabBarAppearance()
```
mormaer, line 32: we should try and avoid use of `UIKit` like this since we're working in `SwiftUI`, what is the reason we need to add this?

Mlem/Models/Trackers/Favorite Community Tracker.swift
```diff
@@ -10,7 +10,29 @@ import Foundation
...
+                ) as? [FavoriteCommunity] ?? []
+            } catch let favoriteCommunitiesDecodingError {
+//                print("Failed while decoding favorite communities: \(favoriteCommunitiesDecodingError)")
+            }
+        } else {
+//            print("Favorite communities file does not exist, will try to create it")
+            do {
+                try createEmptyFile(at: AppConstants.favoriteCommunitiesFilePath)
+            } catch let emptyFileCreationError {
+//                print("Failed while creating empty file: \(emptyFileCreationError)")
```
mormaer, line 33: can we remove the commented out lines and replace them with `TODO`'s to add error handling / remove the named error variable to silence the warnings?

Mlem/Views/Tabs/Feed Root.swift
```diff
@@ -46,30 +47,21 @@ struct FeedRoot: View {
...
-                        rootDetails = CommunityLinkWithContext(community: nil, feedType: feedType)
-                        
-                        shortcutItemToProcess = nil
-                    }
-                }
-            }
-        }
+        .handleLemmyLinkResolution(
+            navigationPath: $navigationPath,
+            local: "Inside root feed"
```
mormaer, line 60: given this is just for debugging, can we remove the argument?

Mlem/Views/Tabs/Feed Root.swift
```diff
@@ -79,14 +71,22 @@ struct FeedRoot: View {
...
-            if appState.currentActiveAccount == nil, let account = accountsTracker.savedAccounts.first {
-                appState.currentActiveAccount = account
+            if rootDetails == nil || shortcutItemToProcess != nil {
+                let feedType = FeedType(rawValue:
+                    shortcutItemToProcess?.type ??
+                    "nothing to see here"
+                ) ?? defaultFeed
+                var dits: CommunityLinkWithContext?
+                if appState.currentActiveAccount != nil {
+                    dits = CommunityLinkWithContext(community: nil, feedType: feedType)
```
mormaer, line 86: what is `dits`? is there a more descriptive name this variable could have?

Mlem/Views/Tabs/Feed Root.swift
```diff
@@ -95,24 +95,24 @@ struct FeedRoot: View {
...
-            
-            DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
+
+            DispatchQueue.main.asyncAfter(deadline: .now()) {
                 if rootDetails == nil {
                     rootDetails = CommunityLinkWithContext(community: nil, feedType: defaultFeed)
                 }
-                _ = HandleLemmyLinkResolution(appState: _appState,
-                                          savedAccounts: _accountsTracker,
+//                didReceiveURL(url)
```
mormaer, line 114: should we remove this rather than comment out?

mormaer (changes requested):

Mlem/MlemApp.swift
tht7: A weird bag that appears only on Pro Max phones where the tab view becomes fully transparent 

This tells the bar to keep the systemThinMaterial all the time (which looks like the default in iOS anyway)

WestonHanners: Is there a native SwiftUI way to do this?

Mlem/Models/Trackers/Favorite Community Tracker.swift
tht7: Aye aye cap

Mlem/MlemApp.swift
mormaer: When I navigate back from Safari my tab bar is transparent for a moment (iPhone 13 Pro), that never used to happen - is there something we're doing that's causing this bug to surface?

mormaer: Example here:

https://github.com/mlemgroup/mlem/assets/5231793/d47e7ae2-729d-487f-9de9-daff1aaffecb

Mlem/Views/Tabs/Feed Root.swift
tht7: Addressed

Mlem/MlemApp.swift
tht7: Actually, if you notice the way we open, Safari is very dodgy, (there's some UIKit window manipulation)

But other than this, I have no idea what we might be doing to surface this,

About swiftUI to do it I'm not sure the swiftUI method applies as thoroughly as this one

mormaer: Yeah, I had changed that bit to stay within `SwiftUI` using a representable but David wasn't keen on it, can we raise a separate issue for this bug, perhaps if we go back to using a different approach for Safari it will alleviate it?

If we note it as an issue we can resolve it later, I don't think it's a blocker to merging

WestonHanners (approved):

Mlem/MlemApp.swift
mormaer: raised as an issue here: https://github.com/mlemgroup/mlem/issues/231

mormaer (approved):

PR #223, opened 2023-07-06
Optimized embedded avatar images to be the needed size

JakeShirley:
# Checklist
      - n/a

# Pull Request Information

## About this Pull Request
This PR updates profile and community avatars to use 32px or 64px avatars depending on their rendered size.

The result is much smoother scrolling with slight degradation of icons visually.

## Screenshots and Videos
n/a

## Additional Context
Probably could play with the format of the image as well (png, etc.)

### DIFF
--- a/Mlem/Views/Shared/Links/Community Link View.swift
+++ b/Mlem/Views/Shared/Links/Community Link View.swift
@@ -135,11 +135,15 @@ struct CommunityLabel: View {
         serverInstanceLocation == .bottom ? AppConstants.largeAvatarSize : AppConstants.smallAvatarSize
     }
     
+    private func avatarUrl(from: URL) -> URL {
+        serverInstanceLocation == .bottom ? from.withIcon64Parameters : from.withIcon32Parameters
+    }
+    
     @ViewBuilder
     private var communityAvatar: some View {
         Group {
             if let communityAvatarLink = community.icon {
-                CachedAsyncImage(url: communityAvatarLink, urlCache: AppConstants.urlCache) { image in
+                CachedAsyncImage(url: avatarUrl(from: communityAvatarLink), urlCache: AppConstants.urlCache) { image in
                     if let avatar = image.image {
                         avatar
                             .resizable()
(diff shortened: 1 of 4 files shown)


Mlem/Views/Shared/Links/Community Link View.swift
```diff
@@ -135,11 +135,15 @@ struct CommunityLabel: View {
         serverInstanceLocation == .bottom ? AppConstants.largeAvatarSize : AppConstants.smallAvatarSize
     }
     
+    private func avatarUrl(from: URL) -> URL {
```
mormaer, line 138: nit

Can we do;
```swift
avatarUrl(from url: URL) -> URL {
    serverInstanceLocation == .bottom ? url.withIcon64Parameters : url.withIcon32Parameters
}
```

it lets the function use `url` internally for the argument name instead of `from`

Mlem/Views/Shared/Links/User Profile Label.swift
```diff
@@ -99,6 +99,10 @@ struct UserProfileLabel: View {
         serverInstanceLocation == .bottom ? AppConstants.largeAvatarSize : AppConstants.smallAvatarSize
     }
     
+    private func avatarUrl(from: URL) -> URL {
+        serverInstanceLocation == .bottom ? from.withIcon64Parameters : from.withIcon32Parameters
```
mormaer, line 103: nit

same as previous comment ^

mormaer (approved):
🎉 - one little nit but looks good

Mlem/Views/Shared/Links/Community Link View.swift
EricBAndrews: What's the significance of the `from` keyword in that case? I've seen it around but don't get what it does

EricBAndrews (approved):

Mlem/Views/Shared/Links/Community Link View.swift
mormaer: It's just semantics I guess, you can have a public name and an internal name, in the case of the function here it lets you write at the call site:

`avatarUrl(from: ...)`

but then inside the function you can refer to the argument as `url` which makes sense, rather than having to call it `from` which doesn't really give a clue as to what it is

PR #232, opened 2023-07-07
Update keychain

mormaer:
# Checklist
      - a small change related to chats had earlier in the development channel

# Pull Request Information

## About this Pull Request
This PR updates our keychain service name to match the application bundle identifier

## Additional Context

Note that this change will mean **ALL USERS** will need to re-add their accounts as the keychain has changed.

### DIFF
--- a/Mlem/App Constants.swift
+++ b/Mlem/App Constants.swift
@@ -19,7 +19,7 @@ struct AppConstants {
     static let dateComponentsFormatter: DateComponentsFormatter = DateComponentsFormatter()
 
     // MARK: - Keychain
-    static let keychain: Keychain = Keychain(service: "com.davidbures.Mlem-keychain")
+    static let keychain: Keychain = Keychain(service: "com.hanners.Mlem-keychain")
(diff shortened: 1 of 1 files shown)


mormaer:
I'm open to suggestions on what to name that new toggle and what wording to use, I've defaulted to `off` so that users have to opt-in as is best practice.

mormaer:
I've updated this today so that we will always store the token we receive from the initial log in. This matches the current behaviour and means that accounts once added will always remain in the accounts list, if a user chooses not to save their credentials then when it comes to refreshing the session we will need to prompt them.

The flow for that is currently not built, but until Lemmy change their token logic we _should_ be OK for now 🙈

mormaer:
After discussion with @tht7 this PR now only updates our keychain service name to match the bundle, credentials are no longer stored

WestonHanners (approved):

WestonHanners (approved):

EricBAndrews (approved):

Issue #102, opened 2023-06-10
Button to jump to the root of the next comment


ShadowJonathan:
Apollo had this as well, where a button (pointing "down") on the bottom-right edge of the screen would scroll the screen immediately to the next comment root, basically skipping the entire current replies section and going to the next.

PR #238, opened 2023-07-07
Removal of legacy community search dropdown

JakeShirley:
# Checklist
      - n/a

# Pull Request Information

## About this Pull Request
With the addition of the explicit search tab, this PR removes the old community search bar at the top of the community view.  It now just says "All", "Subscribed", "Local" with a dropdown to swap between the three, or the community name.

Also cleaned up post links requiring a feed type for some reason.

## Screenshots and Videos

https://github.com/mlemgroup/mlem/assets/1000311/daaf2230-e8a2-40a6-bbde-39d76e346246

## Additional Context
n/a


EricBAndrews (approved):
🧹

d42ohpaz:
I would like to say that's a bit of a shame if switching between Subscribed and All means having to pop back a page to select the feed type. Being able to quickly pull down and select the type, without having to leave the current feed, was actually nice.

JakeShirley:
@d42ohpaz I think that's actually great feedaback.  We debated it a bit internally and we thought it was probably a more consistent and purpose-built experience to back off to the community list but if people are actually using this experience that is good to know.

mormaer:
I think it should stay, at least the access to switching between the feed types.

The search option should go if it's now been given it's own tab 👍 

This is also the only place favourited communities were listed, so they should likely stay for now, or we need to find a new home for those too?

Sjmarf:
I'm not a fan of the old search bar implementation - it feels janky, and the position at the top of the screen makes the controls difficult to reach. However, I do think that it provides useful functionality. 

In my opinion, we should keep it for now and look into alternatives. I'll do some thinking

fwcd:
I liked the old dropdown too. Sure, it's a bit redundant with the addition of the search tab, but it's still a great shortcut for switching communities and keeping it IMO probably wouldn't hurt.

Sjmarf:
Here's something I tried - there could be a small button in the bottom-right, which opens the communities view in a half-height sheet at the bottom of the screen. This is both cleaner and more reachable than the current solution imo.

https://github.com/mlemgroup/mlem/assets/78750526/904cce9d-7020-4178-8a6d-25c2ca7950e9

d42ohpaz:
The one concern I have with @Sjmarf's solution - which, in and of itself, is pretty slick - is that the icon covers a portion of the feed. At least having it at the top, as part of the navigation bar, it does not impede the interface.

Sjmarf:
> The one concern I have with @Sjmarf's solution - which, in and of itself, is pretty slick - is that the icon covers a portion of the feed. At least having it at the top, as part of the navigation bar, it does not impede the interface.

Indeed, that is something to consider. At some point, we plan to have a button in the bottom-right of a post View to jump to the next comment, as implemented in the official app and Apollo (#102). In both apps, I didn't find myself bothered by the comment-jump button - however, I can't say for sure whether I'd like it to exist on the main feed though. Maybe I'll build it on my device later and see.

 Potentially, we could reduce the opacity of the button whilst the user is scrolling, or hide it altogether.

WestonHanners (approved):
Anything to reduce the complexity of CommunityView.

WestonHanners:
I recommend we remove this for 1.0 and give it a more thoughtful design for 1.1 or 1.2.

d42ohpaz:
> I recommend we merge this for 1.0 and give it a more thoughtful design for 1.1 or 1.2.

From an end-user's perspective, it would be better to leave it and refine it, than to take it away only to bring it back (even if refined).

EricBAndrews:
I agree that we should remove it for now and revisit post 1.0. We've got some larger UX questions to consider about basic navigation (instances and accounts)--we should keep our UI options as open as we can until we have that flow worked out.

JakeShirley:
Chatting on Matrix, we decided to keep in quick switching between subscribed, all, and local but remove search.

PR #240, opened 2023-07-08
Nuke image Preloading

tht7:
# Checklist

# Pull Request Information
This PR replaces our old image system with the Nuke library!
Nuke does things like decode images in the background, and in a very multithreaded way,
Makes scrolling the feed silky smooth

### DIFF
--- a/Mlem/MlemApp.swift
+++ b/Mlem/MlemApp.swift
@@ -22,10 +22,10 @@ struct MlemApp: App {
             Window()
                 .environmentObject(accountsTracker)
                 .onAppear {
-                    var imageConfig = ImagePipeline.Configuration.withDataCache(name: "main", sizeLimit: 500_000)
-                    imageConfig.dataLoadingQueue = OperationQueue(maxConcurrentCount: 6)
-                    imageConfig.imageDecodingQueue = OperationQueue(maxConcurrentCount: 6) // Let's use those CORES
-                    imageConfig.imageDecompressingQueue = OperationQueue(maxConcurrentCount: 6)
+                    var imageConfig = ImagePipeline.Configuration.withDataCache(name: "main", sizeLimit: AppConstants.cacheSize)
+                    imageConfig.dataLoadingQueue = OperationQueue(maxConcurrentCount: 8)
+                    imageConfig.imageDecodingQueue = OperationQueue(maxConcurrentCount: 8) // Let's use those CORES
+                    imageConfig.imageDecompressingQueue = OperationQueue(maxConcurrentCount: 8)
                     
                     // I'm leaving that here for mormaer, once I get a handle on rate limites that's where we put em!
                     // imageConfig.isRateLimiterEnabled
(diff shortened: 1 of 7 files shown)


Mlem/MlemApp.swift
```diff
@@ -21,6 +22,15 @@ struct MlemApp: App {
             Window()
                 .environmentObject(accountsTracker)
                 .onAppear {
+                    var imageConfig = ImagePipeline.Configuration.withDataCache(name: "main", sizeLimit: AppConstants.cacheSize)
```
mormaer, line 25: looking at the responses we get from images from Lemmy they contain `cache-control` values, and according to the documentation for this library the default caching mechanism supports that, here we're changing it to use the type of caching that ignores those headers, do we need to?

Mlem/Models/Trackers/Post Tracker.swift
```diff
@@ -7,9 +7,14 @@
 
 import Foundation
 import SwiftUI
+import Nuke
 
 class PostTracker: FeedTracker<APIPostView> {
 
+    private let prefetcher = ImagePrefetcher(pipeline: ImagePipeline.shared,
+                                             destination: .memoryCache,
+                                             maxConcurrentRequestCount: 40)
```
mormaer, line 16: the libraries default here is `2`, is `40` perhaps overkill?

mormaer:
- if we're going to bring a library in for handling images/pre-fetching we should likely also do the work to move all this behaviour into a separate object/class as part of the same work. At the moment having the pre-fetching inside the feed loader feels wrong and I always assumed that was a temporary thing, but we now seem to be building more and more without addressing where it lives... it shouldn't be in there in my opinion 🤷 

- it looks like there are some UI changes introduced by this PR which are not shown in the description - it'd be worth showing some examples of what failed images would look like for review

- similar to the first point, we should probably consider how a change like this affects the wider project, we now have one view (the image with NSFW) using the new `NukeUI` and elsewhere in the project we're still using `CachedAsyncImage` directly... I'd prefer to see changes like this be made in a way that brings consistency across the application as opposed to adding more confusion as to which approach should be used 😞

Mlem/MlemApp.swift
tht7: I didn’t know lemmy returns cache-control, so I choose a cache type that ignores it, 

it makes me feel safer if we’d control those images and not those servers (considering how the api is built) but if you’d like I can change it

Mlem/Models/Trackers/Post Tracker.swift
tht7: Indeed, but it’s very very carefully tuned, without that the feed gets jumpy as the network catches up 

But with preloading so much in advance and in parallel it’s smooth as butter

EricBAndrews:
I agree with @mormaer about the UI changes--I'd like to see how failed and loading images display, then we're clear to merge. Opened some issues to clean up the mess in 1.1

Issue #207, opened 2023-07-05
Ensure we handle invalid tokens


mormaer:
## Introduction
As of now the application stores the users authentication token after successful login. This is then re-used in future sessions for accessing the API.

Although so far the tokens appear to be long-lived, there will/should be scenarios where the token becomes invalid. At this point we currently have no way to notify the user that the session is invalid and/or offer to log in again.

## Requirements
If the API was to return an unauthorised error when we attempt to use the token we have stored, we should gracefully handle that scenario by attempting to silently login again and/or prompt the user.

mormaer:
I was thinking about this, and I believe in order to correctly handle this scenario when it comes around we will need to make some changes to how we currently handle accounts.

At present after login we store the token in the keychain, but should the session be invalidated by the server (or the user if an option such as _sign out from all sessions_ becomes available via Lemmy) we have no way to recover, other than to ask the user to sign in again from scratch.

A more robust approach would be for us to securely store the users credentials in the keychain so that we can make a login call as required to create a new session, as these changes would likely mean existing accounts would need replaced (the model would need to change) I believe this is something that should be addressed ASAP.

d42ohpaz:
FYI, there are no errors or notices that you are logged out when a session becomes invalid. Instead, the only noticeable things that happens is your Subscribed feed is empty, and you cannot interact (i.e., upvote or downvote). It's confusing, and can easily be dismissed as server issues when in fact it's not a server issue.

PR #250, opened 2023-07-09
Unify error handling

mormaer:
# Checklist
      - Some initial steps toward a final resolution of #207

# Pull Request Information

## About this Pull Request
This PR does some preliminary work to make solving #207 possible.

- Introduces a new `ContextualError` type which can be used throughout the app to represent an error
- Updates our `ContentView` to receive and handle this new error type to allow default handling behaviours
- Updates (most of*) the application to unify the way errors are handled to use the above

With those steps done we can now detect invalid sessions regardless of where in the journey it occurs, the next PR will add the required behaviour to present a login flow and refresh the users session.

* I've left most of the persistence layer printing out it's various errors... it needs burned with 🔥 and re-written as it's hideous and makes me cry when I see it 🙈 

## Screenshots and Videos

An example of an expired session being caught at various points in the user flow:

https://github.com/mlemgroup/mlem/assets/5231793/e2548981-aacf-4984-957c-dcea250d956a

## Additional Context
As a small bonus since (most) errors now flow through a single route I put in some VERY PRIMITIVE logging, this also only happens in DEBUG builds to avoid spamming out in production.

a_step_toward_logging

We can now expand this further to implement and actual logging system should we wish.


Mlem/API/APIClient.swift
```diff
@@ -50,6 +51,11 @@ class APIClient {
             // the API, however we may way to consider adding the error model type as an
             // associated value in the same was as the response to allow requests to define
             // their own error models when necessary, or drop back to this as the default...
+            
+            if apiError.error == "not_logged_in" {
```
mormaer, line 55: given we get a `400` when it should likely be a `401` I'm just looking at the error string itself, which is disgusting but it's the best the API gives us as it stands... we'll need to keep an eye on this in future Lemmy releases in case they decide to change this messaging.

Mlem/Extensions/View - Handle Lemmy Links.swift
```diff
@@ -188,7 +188,7 @@ struct HandleLemmyLinkResolution: ViewModifier {
                             // so I'll skip and let the system open it instead
                         // }
                     } catch {
-                        print(String(describing: error))
+                        appState.contextualError = .init(underlyingError: error)
```
mormaer, line 191: an example of using the new `.contextualError` where we don't want to communicate anything to the user, so we only pass in the error itself, the new handling will look at that error and if it's a session issue, or general API error we will display it

Mlem/Views/Tabs/Posts/Components/Post/Components/Community View/Community View.swift
```diff
@@ -102,12 +101,14 @@ struct CommunityView: View {
...
-                            appState.alertTitle = "Could not load community information"
-                            appState.alertMessage = "The server might be overloaded.\nTry again later."
-                            appState.isShowingAlert.toggle()
+                        } catch {
+                            print("Failed while fetching community details: \(error)")
+                            
+                            appState.contextualError = .init(
+                                title: "Could not load community information",
+                                message: "The server might be overloaded.\nTry again later.",
+                                underlyingError: error
```
mormaer, line 110: this is an example where we would want to communicate the error to the user, so we pass in some contextual information (title/message), the new error handling will look at the `underlyingError` first, if it's something wee have default handling for (like a session issue) it'll be handled, if not - an alert will be shown using the information we've fed in

WestonHanners (approved):

JakeShirley (approved):

PR #259, opened 2023-07-11
Implemented User Blocking

WestonHanners:
# Checklist
- [x] This PR addresses one or more open issues that were assigned to me: (Linked in PR)

# Pull Request Information

## About this Pull Request
Implement Block User API
Add Block User Button
Block User Toast
Update Post Tracker

## Screenshots and Videos
image

### DIFF
--- a/Mlem/Logic/Participating/BlockPersonLogic.swift
+++ b/Mlem/Logic/Participating/BlockPersonLogic.swift
@@ -1,0 +1,26 @@
+//
+//  BlockPerson.swift
+//  Mlem
+//
+//  Created by Weston Hanners on 7/10/23.
+//
+
+import Foundation
+
+@MainActor
+func blockPerson(
+    account: SavedAccount,
+    person: APIPerson,
+    blocked: Bool) async throws -> Bool {
+        do {
+            let request = BlockPersonRequest(
+                account: account,
+                personId: person.id,
+                block: blocked
+            )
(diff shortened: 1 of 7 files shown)


tht7 (approved):

Mlem/Logic/Participating/BlockPersonLogic.swift
```diff
@@ -0,0 +1,26 @@
...
+        do {
+            let request = BlockPersonRequest(
+                account: account,
+                personId: person.id,
+                block: blocked
+            )
+            let response = try await APIClient().perform(request: request)
+            return response.blocked
+        } catch {
+            return false
```
mormaer, line 24: this function is declared `throws`, but we're catching all errors here and returning `false` so no throw will ever happen, it looks like the logic in `FeedPost` is written to show an error toast inside its `catch` but I don't think that will work if this function catches them all?

should we remove the `do/catch` from this function to let errors throw and be handled in the feed post?

mormaer (changes requested):

Mlem/Logic/Participating/BlockPersonLogic.swift
mormaer: I'll look at adjusting the new contextual error thing to let us choose between alerts and toasts as the way to notify, so we can get the default behaviours without always showing alerts 👍

mormaer (approved):
🚀

PR #301, opened 2023-07-15
Middleware exploration

mormaer:
# Pull Request Information

## About this Pull Request
This PR introduces the concept of a repository pattern into the application in order to separate the view layer from the network layer. At present only the calls relating to comments have been moved into a new `CommentRepository`.

The models which are returned from the repository are still the same as were returned by the raw network layer, we can now choose to return enriched/scoped models where relevant in future PRs. The aim of this PR was to add the layer in as a template to agree on so that we can begin introducing this concept across the application.

In addition to the introduction of the `CommentRepository` I have tidied up the error handling a bit further.

When encountering an error you can pass it to `ErrorHandler.handle(...)` as before which will look to provide default handling for expired tokens and API errors. When you pass the error into the handler you have three main options;
```swift
 // will check for invalid session / API error, if neither are present the error will be silenced
.init(underlyingError: error)

 // will check for invalid session / API error if neither are present an alert will be displayed
.init(title: "Some title", message: "Some message", underlyingError: error)

 // will check for invalid session / API error if neither are present an error toast will be displayed
.init(title: "Some title", style: .toast, underlyingError: error)
```

I would like to make the above API a bit cleaner but right now the `ContextualError` type must conform to `Equatable` as it is a property that is used via `AppState` from the rest of the application, as we introduce the repositories across the application that approach will be refactored out and allow for a slightly nicer approach which I have up my sleeve 😉

Lastly, there is some `UIKit` jiggery-pokery being done for the error handling which ensures that wherever we encounter our error and pass it to the handler the appropriate modal/alert/toast will be shown to the user... before now there was an issue where using the handler would fail to present errors if you called it from inside a presented context (eg, a sheet). I tried, believe me, I tried to solve that in `SwiftUI` but it was becoming so hideously ugly that I gave up.

Rest assured, that defeat weighs heavy on my soul and if/when I figure out a clean way to do it in `SwiftUI` the PR shall arrive henceforth with speed.

---

Note: This PR will introduce a third party library [swift-dependencies](https://github.com/pointfreeco/swift-dependencies) to help control the dependencies throughout the project. We can achieve something similar using the environment values that SwiftUI offers, however there are additional niceties this offers when considering testing which I believe makes it a worthwhile addition.

### DIFF
--- a/Mlem.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Mlem.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -14,8 +14,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/pointfreeco/combine-schedulers",
       "state" : {
-        "revision" : "0625932976b3ae23949f6b816d13bd97f3b40b7c",
-        "version" : "0.10.0"
+        "revision" : "ec62f32d21584214a4b27c8cee2b2ad70ab2c38a",
+        "version" : "0.11.0"
       }
     },
     {
@@ -50,8 +50,17 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/pointfreeco/swift-clocks",
       "state" : {
-        "revision" : "f9acfa1a45f4483fe0f2c434a74e6f68f865d12d",
-        "version" : "0.3.0"
+        "revision" : "0fbaebfc013715dab44d715a4d350ba37f297e4d",
+        "version" : "0.4.0"
+      }
+    },
+    {
+      "identity" : "swift-concurrency-extras",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/pointfreeco/swift-concurrency-extras",
+      "state" : {
+        "revision" : "479750bd98fac2e813fffcf2af0728b5b0085795",
+        "version" : "0.1.1"
       }
     },
     {
@@ -59,8 +68,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/pointfreeco/swift-dependencies",
       "state" : {
-        "revision" : "de1a984a71e51f6e488e98ce3652035563eb8acb",
-        "version" : "0.5.1"
+        "revision" : "16fd42ae04c6e7f74a6a86395d04722c641cccee",
+        "version" : "0.6.0"
       }
     },
     {
@@ -86,8 +95,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/pointfreeco/xctest-dynamic-overlay",
       "state" : {
-        "revision" : "4af50b38daf0037cfbab15514a241224c3f62f98",
-        "version" : "0.8.5"
+        "revision" : "245d527002f29c5a616c5b7131f6a50ac7f41cbf",
+        "version" : "0.8.6"
       }
     }
   ],


Mlem/Error Handling/ErrorHandler.swift
```diff
@@ -0,0 +1,60 @@
...
+    func handle(_ error: ContextualError?) {
+        guard let error else {
+            return
+        }
+        
+        #if DEBUG
+        log(error)
+        #endif
+
+        Task {
```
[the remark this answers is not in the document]
ckaik: I might have a very funky idea for that. Keep a stack of handlers in the `ErrorHandler` object and pass that as an `EnvironmentObject` to all presented sheets, then when an error happens always use the topmost handler.

To update that stack you would need to push/pop on appear/disappear respectively.

I've tried it - quick and dirty - and it seems to _mostly_ work, if you uncomment the `task` and push a couple of sheets it doesn't seem to be using the top one for some reason.

Most of the Boilerplate could be moved into a `ViewModifier` I believe and you wouldn't have to add much when presenting any sheet.

```swift
typealias Handler = () -> Void

struct RootView: View {
    @State private var error = false
    @StateObject private var errorManager = ErrorManager()

    var body: some View {
        SheetView()
            .environmentObject(errorManager)
            .addingErrorHandler(errorManager) {
                error = true
            }
            .sheet(isPresented: $error) {
                Text("Root")
            }
//            .task {
//                try? await Task.sleep(for: .seconds(3))
//
//                errorManager.error()
//            }
    }
}

class ErrorManager: ObservableObject {
    private var handlers = [Handler]()

    func push(_ handler: @escaping Handler) {
        handlers.append(handler)
    }

    func pop() {
        handlers.removeLast()
    }

    func error() {
        handlers.last?()
    }
}

extension View {
    func addingErrorHandler(_ errorManager: ErrorManager, handler: @escaping Handler) -> some View {
        onAppear {
            errorManager.push(handler)
        }
        .onDisappear {
            errorManager.pop()
        }
    }
}

struct SheetView: View {
    let id = UUID()

    @State private var sheet = false
    @State private var error = false

    @EnvironmentObject private var errorManager: ErrorManager

    var body: some View {
        VStack {
            Text(id.uuidString)

            Button("Show Sheet") {
                sheet = true
            }

            Button("Show Error") {
                error = true
            }
        }
        .buttonStyle(.borderedProminent)
        .sheet(isPresented: $error) {
            VStack {
                Text("ERROR")
                Text(id.uuidString)
            }
        }
        .sheet(isPresented: $sheet) {
            SheetView()
                .addingErrorHandler(errorManager) {
                    error = true
                }
        }
    }
}
```

Mlem/API/APIClient.swift
```diff
@@ -102,3 +130,100 @@ class APIClient {
         }
     }
 }
+
+// MARK: Comment Requests
+
+extension APIClient {
+    func loadComments(
+        for postId: Int,
+        maxDepth: Int = 15,
```
WestonHanners, line 139: Thanks for all the defaults

Mlem/Views/Tabs/Posts/Components/Comment/Components/Comment Item.swift
```diff
@@ -153,9 +157,6 @@ struct CommentItem: View {
         .clipped()
         .padding(.leading, depth == 0 ? 0 : indent)
         .transition(.move(edge: .top).combined(with: .opacity))
-        .alert("Not yet implemented!", isPresented: $isShowingAlert) {
-            Button("I love beta apps", role: .cancel) { }
-        }
```
EricBAndrews, line 158: 😢 goodnight sweet prince

EricBAndrews (approved):
Very nice! Please wait for #339 to merge because `dev` is broken rn

Mlem/Views/Tabs/Posts/Components/Comment/Components/Comment Item.swift
WestonHanners: Will not be missed at ALL

Mlem/API/APIClient.swift
```diff
@@ -102,3 +130,100 @@ class APIClient {
...
+            content: content,
+            languageId: languageId,
+            parentId: parentId,
+            postId: postId
+        )
+        
+        return try await perform(request: request)
+    }
+    
+    func applyCommentScore(id: Int, score: Int) async throws -> CommentResponse {
```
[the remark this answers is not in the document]
EricBAndrews: Ah, yes, that is how it should work. Disregard that comment.

PR #302, opened 2023-07-15
Option to Fully Hide NSFW

EricBAndrews:
To ensure we comply *fully* with App Store guidelines, I've added the option to entirely remove NSFW content from the feed (and enabled it by default). I've also moved the Adult Content section of the Privacy Policy to the EULA, since it's about the user's responsibility to interact with the app safely/accept that there might be the odd piece of untagged NSFW and not about privacy at all.

Also fixed some initial toggle states.


WestonHanners:
I think the actual appstore rules are that the SERVER has to have the option and we must respect it. This is why Apollo removed the toggle from the app. I don't think we need this.

PR #331, opened 2023-07-20
added duplicate avoidance, logic for deleting the last account

EricBAndrews:
Little PR to fix up some logic around the account switcher. Deleting the active account is now possible; if another account is available it swaps you to it, otherwise it chucks you out to the onboarding flow.

Added an environment function to force into the onboarding flow.

### DIFF
--- a/Mlem/Models/Trackers/Saved Account Tracker.swift
+++ b/Mlem/Models/Trackers/Saved Account Tracker.swift
@@ -40,13 +40,8 @@ class SavedAccountTracker: ObservableObject {
     
     // TODO: pass in AppState using a dependency or something nice like that
     func removeAccount(account: SavedAccount, appState: AppState, forceOnboard: () -> Void) {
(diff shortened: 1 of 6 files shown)


WestonHanners (approved):

Mlem/Models/Trackers/Saved Account Tracker.swift
```diff
@@ -29,12 +29,18 @@ class SavedAccountTracker: ObservableObject {
     }
     
     func addAccount(account: SavedAccount) {
+        // prevent dupes
+        guard !savedAccounts.contains(account) else {
+            return
```
WestonHanners, line 34: actually, do we need a message returned here?

EricBAndrews: I don't think so--it just quietly succeeds, user doesn't need to know they logged into an account they already had lol

PR #343, opened 2023-07-20
Onboarding Bug fix

Sjmarf:
Fixed a bug with onboarding that was introduced in my last PR. Oops 😅


EricBAndrews:
Cleaner fix is simply using `AddSavedInstanceView` at the `Window` level--that's implemented in #345, so I'm closing this one

Issue #134, opened 2023-06-09
Editing of posts and comments


buresdv:
## Introduction
I often make typos and forget to write words. It would be great to have the option to edit your comments and posts

Issue #36, opened 2023-06-26
Edit posts


EricBAndrews:
Users should be able to edit posts. This requires the new editing interface.

Child issue of mlemgroup/mlem#65

Sjmarf:
#363 

I've done some prep for this feature. If someone wants to implement this, have at it

PR #363, opened 2023-07-23
Edit Post Prep

Sjmarf:
Backend prep for the editing posts feature. No functionality is changed in this PR. I would have done more, but I'll be going away for a week so I thought I'd submit what I've done so far - if someone else wants to finish this feature, have at it.

I've separated the editing interface from `PostComposerView` into a separate `PostDetailEditorView`. I imagine this new `PostDetailEditorView` could be used in a `PostEditorView` sheet similar to `PostComposerView` but for editing the post. I've also implemented the `editPost` function to perform the post edit. 

Related: #36, #134

### DIFF
--- a/Mlem/Logic/Participating/Edit Post.swift
+++ b/Mlem/Logic/Participating/Edit Post.swift
@@ -21,7 +21,7 @@ import SwiftUI
             account: account,
             postId: postId,
             name: postTitle,
-            url: URL(string: postURL!),
+            url: postURL,
             body: postBody,
             nsfw: postIsNSFW,
             languageId: nil
--- a/Mlem/Logic/Participating/Edit Post.swift
+++ b/Mlem/Logic/Participating/Edit Post.swift
@@ -1,0 +1,39 @@
+//
+//  Edit Post.swift
+//  Mlem
+//
+//  Created by Sjmarf on 23/07/2023.
+//
+
+import SwiftUI
+
+// swiftlint:disable function_parameter_count
+@discardableResult func editPost(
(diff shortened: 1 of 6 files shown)


Mlem/Logic/Participating/Edit Post.swift
```diff
@@ -0,0 +1,39 @@
...
+    postURL: String?,
+    postIsNSFW: Bool,
+    postTracker: PostTracker,
+    
+    account: SavedAccount) async throws -> APIPostView {
+        let request = EditPostRequest(
+            account: account,
+            postId: postId,
+            name: postTitle,
+            url: URL(string: postURL!),
```
EricBAndrews, line 24: For consistency, url shouldn't be force unwrapped and instead given a default value of `""` down in EditPostRequest--this is how we handle this problem in CreatePostRequest, which isn't ideal but we should at least be consistent about it until we get a more robust (probably middleware) solution in place

EricBAndrews (changes requested):

EricBAndrews:
I'm about to start work on editing comments, as part of that I'm moving the `ResponseEditor` sheet up to the top level, where it can be popped by calling an `EnvironmentObject`. Long-term I'd like to get all the editing functionality unified there, since all those editors are going to be built around a pretty heavy-duty markdown editing interface--I'll hold off on putting post editing there for now to avoid merge conflicts.

EricBAndrews (approved):

EricBAndrews:
Been idle for 12+ hours, merging to unblock #36

Issue #73, opened 2023-06-26
2.0: Rewrite API and Middleware


EricBAndrews:
Create a set of enriched data models to feed to the frontend, decoupling our frontend code from the API spec and enabling features such as:
- Elegantly handling instant vote feedback
- Including whether a user is a moderator in the enriched PostView struct
- Distinguishing between post and comment replies

EricBAndrews:
https://github.com/mlemgroup/mlem/blob/sjmarf/middleware-rewrite-2/Mlem/Models/Content/NewCommunity/README.md

EricBAndrews:
Renamed to "Rewrite API and Middleware"

EricBAndrews:
2.0 middleware is sufficiently robust to call this one closed 🎉

PR #376, opened 2023-07-26
add `PersistenceRepository`

mormaer:
# Checklist
      - Part of #73

# Pull Request Information
This PR moves all the disk based persistence into a repository called `PersistenceRepository` and makes the trackers which rely on it follow a unified pattern for loading and saving.

## About this Pull Request
**Describe what this PR contains in as much detail as possible**

## Screenshots and Videos
No UI changes introduced in this PR

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -19,6 +19,7 @@
 		500C168E2A66FAAB006F243B /* HapticManager+Dependency.swift in Sources */ = {isa = PBXBuildFile; fileRef = 500C168D2A66FAAB006F243B /* HapticManager+Dependency.swift */; };
 		5016A2B12A67EB8600B257E8 /* UIViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5016A2B02A67EB8600B257E8 /* UIViewController.swift */; };
 		5016A2B32A67EC0700B257E8 /* NotificationDisplayer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5016A2B22A67EC0700B257E8 /* NotificationDisplayer.swift */; };
+		503A5D752A78EF3C00488C38 /* Encodable+Export.swift in Sources */ = {isa = PBXBuildFile; fileRef = 503A5D742A78EF3C00488C38 /* Encodable+Export.swift */; };
 		503BA26F2A2C94540052516C /* URL+Identifiable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 503BA26E2A2C94540052516C /* URL+Identifiable.swift */; };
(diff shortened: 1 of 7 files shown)


EricBAndrews (approved):

WestonHanners (approved):

tht7:
I thought when we said protocol based that we'll so something more like this:
```swift
protocol Persistable {
    associatedtype DataType: Codable
    static var preferredFilename: String { get }   
}

extension Presistable {
    private static var path: URL {
            rootPath.appendingPathComponent(preferedFilename, conformingTo: .json)
    }
    private static var rootPath: URL { /* gets the application support path */ }
}
```

then the usage would be like
```swift
extension FiltersTracker: Presistable {
    typealias DataType = [[String]]
    static var preferedFilename: String { "Filters" }
}

class FiltersTracker: ObservableObject {
    @Published var filteredKeywords: [String] = .init()
    @Published var filteredUsers: [String] = .init()

    init() {
       let diskValue = persistenceRepository.load(Self)
        _filteredUsers = .init(initialValue: diskValue?.first ?? [])
        _filteredKeywords = .init(initialValue: diskValue?.last ?? [])
    }
    ...
}
```

tht7:
I have a few points of disagreement, while you are correct that the trackers themselves are not really "persistable" I would also argue that the data itself is not really "persistable" either because without the context of the tracker, it's really meaningless.

What the hell is  '[String]' ? Without the tracker around it to give it context, it means nothing.

To your second point about considering multiple of the same tracker, I agree that this should be an established mechanism in the persistence layer, which is why I named the "filename" "** preferred** Filename" and I think making the file name (aka the key) a second *optional* parameter to the .load function would be fun, 
But consider this, it doesn't matter where you load or save data to without the tracker to give it context it's some meaningless bits, which is why to me, the heart of the mechanism is the tracker itself

You can't save a string or an Int without giving it a name or a container , you can't interact with those values without something to listen to your changes and persist them for you (which I don't think is the responsibility of the persistence layer) (also, if you're talking about raw values, I don't think we should make utilities to encourage the abuse of the file system like this, I think we should make a mechanism that encourages developers to build a proper container for their information instead of trying to store loose bits around)

Also let's talk about the last case you gave, which is actually the coolest.
Since the account tracker doesn't need to know the implementation details of the loading and the saving mechanic it would stand to reason that we could make the token a special type that's recognized by the persistence layer, and is removed from the file system, and saved into the keychain and hydrated back from the keychain Into the tracker all invisibly

If we make the system work like that  (where trackers are the holders and owners of the data and the types can affect their own serialization) we can reuse the same 'SecureToken' (or 'SharedString' that saves itself on iCloud or stuff like that, maybe even 'Image' that will serialize into another file and just leave a reference in the original file) types in other trackers, and receive all the beautiful keychain security (and other features) automatically,

But I feel like I have raised the issue and creating a more synergetic persistence layer might be more trouble than it's worth at this point, so I don't mind giving this a stamp of approval as it stands right now, and we can discuss it further down the line if we need to change stuff

tht7 (approved):

PR #389, opened 2023-07-29
Fixed expanded post not updating when edited

EricBAndrews:
Expanded posts weren't updating when the body of the post was updated. Since they aren't handled with the post tracker, this required some slight jank--PostEditor now has an optional callback that takes in the updated post. This gets used to update the state back in `ExpandedPost` when the edit call completes.

https://github.com/mlemgroup/mlem/assets/44140166/9fe9a552-33a6-4b7c-96b9-f0d3b9e5ede5


Sjmarf (approved):
LGTM 👍

PR #401, opened 2023-07-31
Read Posts

### LINKED ISSUE #383: Content filters not being applied
Adding keywords to the content filters does not appear to actually do anything.

### LINKED ISSUE #400: Indicate and optionally hide read posts
Posts that the user has already read (defined as "viewed detail or interacted with" should be visually indicated, and there should be an option to hide them.

EricBAndrews:
# Checklist
      - closes #400
      - closes #383 

# Pull Request Information

## About this Pull Request
This PR introduces a visual indication of read posts and the option to hide them.

To support this, makes the following backend changes:
- Added API types to mark a post as read
- Added `PostRepository` and put a function to mark posts as read in there
- Reworked how infinite loading feed operates:
  - Added a `shouldLoad` state bool to the feed view that, when true, indicates that more items should be loaded
  - Added and `onChange` task to the top-level feed view that listens to `shouldLoad` and performs the loading. This prevents loading being cancelled by view derenders.
  - Feed items still check whether more content should load when they appear, but now they simply update `shouldLoad` instead of loading that content themselves.
  - Added `shouldLoadContentPrecisely` to the feed tracker, which is very similar to `shouldLoadContent` but only returns true if the item in question is precisely the threshold item. This prevents `shouldLoad` from being toggled by several items in one batch.
- Tweaked how `PostTracker.loadNextPage` works. With filtering read posts it becomes quite common to fetch an entire page that gets filtered out; this was prompting an "end of feed" state because no new posts were added to the feed, thus no new trigger for "should load more posts" was rendered. Now the function keeps loading pages until it either receives an empty *unfiltered* array *or* it adds the threshold number of items.

This PR also performs some minor filtering changes, so I went ahead and fixed the filters problem too while I was at it. Keyword filtering is now case insensitive.

## Screenshots and Videos

https://github.com/mlemgroup/mlem/assets/44140166/30ddeab8-b3d3-43af-b517-153b96004e69

## Additional Context
**Any additional context you'd like to add to help us review this PR**

### DIFF
--- a/Mlem/Enums/Settings/InternetSpeed.swift
+++ b/Mlem/Enums/Settings/InternetSpeed.swift
@@ -1,0 +1,23 @@
+//
+//  InternetSpeed.swift
+//  Mlem
+//
+//  Created by Eric Andrews on 2023-08-02.
+//
+
+import Foundation
+
+enum InternetSpeed: String, SettingsOptions {
+    case slow, fast
+    
+    var label: String { self.rawValue.capitalized }
+    
+    var id: Self { self }
+    
+    var pageSize: Int {
+        switch self {
+        case .slow: return 25
+        case .fast: return 50
+        }
+    }
+}
--- a/Mlem/Views/Shared/Posts/Post Sizes/Large Post.swift
+++ b/Mlem/Views/Shared/Posts/Post Sizes/Large Post.swift
@@ -24,6 +24,7 @@ struct LargePost: View {
     
     // computed
     let maxHeight: CGFloat
+    let titleColor: Color
 
     // initializer--used so we can set showNsfwFilterToggle to false when expanded or true when not
     init(
@@ -33,6 +34,7 @@ struct LargePost: View {
         self.postView = postView
         self.isExpanded = isExpanded
         self.maxHeight = isExpanded ? .infinity : AppConstants.maxFeedPostHeight
+        self.titleColor = !isExpanded && postView.read ? Color.secondary : Color.primary
     }
 
     var body: some View {
@@ -49,7 +51,7 @@ struct LargePost: View {
                     .font(.headline)
                     .frame(maxWidth: .infinity, alignment: .leading)
                     .italic(postView.post.deleted)
-                    .foregroundColor(postView.read ? .secondary : .primary)
+                    .foregroundColor(titleColor)
                 
                 Spacer()
                 if postView.post.nsfw {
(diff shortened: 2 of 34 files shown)


WestonHanners:
What is the indicator? It's not clear on the video.

EricBAndrews:
# UPDATE: ACCESSIBILITY AND POST LOADING

Updated how post loading paginates. Previously, the first page loaded 25 items, and all subsequent pages loaded 50. Because the Lemmy API does "dumb" pagination, this mean that posts 25-49 were being dropped. To fix this, I have:
- Set all post loading to use a static page size
- Added an "internet speed" option in settings. When it's set to slow, pages are 25 items; when it's set to fast, pages are 50

Also added support for indicating read posts when "differentiate without color" is enabled. Users can choose between a check mark or a bar on the leading edge; the width of the bar can be customized.

**Default (differentiate without color off)**
default

**Bar style**
bar

**Check style**
check

**Settings page**
new a11y

Mlem/Enums/Settings/InternetSpeed.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  InternetSpeed.swift
+//  Mlem
+//
+//  Created by Eric Andrews on 2023-08-02.
+//
+
+import Foundation
+
+enum InternetSpeed: String, SettingsOptions {
```
WestonHanners, line 10: Interesting, could we also use this to download lower resolution images from pictrs?

Mlem/Views/Shared/Posts/Post Sizes/Large Post.swift
```diff
@@ -33,6 +34,7 @@ struct LargePost: View {
         self.postView = postView
         self.isExpanded = isExpanded
         self.maxHeight = isExpanded ? .infinity : AppConstants.maxFeedPostHeight
+        self.titleColor = !isExpanded && postView.read ? Color.secondary : Color.primary
```
WestonHanners, line 37: nit, I think we should make this value a computed property

Mlem/Enums/Settings/InternetSpeed.swift
EricBAndrews: I don't see a reason why not, unless pictrs doesn't have the backend support

Mlem/Views/Shared/Posts/Post Sizes/Large Post.swift
EricBAndrews: Resolved in latest commit

WestonHanners (approved):

Issue #427, opened 2023-08-05
Experimental Settings menu for developers and beta users


boscojwho:
## Introduction
An experimental settings page that allows:
- Developers to quickly test certain values (e.g. animations or layout) on actual devices without having to manually recompile
- Beta users to provide useful feedback by varying certain values to their liking.

## Requirements
- [ ] Ability to determine what deployment environment app is running in.
- [ ] Add "Experimental" section to settings view.
- [ ] Ensure experimental settings values are never used in production builds.

## Discussion
We need a way for our code to seamlessly use production values on App Store builds, and use experimental values on other builds.
- Currently, settings values are stored in UserDefaults with the `@AppStorage` property wrapper.
- Perhaps a property wrapper like `@Experimental` could work?

Sjmarf:
We have a page for developer toggles under Advanced -> Developer.

PR #428, opened 2023-08-05
Boscojwho/experimental settings

boscojwho:
# Checklist
      - #427 

# Pull Request Information

## About this Pull Request
Adds an "Experimental" section to settings view that only appears in all builds (including TestFlight) except for App Store.

## Screenshots and Videos
Simulator Screenshot - iPhone 14 Pro - 2023-08-04 at 19 00 46
Simulator Screenshot - iPhone 14 Pro - 2023-08-04 at 19 00 49

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -369,6 +369,7 @@
 		CDF842682A49FB9000723DA0 /* Inbox View Logic.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDF842672A49FB9000723DA0 /* Inbox View Logic.swift */; };
 		CDF8426B2A4A2AB600723DA0 /* Inbox Item.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDF8426A2A4A2AB600723DA0 /* Inbox Item.swift */; };
 		CDF8426F2A4A385A00723DA0 /* Inbox Item Type.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDF8426E2A4A385A00723DA0 /* Inbox Item Type.swift */; };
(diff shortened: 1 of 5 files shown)


EricBAndrews:
By and large I really like this idea, but I agree with @mormaer that some granularity would be extremely valuable

boscojwho:
> > Personally, I'd prefer us to avoid introducing this to the codebase, if the main purpose of it is to allow switching between two values without the need to re-build, given it is a single flag it only allows flipping between two values for testing and also lacks granularity if you wanted to control different settings independently.
> 
> To expand on my above comment, as re-reading it it may come across as dismissive (apologies, I was short on time!) 🙃
> 
> I don't object to the idea of having experimental features (aka 'Labs') in the project. I think that would be a nice addition, and I enjoy seeing that in other apps I use. Even one of my banking apps has them. My objection is just to the proposed approach here where there is a single global flag to switch things on/off.
> 
> I feel that a global approach only really solves the problem for a single developer looking to test a single change/feature on their branch, at which point I think it should be kept on the branch rather than introduced to the project as a whole. Once it's in the project and potential in use by multiple devs granularity becomes a must.
> 
> What I would support would be a more granular system where individual features/experiments could be toggled on and off.
> 
> In addition to that, rather than determine the environment in code at runtime we should consider making use of `xcconfig` files so that we can define our environments without relying on values that may change.

Absolutely haha, I meant for this to actually have actual feature flags or sliders (etc.) to adjust certain values, as in if you enable the "use experimental features" toggle, it would reveal all the available options.

I figured it'd be easier to kickstart discussions/preview some of that code with a PR since I'm working on my own forked version =)

- One use case I wanted this for was changing some spring animation values, which I just wanted to change throughout the day while I'm using the app and dial in what felt nice.
- Agree, we'll need `xcconfig` env values at some point.

Having to manually change the `bundle id` to `[bundle id].dev` and open Xcode to change values is getting a bit tedious, so anything to change that workflow would be nice =)

boscojwho:
Would be great to hear more ideas on how everyone wants to see this implemented.

EricBAndrews:
Closing this as stale--it's a good idea, though, we can definitely revisit if a use case arises

Issue #437, opened 2023-08-06
Emphasise post headlines more for large post sizes


fer0n:
## Introduction
I find it sometimes hard to focus on content in the the large post size feed, because everything looks very similar.

## Requirements
Emphasise the post headlines more.

## Discussion
It would help a lot to put more emphasis on the post headlines (especially for text posts) by de-emphasising the body of the posts, like so (upvote buttons are also de-emphasises slightly): 
![Comparison/proposal](https://github.com/mlemgroup/mlem/assets/20093825/a5b5fb04-a99e-4139-8b11-e0c7f0478724)

EricBAndrews:
I like the idea of graying out the post text a lot. Putting this under 1.2 because 1.1's already got scope creep problems but I'd happily merge this PR whenever.

fer0n:
That was just a mockup, but I might have a look at implementing it

fer0n:
@EricBAndrews not sure if I understood you correctly there and you'd prefer to not deal with this at all for now and I'll remove the PR again. If not and you wanna have a look, I made a very simple implementation of this here: #451

fer0n:
I guess this can be closed sind's it's merged already?

PR #451, opened 2023-08-06
Deemphasize body in feed of large posts

fer0n:
Not sure what you'd like the process to be for this one, but since I couldn't find the task in the backlog I'll just ask you here. Let me know if you'd like this to go another way, or if you want to postpone this until later on.

# Checklist
      - #437

# Pull Request Information
This deemphasises the body in large posts by simply making it 50% opaque.

## Screenshots and Videos
Simulator Screenshot - iPhone 14 Pro - 2023-08-06 at 20 32 45

Simulator Screenshot - iPhone 14 Pro - 2023-08-06 at 20 32 52

## Additional Context
I'm not sure if that's the correct approach, it's certainly a very simple solution. I'm assuming that links should stay clickable, so the alternative approach I can think of is adding a new theme for the `MarkdownView`. That would set all the colors explicitly which is probably better, but would also introduce some more code to maintain. I can give that one a go if you'd prefer it (or any other if you have a suggestion).

### DIFF
--- a/Mlem/Views/Shared/Markdown View.swift
+++ b/Mlem/Views/Shared/Markdown View.swift
@@ -171,6 +171,11 @@ extension Theme {
                         .stroke(Color(.secondarySystemBackground), lineWidth: 1.5)
                 )
         }
+    
+    static let mlemDeemphasized = mlem
+        .text {
+            ForegroundColor(.secondary)
+        }
 }
 
 private extension Color {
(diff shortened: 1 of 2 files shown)


EricBAndrews:
Let's go with the custom Markdown theme and use .secondary--that will make sure this change respects the iOS "decrease contrast" accessibility setting

fer0n:
You're right, it's much better with the theme. I left the link and everything else the same probably makes the most sense. I also updated the screenshots, what do you think?

EricBAndrews (approved):

EricBAndrews:
Looks good!

PR #459, opened 2023-08-07
Polishing for App Store

### LINKED ISSUE #458: ExpandedPost displays compact username and community name

EricBAndrews:
# Checklist
      - closes #458 

# Pull Request Information

## About this Pull Request
Fixes an issue where the username was not properly displaying in compact mode.

Also takes the opportunity to make the following changes:
- Scaling of the flair is now done using `.imageScale` to make it small in trailing mode
- Flair image is no longer disabled when instance location is `.trailing`
- Avatar URL now always takes the Icon64 URL, since that's what gets prefetched
- Avatar size is now a computed var instead of a function and represented as a `CGSize`
- Instance is now `UIColor.tertiaryLabel` instead of `.secondary` with opacity 0.6

Also made some minor miscellaneous tweaks:
- Fix font size in `.mlemDeemphasized` to be consistent with `.mlem`
- Tab switch animation has been greatly sped up
- Nav bar swipe up is disabled on devices with a home button
- Added haptic feedback to account switcher
- Added 2.5px padding to tab bar when device has a home button (this is the Apple amount as far as I can tell)

Screenshot 2023-08-07 at 6 11 19 PM
Screenshot 2023-08-07 at 6 13 44 PM
Screenshot 2023-08-07 at 6 14 04 PM

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -2594,7 +2594,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = "local build";
 				DEVELOPMENT_ASSET_PATHS = "\"Mlem/Preview Content\"";
-				DEVELOPMENT_TEAM = 8B9GNJW88W;
+				DEVELOPMENT_TEAM = 76ULHQGAPN;
 				ENABLE_PREVIEWS = YES;
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = Mlem/Info.plist;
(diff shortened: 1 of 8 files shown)


Mlem.xcodeproj/project.pbxproj
```diff
@@ -2612,7 +2612,7 @@
 					"@executable_path/Frameworks",
 				);
 				MARKETING_VERSION = 1.0.1;
-				PRODUCT_BUNDLE_IDENTIFIER = com.hanners.Mlem;
```
WestonHanners, line 2615: nope

Mlem/MlemApp.swift
```diff
@@ -54,6 +55,15 @@ struct MlemApp: App {
                         appearance.backgroundEffect = UIBlurEffect.init(style: .systemThinMaterial)
                         UITabBar.appearance().standardAppearance = appearance
                         UITabBar.appearance().scrollEdgeAppearance = appearance
+                        
+                        // check whether this device has a home button
+                        if #available(iOS 11.0, *),
```
Sjmarf, line 60: We only support iOS 16+, right? So we don't need to check whether the user is on iOS 11+.

Sjmarf (approved):
LGTM, other than one nitpick. I'll approve anyway though, it doesn't matter all that much.

Mlem/MlemApp.swift
EricBAndrews: Fair enough lmao, that's what I get for copying SO :P

Issue #421, opened 2023-08-04
Gesture to close quick look image preview


tombuben:
It is currently very easy to open the quick look by tapping g any image on screen, but to close it you have to reach to the very top right of the screen and press "Done".

If there was a swipe away gesture to close the view it'd be much more pleasant to use.

ShadowJonathan:
This is the same as how it was done in Apollo, and an essential feature to the preview imo

boscojwho:
@EricBAndrews mentioned something about a new fancier image viewer in the works, but I'm not sure when that'll be ready.

If that's not going to be ready in the next few weeks, we could easily spin up a SwiftUI.UIViewRepresentable version of QuickLook in a day and make users happy in the mean time? =)

EricBAndrews:
Absolutely, I'd happily merge that PR--this is probably our #1 request right now. Do you want the issue? You seem to have an implementation in mind

boscojwho:
> Absolutely, I'd happily merge that PR--this is probably our #1 request right now. Do you want the issue? You seem to have an implementation in mind

Yea, might be able to get to this today or tomorrow =)

boscojwho:
This has been addressed =) See #460

PR #460, opened 2023-08-08
Quick look UI

boscojwho:
# Checklist
      - #421 

# Pull Request Information

## About this Pull Request
Stop-gap measure to fix issue where interactive dismiss gesture doesn't work in Quick Look:
- Replace SwiftUI Quick Look with UIKit Quick Look.

## Additional Context
- Same features as SwiftUI-native implementation.
- Would be pretty easy to add support for multiple preview images.


boscojwho:
Are there even multi-image posts anywhere on Lemmy? 😂 Would be pretty straightforward to add it in a separate PR.

Mlem/Views/Shared/Cached Image.swift
```diff
@@ -14,9 +14,10 @@ import Nuke
 
 struct CachedImage: View {
     
+    @EnvironmentObject var quickLookPaths: QuickLookPaths
```
EricBAndrews, line 17: Why is this an `@EnvironmentObject`? It's not used anywhere else--can this be an `@State var quickLookUrl: URL?`?

Mlem/Views/Shared/Cached Image.swift
```diff
@@ -108,18 +109,15 @@ struct CachedImage: View {
...
                                 }
                             }
                         }
-                        .quickLookPreview($bigPicMode)
-                        .onChange(of: bigPicMode) { mode in
-                            if mode == nil, let dismissCallback {
-                                dismissCallback()
-                            }
+                        .fullScreenCover(item: $quickLookPaths.url) { url in
+                            QuickLookView(urls: [url])
```
EricBAndrews, line 120: This breaks the functionality of marking posts read when the preview is opened--it should be a pretty easy fix, it just needs to call `dismissCallback()` at any point after the viewer is opened.

If you go with the `URL?` suggested above, the previous `onChange` will work--otherwise you'll need to figure something out, since I think applying that `onChange` to an `@EnvironmentObject` is going to mark all rendered posts as read.

EricBAndrews (changes requested):
Fantastic PR, massive quality of life improvement.

I have two pieces of feedback:
- This PR appears to break marking post read when the image preview is viewed
- Why are we using an environment object to drive the quicklook now?

See comments for more.

Overall, though, really nice job!

EricBAndrews:
I've seen "multi-image posts" that are just text posts with a lot of embedded images--we have some very, very tentative ideas about a nice UX to handle those, but that's *way* down the line. AFAIK there's no native Lemmy support for a Reddit-like "image gallery'

ShadowJonathan:
> This PR appears to break marking post read when the image preview is viewed

Imo that feels like expected behaviour? (Maybe toggle-able?)

Mlem/Views/Shared/Cached Image.swift
boscojwho: Initial version's dismiss gesture didn't work, so I tried moving the quick look sheet to the window view. No longer necessary, since interactive gesture works if we present QLPreviewController inside another UIKit view controller. 

Moved that to a @StateObject on CachedImage.

Mlem/Views/Shared/Cached Image.swift
boscojwho: Reverted to the old `URL?` logic, we won't need that `QuickLookPaths` object if multi-image support isn't a thing anyway.

EricBAndrews (approved):

Issue #435, opened 2023-08-06
Tab/navigation bar appearance: non-translucent


fer0n:
I'd love to have an option to make the tab bar and navigation bar non-translucent. Just personal preference, but it would fit in with the other appearance options there are.

I'd be happy to take a look at this and see if I can manage to implement it myself.

EricBAndrews:
Slating for 1.2 customization alongside animation toggle and profile label customizability.

fer0n:
Would you mind if I take a look at this?

fer0n:
@EricBAndrews Would you mind if I took a look at maybe implementing this?

fer0n:
@EricBAndrews I took the freedom to create a PR for this (#483). I'd be happy if you could have a look, but I totally understand if that's not a priority and you got other things to do.

PR #483, opened 2023-08-11
Option for solid coloured navigation bar/tab bar

fer0n:
# Checklist
      - #435

## About this Pull Request
- New option under Settings → Appearance → Tab bar → Translucent (on by default)
- New `.navigationBarColor()` view extension that sets the color/translucency according to the setting
- Added `.navigationBarColor()` modifier to contents of multiple `NavigationStack`

## Screenshots and Videos
**New Tab Bar Setting:**
New Tab Bar Setting

Translucent (unchanged)             |  Non-translucent (new)
:-------------------------:|:-------------------------:
Translucent Tab/Navigation bar  |  Non-translucent Tab/Navigation bar

## Discussion
- Not sure if "Tab bar" is the correct place, since it's also affecting the navigation bar. I could rename it to "Tab Bar/Navigation Bar", or move it to theme as "Translucent Bars" or something else.
- I could't find a way to globally set the navigation bar color without using the modifier under every `NavigationStack`, maybe there's a better way to do this?

### DIFF
--- a/Mlem/Views/Tabs/Feeds/Feed View.swift
+++ b/Mlem/Views/Tabs/Feeds/Feed View.swift
@@ -53,14 +53,14 @@ struct FeedView: View {
     @State var isLoading: Bool = false
     @State var shouldLoad: Bool = false
     
-    @AppStorage("isTranslucentBar") var isTranslucentBar: Bool = true
+    @AppStorage("hasTranslucentInsets") var hasTranslucentInsets: Bool = true
     
     // MARK: - Main Views
     
     var body: some View {
         contentView
             .frame(maxWidth: .infinity, maxHeight: .infinity)
-            .background(isTranslucentBar ? Color.secondarySystemBackground : Color.systemBackground)
+            .background(hasTranslucentInsets ? Color.secondarySystemBackground : Color.systemBackground)
             .toolbar {
                 ToolbarItem(placement: .principal) { toolbarHeader }
                 ToolbarItem(placement: .navigationBarTrailing) { sortMenu }
--- a/Mlem/Views/Tabs/Feeds/Feed View.swift
+++ b/Mlem/Views/Tabs/Feeds/Feed View.swift
@@ -67,6 +67,7 @@ struct FeedView: View {
                 ToolbarItemGroup(placement: .navigationBarTrailing) { ellipsisMenu }
             }
             .navigationBarTitleDisplayMode(.inline)
+            .barBackgroundColor()
             .environmentObject(postTracker)
             .task(priority: .userInitiated) { await initFeed() }
             .task(priority: .background) { await fetchCommunityDetails() }
--- a/Mlem/Views/Tabs/Feeds/Feed View.swift
+++ b/Mlem/Views/Tabs/Feeds/Feed View.swift
@@ -67,7 +67,6 @@ struct FeedView: View {
                 ToolbarItemGroup(placement: .navigationBarTrailing) { ellipsisMenu }
             }
             .navigationBarTitleDisplayMode(.inline)
-            .barBackgroundColor()
             .environmentObject(postTracker)
             .task(priority: .userInitiated) { await initFeed() }
             .task(priority: .background) { await fetchCommunityDetails() }
--- a/Mlem/Views/Tabs/Feeds/Feed View.swift
+++ b/Mlem/Views/Tabs/Feeds/Feed View.swift
@@ -53,18 +53,21 @@ struct FeedView: View {
     @State var isLoading: Bool = false
     @State var shouldLoad: Bool = false
     
+    @AppStorage("showSolidBarColor") var showSolidBarColor: Bool = false
+    
     // MARK: - Main Views
     
     var body: some View {
         contentView
             .frame(maxWidth: .infinity, maxHeight: .infinity)
-            .background(Color.secondarySystemBackground)
+            .background(showSolidBarColor ? Color.systemBackground : Color.secondarySystemBackground)
             .toolbar {
                 ToolbarItem(placement: .principal) { toolbarHeader }
                 ToolbarItem(placement: .navigationBarTrailing) { sortMenu }
                 ToolbarItemGroup(placement: .navigationBarTrailing) { ellipsisMenu }
             }
             .navigationBarTitleDisplayMode(.inline)
+            .barBackgroundColor()
             .environmentObject(postTracker)
             .task(priority: .userInitiated) { await initFeed() }
             .task(priority: .background) { await fetchCommunityDetails() }
--- a/Mlem/Views/Tabs/Feeds/Feed View.swift
+++ b/Mlem/Views/Tabs/Feeds/Feed View.swift
@@ -53,14 +53,14 @@ struct FeedView: View {
     @State var isLoading: Bool = false
     @State var shouldLoad: Bool = false
     
-    @AppStorage("showSolidBarColor") var showSolidBarColor: Bool = false
+    @AppStorage("isTranslucentBar") var isTranslucentBar: Bool = true
     
     // MARK: - Main Views
     
     var body: some View {
         contentView
             .frame(maxWidth: .infinity, maxHeight: .infinity)
-            .background(showSolidBarColor ? Color.systemBackground : Color.secondarySystemBackground)
+            .background(isTranslucentBar ? Color.secondarySystemBackground : Color.systemBackground)
             .toolbar {
                 ToolbarItem(placement: .principal) { toolbarHeader }
                 ToolbarItem(placement: .navigationBarTrailing) { sortMenu }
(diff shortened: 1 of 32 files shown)


kra-mo:
Does this follow the global "Reduce Transparency" setting in iOS?

fer0n:
@kra-mo Not sure what you mean by that. The default is currently the unchanged "Translucent" setting which does follow the "reduce transparency" setting (just checked it). The new "Translucent - off" option that's implemented in this PR literally has no transparency in it anymore that could be reduced. Am I misunderstanding something here?

kra-mo:
If it does, by default follow the global setting, why is there a need for a separate setting inside of the app?

kra-mo:
I also checked and it doesn't look like it does follow the setting? At least on the latest Testflight build.

kra-mo:
Adding preferences has a real cost in terms of UX. The more preferences there are, the less likely someone is to find what they are looking for. (Not to mention the maintenance cost.)

I'd say a more sensible solution would be to not expose this setting to the user and make it follow the global accessibility setting instead. Also to increase the opaqueness of the bar by default as it does seem to be a little too transparent.

fer0n:
There is a difference if you turn on "Reduce transparency", although it's not a huge one and there's still quite some transparency left. If that's an issue for you I'd say open a bug, I don't think that's related to this PR.

I agree that the default transparency is a bit too much and also that it should be avoided to have too many options for the reasons you mentioned. However, I also think that it's fine to have a few theming/customization options and it probably doesn't hurt as much to have them there. 

But ultimately that's not for me to decide.

kra-mo:
I don't understand your line of reasoning.

What you're saying is that there should be 3 different levels of transparency and 2 different places to toggle it. That just seems horrible from a UX standpoint.

> However, I also think that it's fine to have a few theming/customization options and it probably doesn't hurt as much to have them there.

You have to think about who and how many people will actually toggle this option. I would be surprised if it was more than 5% of users.

I would also be surprised if making the setting be bound to the "Reduce transparency" one instead, whilst decreasing the default transparency wouldn't also solve that <5%'s issue (while keeping the Mlem's settings cleaner and more user friendly in terms of accessibility).

fer0n:
> What you're saying is that there should be 3 different levels of transparency and 2 different places to toggle it.

Haha, I get what you're saying. "Reduce transparency" is an accessibility option, though, that applies to the entire OS. Also, reducing transparency isn't the same as having a solid color. Regarding the additional toggle, there's a reason why many apps have a theme with dark/light/automatic option even though that can be set globally (including Mlem), or even offer different colors to choose from. It's because of personal preference. I'd argue setting the tab & navigation bar color (which this essentially does) is a much more substantial cosmetic change than many of the other settings that are currently in Mlem under the appearance page.

> You have to think about who and how many people will actually toggle this option. I would be surprised if it was more than 5% of users.

I know that's a slippery slope argument, but that goes for every setting. Basically no-one ever uses settings, except for the people that do. That doesn't mean there shouldn't be options.

kra-mo:
> much more substantial cosmetic change than many of the other settings that are currently in Mlem under the appearance page.

Then it is an issue with those settings, it's not related to this one.

> I know that's a slippery slope argument, but that goes for every setting. Basically no-one ever uses settings, except for the people that do. That doesn't mean there shouldn't be options.

Which is why the imagined ideal app doesn't have any settings. You should strive to create products that satisfy as many people's needs by default as possible without them having to change anything.

I doubt most people who are bothered by the amount of transparency would seek out a setting to toggle it. The reality is that most users don't even open the settings page unless they need to.

Which is why I suggest just changing the default value to be lower (higher opacity). This would most likely solve most people's problems without adding an extra setting.

And for those for whom it is a serious accessibility issue, the "Reduce transparency" setting should make the bar completely (or close to) opaque. (Like the toggle does on your fork now.)

kra-mo:
Especially because this isn't a question of aesthetics. It's a matter of usability.

fer0n:
#435 is about aesthetics/theming though (and probably where we should've had that discussion). If you think the current transparency is a usability problem feel free to create an issue for that.

kra-mo:
I don't want to create a separate issue or move the discussion elsewhere. I am against merging *this* PR in its current state as I think the problem it solves can be better solved by sensible defaults.

fer0n:
It's not trying to solve a problem though, it's about customizing the look for personal preference.

Just because that happens to also touch on the transparency issue you're having (by showing something else) doesn't mean it's the same issue. There's a better way to address that particular problem, I agree. Again, #435 isn't about accessibility or the level of transparency. It's about a different look.

kra-mo:
> the transparency issue you're having

I am not having any issues with transparency. You yourself have said "[I find the translucent tan/navigation bar to be distracting and much prefer a solid color.](https://photon.lemmy.world/post/lemmy.world/3065525)" meaning it is not about aesthetics, so which one is it?

If it is only about aesthetics though, it seems a useless preference and a maintenance burden. No other app that I know of offers a similar setting and I don't see why this one should in particular.

Mlem/Views/Tabs/Feeds/Feed View.swift
```diff
@@ -53,18 +53,21 @@ struct FeedView: View {
...
     
+    @AppStorage("isTranslucentBar") var isTranslucentBar: Bool = true
+    
     // MARK: - Main Views
     
     var body: some View {
         contentView
             .frame(maxWidth: .infinity, maxHeight: .infinity)
-            .background(Color.secondarySystemBackground)
+            .background(isTranslucentBar ? Color.secondarySystemBackground : Color.systemBackground)
```
EricBAndrews, line 63: I don't think this should change--the `.secondarySystemBackground` is a good visual indicator that a space is "non-content." Things like the refresh and the end-of-feed indicator don't look quite right on default background.

Mlem/Views/Tabs/Settings/Components/Views/Appearance/TabBar/TabBarSettingsView.swift
```diff
@@ -54,8 +55,16 @@ struct TabBarSettingsView: View {
                                        settingName: "Show Unread Count",
                                        isTicked: $showInboxUnreadBadge)
             }
+            
+            Section {
+                SwitchableSettingsItem(settingPictureSystemName: "paintpalette",
+                                       settingName: "Translucent",
+                                       isTicked: $isTranslucentBar)
+            }
         }
```
EricBAndrews, line 64: This should be in Accessibility settings under a section labeled "Transparency," should be labeled "Opaque Insets," and should be off by default (this will also require switching that toggle down through the rest of the code, sorry)

EricBAndrews (changes requested):

EricBAndrews:
I've requested some changes on this and wanted to take the opportunity to touch on some of the points raised in this thread.
- Current tab bar default translucency will remain the same; it is the same translucency that the default SwiftUI tab bar (and all of Apple's native apps) uses. 
- Because we're using `.material` for the tab bar, we get default "reduce transparency" behavior for free; it appears to convert it to a `.thickMaterial` or `.ultraThickMaterial`
- Re: where should it go and the point about settings fatigue, this is not a tab bar setting, since it also affects the nav bar. Fortunately this is offering a modified version an existing Apple accessibility setting; therefore it should live in Accessibility.
- I've also requested that this be renamed "Opaque Insets" (since it affects both nav and tab bar)

Mlem/Views/Tabs/Feeds/Feed View.swift
fer0n: I think it makes for a cleaner look with the adjusted background, especially during the start up: [with changes](https://github.com/mlemgroup/mlem/assets/20093825/2db91c87-e711-45e6-8c6e-cf89fd432a90) | [without changes](https://github.com/mlemgroup/mlem/assets/20093825/ec13ba43-4f45-4e7f-85a0-189542f83bc4) (less pop-in). It also fits the same aesthetic that an expanded post has (it's also the same color on top and on the bottom there. 

The top of the feed always has the progress indicator, while the bottom is probably never reached anyways. I understand what you mean, though. 

Have I managed to change your mind or do you still prefer the `.secondarySystemBackground`? I'll change it if you still prefer that, of course.

Mlem/Views/Tabs/Settings/Components/Views/Appearance/TabBar/TabBarSettingsView.swift
fer0n: I adjusted it. "translucent insets" turned on would mean a translucent tab bar, did I get that right?

Mlem/Views/Tabs/Feeds/Feed View.swift
EricBAndrews: We can merge it as-is--you're right that it's not consistent across the app, though, which should probably be addressed. I've opened a followup #493 to fix that, we can revisit this then.

fer0n: Nice :) 
Maybe it should be the same background regardless of the transparency then? With the transparency it fits better in but  it's the same inconsistency

Mlem/Views/Tabs/Settings/Components/Views/Appearance/TabBar/TabBarSettingsView.swift
EricBAndrews: Yep! Generally speaking our settings follow "default = on", although that's not particularly consistently enforced.

EricBAndrews (approved):

PR #489, opened 2023-08-11
1.0.1 Hotfix: Delete Accounts

EricBAndrews:
Hotfix to comply with the almighty App Review. Adds account deletion flow.

https://github.com/mlemgroup/mlem/assets/44140166/644358ac-48c2-4e6d-9814-959360b3b7f0


Sjmarf (approved):
Nice 👍

IMO we should only show the option in the 'accounts' page in settings, and not in the quick-switcher.

JakeShirley (approved):

EricBAndrews:
Agreed--going to leave as-is for now so Apple Review has a better chance of finding it lol

WestonHanners (approved):

PR #492, opened 2023-08-12
Added haptic feedback to widget wizard

EricBAndrews:
The widget wizard now plays a little haptic on pickup and drop.

Also performed some light refactoring.


Sjmarf (approved):
Nice 👍

Issue #474, opened 2023-08-10
Cursor jumps to the end when starting to edit a post/comment


fer0n:
**Expected Behavior**
If I move the cursor somewhere and start typing, it should enter the text at the cursor.

**Actual Behavior**
The cursor jumps to the end after the first typed letter.

**To Reproduce**
- tap edit in the contextMenu of a comment or post
- move the cursor somewhere else
- start typing 
- the first letter will be where the cursor was, but everything else will be at the end 
![video](https://github.com/mlemgroup/mlem/assets/20093825/81b1661a-5e9e-42b6-9a38-54dc2c2d1f96)

**Mlem Information:**
1.0.1 (475)
iPhone 12 Pro, iOS 16.6

EricBAndrews:
That's very strange--probably due to #433 if I had to guess.

Unable to replicate on my own device:


https://github.com/mlemgroup/mlem/assets/44140166/69ae1310-5c23-497f-8d39-70ab47643326

fer0n:
@EricBAndrews this has been around before the latest release. Maybe the view re-renders after the first change? I also added my device information

fer0n:
I reinstalled the app and at least so far, it's not come up again. I'll update if it does

fer0n:
It's back again. I tried a few different settings, restarting it etc., but it's always there now. Always after the first character and always working after.

PR #494, opened 2023-08-13
Fixed cursor jumping to the end when editing a post/comment (#474)

fer0n:
# Checklist
      - #474 

# Pull Request Information
Fixes the cursor jumping to the end once when starting to edit a commend/post

## About this Pull Request
It moves the `NavigationStack` from both the `ResponseEditorView` and `PostEditorView` into a parent view

## Additional Context
The state seems to be fine and I couldn't quite figure out what was going on. It's especially odd since it's only happening once and not for everyone. I started poking around and messing with some things. Moving the `NavigatioStack` out of the editing views seems to have fixed it. I'll keep testing this on my device for some time to see if the issue pops up again and leave this PR as a draft until I'm reasonably certain it's gone for good.

@EricBAndrews assuming it stays fixed and doesn't reappear, would this be an acceptable solution?

### DIFF
--- a/Mlem/ContentView.swift
+++ b/Mlem/ContentView.swift
@@ -104,10 +104,14 @@ struct ContentView: View {
                 .presentationDetents([.medium, .large])
         }
         .sheet(item: $editorTracker.editResponse) { editing in
-            ResponseEditorView(concreteEditorModel: editing)
+            NavigationStack {
+                ResponseEditorView(concreteEditorModel: editing)
+            }
         }
         .sheet(item: $editorTracker.editPost) { editing in
-            PostComposerView(editModel: editing)
+            NavigationStack {
+                PostComposerView(editModel: editing)
+            }
         }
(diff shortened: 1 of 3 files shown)


fer0n:
I've been playing around with it basically all day and I had 1 or 2 instances where it still happened, but 99% of it is gone and it's generally a huge improvement for a very minor change. So I think it's worth merging.

EricBAndrews (approved):
Looks good to me, glad you figured out how to fix that issue--I was starting to go insane trying to reproduce it 😅

fer0n:
I don't think it's 100% fixed and I still don't understand where it's coming from, but this definitely makes it much much better :)

fer0n:
I don't think it's 100% fixed and I still don't understand where it's coming from or why it's only happening on some devices (I tested it on another iPhone and it wasn't an issue there). But this certainly seems to make it much much better

Issue #480, opened 2023-08-11
Expand images in posts (large post size)


fer0n:
## Requirements
Currently (text in image is cut off):
![screenshot currently](https://github.com/mlemgroup/mlem/assets/20093825/e203a28a-d127-4829-aeed-8ca67979efee)

Possible solution (image fills width and not cut off):
![screenshot proposed](https://github.com/mlemgroup/mlem/assets/20093825/59cf4f6d-7f4f-4363-81a6-ea3c8344dcf7)

## Discussion
I think images shouldn't be cut off in the large view. Having to tap on an image to see everything and then dismissing it again could be avoided.

It could keep the rounded mask + padding, but the image's height shouldn't be cut.

## What I Need
Images shouldn't be cut off anywhere in large posts.

fer0n:
@EricBAndrews is or a conscious decision to have it the way it currently is, or do you think something like this would make sense?

PR #501, opened 2023-08-13
Minor Polishing

EricBAndrews:
# Checklist
- [x] I have read [CONTRIBUTING.md](./CONTRIBUTING.md)
- [x] I have described what this PR contains
- [x] This PR addresses one or more open issues that were assigned to me:
      - 🧹 housekeeping
- [x] If this PR alters the UI, I have attached pictures/videos

# Pull Request Information

This PR makes some minor housekeeping/polishing tweaks:

- CachedImage now accepts a `contentMode` parameter to determine how the image should be resized. By default this is `.fit`, which causes tall images to get shrunk down to fit the maxHeight instead of being clipped; for avatars and such, it has been overridden to `fill`.
- Gave the flair icon the font size of `.footnote` so it's a little less overwhelmingly large
- Used the previously uncalled `postBodyText` function in `LargePost` to re-enable truncating large posts in feed
- Temporarily disabled the "maximize" icon in minimized posts so we can see how it feels without it on the nightly

Simulator doesn't seem to indicate any regression on scrolling performance.

Screenshot 2023-08-13 at 7 47 41 PM
Screenshot 2023-08-13 at 7 48 09 PM
Screenshot 2023-08-13 at 7 48 24 PM

PR #503, opened 2023-08-14
Expand images in feed (large post size)

fer0n:
# Checklist
      - #480

# Pull Request Information
Removes limiting the height of images in the feed (large post size).

## Screenshots and Videos

| Currently | New |
|---|---|
| Simulator Screenshot - iPhone 14 Pro - 2023-08-14 at 08 08 50 | Simulator Screenshot - iPhone 14 Pro - 2023-08-14 at 08 10 36 |

There are two variants of how to expand the image:

| Masked with padding | full width |
|---|---|
| Simulator Screenshot - iPhone 14 Pro - 2023-08-14 at 08 10 02 | Simulator Screenshot - iPhone 14 Pro - 2023-08-14 at 08 10 49 |

The "full width" variant is the latest commit, the "masked with padding" variant is the commit before that.

### Old approach intentional?
I am not sure if the previous behaviour of cutting of the height was intentional, since this approach has the slight downside of potentially having to scroll over a perhaps quite long image post that the user is not interested in. Since this was relatively easy to implement I thought it wouldn't hurt to open a PR to discuss this. 

I'd argue that "the long image is annoying to scroll over" is far less common and having it the other way around, where a comic, meme, etc is being cut off and forces the user to tap, zoom/scroll, and dismiss to view the image happens way more often and is the bigger issue. Additionally, you might see a spoiler of a comic/the image, since the "view port" for the image is centred (aligning to the top while cutting off the height has different downsides ofc).

If the height should be limited for image posts, I think hiding the post content (text) would be the better choice. Better to see the full image and tap to see the discussion about it than see half of both. 

One alternative would also be to only expand the image when there's no text added to it and otherwise keep it the way it currently is. Or make it an option in settings to show images with their full height.

### Between the two variants
I'd go with the full width variant, because the image posts can be better differentiated from link previews and the additional width helps with seeing/reading smaller things in the image. The masked variant looks more uniform and has slightly less height, but I think that's negligible.

### DIFF
--- a/Mlem/App Constants.swift
+++ b/Mlem/App Constants.swift
@@ -105,7 +105,7 @@ struct AppConstants {
     static let localFeedSymbolNameFill: String = "house.circle.fill"
     static let subscribedFeedSymbolName: String = "newspaper.circle"
     static let subscribedFeedSymbolNameFill: String = "newspaper.circle.fill"
-    static let limitImageHeightInFeedSymbolName: String = "rectangle.expand.vertical"
+    static let limitImageHeightInFeedSymbolName: String = "rectangle.compress.vertical"
     
     // sort types
     static let activeSortSymbolName: String = "popcorn" // not married to this idea 
--- a/Mlem/App Constants.swift
+++ b/Mlem/App Constants.swift
@@ -28,6 +28,7 @@ struct AppConstants {
     
     // MARK: - Sizes
     static let maxFeedPostHeight: CGFloat = 400
+    static let maxFeedPostHeightExpanded: CGFloat = 3000
     static let thumbnailSize: CGFloat = 60
     static let largeAvatarSize: CGFloat = 32
     static let smallAvatarSize: CGFloat = 16
--- a/Mlem/App Constants.swift
+++ b/Mlem/App Constants.swift
@@ -105,6 +105,7 @@ struct AppConstants {
     static let localFeedSymbolNameFill: String = "house.circle.fill"
     static let subscribedFeedSymbolName: String = "newspaper.circle"
     static let subscribedFeedSymbolNameFill: String = "newspaper.circle.fill"
+    static let limitImageHeightInFeedSymbolName: String = "rectangle.expand.vertical"
     
     // sort types
     static let activeSortSymbolName: String = "popcorn" // not married to this idea 
--- a/Mlem/Views/Shared/Cached Image.swift
+++ b/Mlem/Views/Shared/Cached Image.swift
@@ -51,7 +51,6 @@ struct CachedImage: View {
         self.cornerRadius = cornerRadius ?? 0
         
         screenWidth = UIScreen.main.bounds.width - (AppConstants.postAndCommentSpacing * 2)
-        // Problem: Size doesn't update properly when parent size changes
         
         // determine the size of the image
         if let fixedSize {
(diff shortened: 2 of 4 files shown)


fer0n:
Shrinking is a good solution as well, although I personally prefer just having it be a "native part of the feed". 

If you're thinking about making in an option, why not go all the way and make it unlimited? Moving from 400 to 800 doesn't seem to be a huge leap with the current image heights that I've come across and depending on the post you'd still have to scroll and zoom around in a second view.

Since you've already implemented it in your other PR, would you like me to add such an option on top of your commit? And close this one?

EricBAndrews:
OK--let's try it with two options, shrunk and unlimited (that should be a settings toggle called "Limit Image Height In Feed", default on).

Let's  roll it into my in-flight--you can open a PR against `eric/feed-items` and we'll get them both merged at once, since they're both addressing the same issue

fer0n:
Okay great :) 
I'll have a look tomorrow (in roughly 8h), so if you want to merge your PR in the meantime that's fine by me. I'll add it on top of yours then.

EricBAndrews:
#501 is in fyi

fer0n:
I'm having an issue that I didn't get past yet, maybe you know what's going on here or how to fix it: the size of the image itself won't change dynamically. Which means:
- Changing the setting is either not working (when turning the expansion on) or leads to the extended image being cut off half way down, with some overflow ([image]([Simulator Screenshot - iPhone 14 Pro - 2023-08-15 at 10 20 37](https://github.com/mlemgroup/mlem/assets/20093825/8b816d9a-4939-4ff6-bd12-7704d6044e4b))
- The post detail can't show the image collapsed if it should be expanded in the feed

I've spent quite some time trying to fix this, I'm assuming it's a layout issue, but maybe it's related to Nuke as well. Restarting of course solves it and it only affects images that have already been displayed. Any idea?

fer0n:
Apart from that issue, I also made the shrunken version have rounded corners and slightly decreased the padding for images to make them look a bit different from links (maybe that helps subconsciously, I've seen people bring this up).
| round image | smaller image padding |
|---|---|
| Simulator Screenshot - iPhone 14 Pro - 2023-08-15 at 10 47 20  | Simulator Screenshot - iPhone 14 Pro - 2023-08-15 at 10 47 55  |

EricBAndrews:
Re: image sizing: I ran into this same problem when I migrated the avatars to `CachedImage`--you'll notice that changing the "show user instance" results in the same bad behavior. `CachedImage` is slowly accumulating bloat and at this point is venturing into spaghetti territory--I *think* resizing is happening both on the main thread within the `LazyImage` closure *and* in the background using proper Nuke processors but with absolutely no logic to handle resizing. Unfortunately we can't really do much to it right now because of the in-flight media viewer (estimated to land within the next 2 weeks, maybe)--once we can pull QuickLook out of there and we have an idea of how the new viewer hooks up to the image we can properly refactor/rewrite. Since that resizing problem is an app-wide issue, this PR can go in without resolving it because it should theoretically be properly fixed when the issue itself is.

Re: visual tweaks--really love the rounded corners on shrunken images! Per the conversation over in the community, we already differentiate links in large posts with the website complex, so for the sake of margin consistency let's keep the padding the same.

fer0n:
Yeah you're right about the padding, that did more harm than good. I removed it and cleaned everything up. I also added a max height for the expanded images (3000) which should be plenty but it's probably good to not have it be infinity. Should be good to go now if you think all that makes sense and don't have anything else to add :)

EricBAndrews (approved):

Issue #361, opened 2023-07-23
Inconsistent markdown rendering between feed and post screen


Rexios80:
**Describe the bug**
The markdown rendering is working differently on the feed and post screens

**Expected Behavior**
They markdown renders the same no matter what screen it is on

**Actual Behavior**
The markdown is rendering differently on different screens

**To Reproduce**
Look at a post with the issue

**Mlem Information:**
1.0.0 (274)

**Additional Context**
Post: https://beehaw.org/post/6747080
![IMG_1992](https://github.com/mlemgroup/mlem/assets/7896519/25cb014f-0e98-44d7-a402-a2fdadc6c40c)
![IMG_1993](https://github.com/mlemgroup/mlem/assets/7896519/99550a17-fefc-4a14-836d-fc9daacb380c)

EricBAndrews:
Labeling this one as 1.2 and a child of #62 

Turns out this is a super thorny problem--Swift offers a lovely little method for truncating lines (necessary in the feed), but it does not play well *at all* with Markdown. 

Current theories on how best to handle this:
- Strip all Markdown from feed presentation
- Custom Markdown handler with some sort of truncation feature

fer0n:
Would you say stripping away markdown is sufficient?

I'm assuming links should stay as text and maybe even still be tappable? Hashtags as well? Don't know if I'll manage to make a decent solution for this, but I might give it a go if you want to assign this to me @EricBAndrews

EricBAndrews:
We should be able to handle this one properly once the new markdown parsing (#1038) is in place--flagging for 2.0

Sjmarf:
This is sorta fixed in 2.0. It'll render all "normal" markdown the same way, though inline images and spoilers aren't rendered and certain elements appear slightly different. This is because the post content needs to be entirely text in the feed so that we can apply `.lineLimit` - we don't have this limitation in the expanded post.

PR #505, opened 2023-08-14
Small improvements to markdown in feed (large post size)

fer0n:
# Checklist
      - #361

# Pull Request Information
This addresses a few of the markdown issues inside the feed (large post size). It's far from being a proper solution for the markdown situation, but it's a small fix that avoids many of the larger issues in the meantime.

## About this Pull Request
- Uses a markdown theme with minimal styling to avoid some of the current issues where the entire post takes on the same styling that the first element in the body has (e.g. everything is a heading, quoted post, ...)
- Also fixes text not being truncated at all in the feed

## Screenshots and Videos
| Previous | New |
|---|---|
| Simulator Screenshot - iPhone 14 Pro - 2023-08-14 at 21 26 27  |  Simulator Screenshot - iPhone 14 Pro - 2023-08-14 at 21 27 34 |
| Simulator Screenshot - iPhone 14 Pro - 2023-08-14 at 21 26 16  | Simulator Screenshot - iPhone 14 Pro - 2023-08-14 at 21 24 57 |

## Discussion 
This doesn't really address #361 in the way it's being described there. I don't think it makes sense to show the exact same markdown from the expanded post in  the feed, it's better to make it a short summary, at least strip away line breaks. That also means block quotes etc would have to be displayed differently.

### DIFF
--- a/Mlem/Views/Shared/Markdown View.swift
+++ b/Mlem/Views/Shared/Markdown View.swift
@@ -172,10 +172,16 @@ extension Theme {
                 )
         }
     
-    static let mlemDeemphasized = mlem
+    static let plain = Theme()
+        .link {
+            ForegroundColor(.link)
+        }
         .text {
             ForegroundColor(.secondary)
             FontSize(16)
+        }
+        .code {
+            
         }
 }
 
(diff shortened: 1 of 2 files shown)


EricBAndrews (approved):
Nice 👍 --we've still got a lot of Markdown work to do across the board but this is definitely way better than before

PR #519, opened 2023-08-17
Refactoring - Update app to use `APIClient` as a dependency 

mormaer:
# Checklist
      - Discussed in development chat, removing direct instantiation of `APIClient` to unblock other work

# Pull Request Information

## About this Pull Request
As discussed the other day in the development chat the fact that we were instantiating the `APIClient` directly throughout the codebase had become a bit of a blocker for other work I've been doing around improving the testability of the codebase.

This PR unfortunately as a result touches _a lot_ of files, but on the whole is not as scary as it looks. The changes can be summarised as the same steps repeated over and over again:

- Update an `APIRequest` to take an `APISession` instead of the `SavedAccount` object
- Add a method via extension on the `APIClient` that allows callers to make the request
- Declare an `@Dependency(\.apiClient)` at the original callsite
- Update the logic to use the dependency instead of instantiating `APIClient()` and creating the request manually

There are 23 commits included, reviewing by commit is recommended as each one will mostly repeat the above steps for each area that was refactored.

In addition to the above changes I've also taken the opportunity to update various places where we were passing a `ContextualError` into the `AppState`, as going forward these should be passed to the `.errorHandler`. It seemed appropriate to make those updates at the same time since they were mostly in the `catch` of the calls I was updating.

One upshot of this is that I was able to remove the `AppState` requirement from quite a few of the views/models, as we no longer require access to it with the above changes 🎉

## Additional Context
I've spent some time testing the areas that were changed in this PR but I've not been able to test the private message / mention related calls. I can't work out how to send a message in the app 🤔

Also, this is partly a _half job_ as the calls should be accessed via the repository pattern however this change was required for me to iron out that pattern on my other branches, we'd still need those calls defined on the `APIClient` like is done here, so at least part of the work is now in place 🙏

### DIFF
--- a/Mlem/Views/Shared/Posts/Feed Post.swift
+++ b/Mlem/Views/Shared/Posts/Feed Post.swift
@@ -326,7 +326,9 @@ struct FeedPost: View {
             postTracker.update(with: updatedPost)
         } catch {
             hapticManager.play(haptic: .failure, priority: .high)
-            appState.contextualError = .init(underlyingError: error)
+            errorHandler.handle(
+                .init(underlyingError: error)
+            )
         }
     }
 
--- a/Mlem/Views/Shared/TokenRefreshView.swift
+++ b/Mlem/Views/Shared/TokenRefreshView.swift
@@ -25,8 +25,6 @@ struct TokenRefreshView: View {
         case onetimecode
     }
     
-    @EnvironmentObject var appState: AppState
-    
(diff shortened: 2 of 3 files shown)


EricBAndrews (approved):
Another beauty of a PR! Your contributions keep this app manageable :)

Regression testing checks out--let's get this merged and put it through its paces on the nightly build

PR #523, opened 2023-08-19
allow control of transport behaviour in `APIClient`

mormaer:
# Checklist
      - Small tweak to allow/improve testability of the application

# Pull Request Information

## About this Pull Request
This PR is the small change I had looked to make on one of my other branches which ended up requiring the big ol' refactor that was #519.

This change introduces a `transport` function which is passed to the `APIClient` when it is initialised that describes how it will interact with the network. The one that is now passed in at the live values dependency injection point is the same behaviour as before so there is no change seen when running the application.

However, it now allows us to pass different values under test.

The value currently sent to the test value is `unimplemented()` which will cause any test that attempts to hit the network to fail immediately. In general our tests should not be expected to touch the network layer (we'll more often stub the repository layer).

If we did want to hit the client under test and return mocked data this change also makes that possible as we could do something along the lines of:
```swift
let model = withDependencies {
    $0.apiClient = .init(transport: { _, request in
        guard let urlString = request.url?.absoluteString else {
            throw APIClientError.cancelled
        }
                
        switch urlString {
        case "some_url_we_care_about":
        // return whatever data/response we want...
            return (Data(), URLResponse())
        default:
            throw APIClientError.cancelled
        }
    }) 
} operation: {
    SomeModel()
}
```

The ability to stub at this layer would come in handy if we started introducing caching to the repositories, as we could assert what URLs are being hit etc... but as above, most tests will not touch the client at all - and so the `.testValue` supplied is `unimplemented` to fail if they do.

### DIFF
--- a/Mlem/API/APIClient.swift
+++ b/Mlem/API/APIClient.swift
@@ -31,6 +31,7 @@ class APIClient {
 
     let urlSession: URLSession
     let decoder: JSONDecoder
+    let transport: (URLSession, URLRequest) async throws -> (Data, URLResponse)
     
     private var _session: APISession?
     private var session: APISession {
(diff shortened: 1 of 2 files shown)


EricBAndrews (approved):
LGTM

Issue #129, opened 2023-06-13
Refactor views


mormaer:
As discussed in Mlem Development we want to adopt a solid architecture in the application.

Although we have many ideas about how that will look and patterns we'd like to use, as a first step we should work our way through the application and tidy up the current code (specifically within our views) so that we have an understandable base to look over and help inform the next decisions we make regarding the architecture.

This will bring us a few benefits as we work through this task;

- We can each gain an understanding of the areas of the application we choose to refactor
- The scope of each task remains fairly small (eg, just refactor it, don't _change_ what/how it works)
- Identify/discuss any issues in the current implementations
- Allow us to move between different areas of the code base more easily, as the format will become standardised

**Each view we choose to refactor will be given it's own task and should (ideally) be a single PR.**

**Each PR should receive a review of both the code changes and a regression check before merging.**

Issue #21, opened 2023-06-21
Discovery / Guest Mode


mormaer:
## Introduction
This issue is a place to continue discussions that started on mlemgroup/mormaer-mlem#198.

The idea is that when first opening the app as a new user, who potentially is brand new to Lemmy in general, they would be presented with content/options to explore the _fediverse_ before proceeding to sign-up to an instance. Alternatively, a guest mode could be implemented where users could explore without registration but with limited options as they would not be able to interact with any posts etc.

## Requirements
At present it's unclear, more discussion is required.

## Discussion
Here are some of the comments from the attached PR (#198) where this discussion started:

@JakeShirley:
> I can see a world where there is a "Discover a server" experience in the app. Lemmy is definitely less than user friendly.

@buresdv:
> I can see a "guest mode" working is if we choose a specific set of instances (further discussion is warranted to choose which instances they will be), give them some context (for example "this is a homepage for tech nerds", "this is a homepage for general discussion", etc.), and then let the user pick from them. We should make it discoverable to the average user without them having to scour for instances.

> Before a guest mode is implemented, we also have to implement a registration workflow so the users can sign up to their chosen instance right from the app.

@ShadowJonathan:
> Random; We could probably also hook this into the signup flow, i.e. now have a user scour for instances but allow them to pick from a list, which new instance to sign up at.

> This should probably be discussion for a tracker issue though.

chrispomeroyhale:
I have some bandwidth -- I would be happy to take this on as a first time contributor as I outlined in #20 -- but I also see someone else already gave it a first attempt in https://github.com/mlemgroup/mormaer-mlem/pull/198

In my experience, while we can design a perfect guest mode experience, it's usually easier from a contribution standpoint to keep-it-simple and then evolve from there.

It seems to me the central issue is that the Lemmy APIs themselves don't always require authentication thereby allowing a kind of "guest mode". But the app makes the assumption that all API calls require auth.

The app already does error handling (with throws/catches) which is great! So at a bare minimum we need to figure out how we want to architect APIs without accounts.

There are many ways to do this. https://github.com/mlemgroup/mormaer-mlem/pull/198 makes SavedAccount optional and provides a global default instance. But this introduces global state.

My proposal was making `username` and `accessToken` optional on `SavedAccount` and then throwing (or returning optional) from the `Request` inits. But then `SavedAccount` becomes a bit of a misnomer.

What do you all think?

EricBAndrews:
Related to #206

[...]

PR #546, opened 2023-08-25
Separate site information from `AppState`

mormaer:
# Checklist
      - a very loosely related pre-cursor to #21 

# Pull Request Information

## About this Pull Request
This PR removes the responsibility of loading site information from our `AppState` object and instead moves it up to the `Window` level where other _onLogin_ actions are taking place.

We still only use a single value from the site response but that will change if/when we start taking site languages and possibly versions into account later.

With this change the only responsibility left in our `AppState` is holding/managing the active account, I'm going to consider how that handling works as part of #21 and may move that handling elsewhere, effectively allowing us to bin the `AppState` all together 😱 

## Screenshots and Videos

No visible change to behaviour / UI

| VIDEO |
| --- |
| <video src="https://github.com/mlemgroup/mlem/assets/5231793/7c9d0909-80ae-4d32-9d1b-c97ddf91f404"> |

## Additional Context

The object that holds the site information has been designed as an `@Dependency` instead of an `@EnvironmentObject` like most of our trackers.

The reason being that we receive the same behaviour this way, but the object can then be accessed from anywhere using `@Dependency(\.siteInformation)` as opposed to using the native `@EnvironmentObject` approach which is tightly coupled to `SwiftUI` view hierarchies meaning you cannot easily gain access to a tracker from a non-view object.

This will become more apparent as we work toward separating our business logic as part of #129.

## Regression checks

If you wanted to confirm this hasn't broken anything in lieu of test coverage please check the attached video and/or carry out these steps:

- [ ] Log in to the app on an instance that allows downvotes (`test-mlem.jo.wtf`)
- [ ] Confirm when you navigate into the feed/comments etc the downvote option/swipe actions are present
- [ ] Switch account to an instance that does not allow downvoting (`beehaw.org`) or proxy the API and return `false` for the `siteView.localSite.enableDownvotes` property in the `GET /site` call.
- [ ] Confirm when you navigate into the feed/comments etc the downvote option/swipe actions are **not** present

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -40,6 +40,8 @@
 		507573942A5AD59E00AA7ABD /* EquatableError.swift in Sources */ = {isa = PBXBuildFile; fileRef = 507573932A5AD59E00AA7ABD /* EquatableError.swift */; };
 		507573962A5AD5CF00AA7ABD /* ContextualError.swift in Sources */ = {isa = PBXBuildFile; fileRef = 507573952A5AD5CF00AA7ABD /* ContextualError.swift */; };
 		507573982A5AD60100AA7ABD /* ErrorAlert.swift in Sources */ = {isa = PBXBuildFile; fileRef = 507573972A5AD60100AA7ABD /* ErrorAlert.swift */; };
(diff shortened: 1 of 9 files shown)


Sjmarf (approved):
Nice 👍

PR #547, opened 2023-08-25
Error handling

Sjmarf:
I've made some changes to the way errors are handled in the `ExpandedPostView` and `UserView`. I've also refactored `UserView` quite a bit.

Error changes:

If the error handler is told to show an alert, and the internet is offline, it will show a simple "You're offline" toast instead of a full error message. 

<br/>  
<br/>  
If the user loads a comments page, and an error is thrown, the error message will be shown inline instead of as an alert. This makes it less intrusive. If the error was caused by the user being offline, the page will attempt to reconnect every second. If the internet comes back online, it will load the comments automatically. 

<br/> 
<br/>  

<br/>
<br/> 

If the error was _not_ caused by the internet being offline, it'll instead show "Something went wrong" with a "Try again" button that the user can press to attempt to load the comments again. There is also a small button that allows the user to read the full error message (this is just `error.localizedDescription`). 

If there are any common errors that come up here (I'm not aware of any myself), we should provide more user-friendly descriptions for those errors instead of relying on the debug description. If a custom description were to be provided for an error, the button to show the full error message won't be shown - unless "Developer Mode" is enabled in Settings, in which case viewing the full error will always be an option.

All of the above only applies to the _first_ time the comments are loaded - if the user refreshes the page and an error is thrown, the regular error handler is used instead. 

I've implemented the same system for `UserView`. I'd like to expand this system to more places in future, if others approve.

### DIFF
--- a/Mlem/Views/Shared/Error View.swift
+++ b/Mlem/Views/Shared/Error View.swift
@@ -6,16 +6,131 @@
 //
 
 import SwiftUI
+import UniformTypeIdentifiers
+import Combine
+
+struct ErrorDetails {
+    var title: String?
+    var body: String?
+    var error: Error?
+    var icon: String?
+    var buttonText: String?
+    var refresh: (() async -> Bool)?
+    var autoRefresh: Bool = false
+}
 
 struct ErrorView: View {
-    let errorMessage: String
+    @AppStorage("developerMode") var developerMode: Bool = false
+    
+    @State var errorDetails: ErrorDetails
+    
+    @State private var showingFullError: Bool = false
+    @State private var refreshInProgress: Bool = false
+    
+    var timer = Timer.publish(every: 1, tolerance: 0.5, on: .main, in: .common)
+        .autoconnect()
+    
+    init(_ errorDetails: ErrorDetails) {
+        var errorDetails = errorDetails
+        if !InternetConnectionManager.isConnectedToNetwork() {
+            if errorDetails.title == nil {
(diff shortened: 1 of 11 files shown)


Mlem/Views/Shared/Error View.swift
```diff
@@ -6,16 +6,131 @@
...
+                if let body = errorDetails.body {
+                    Text(errorDetails.title ?? "Something went wrong.")
+                        .font(.title3.bold())
+                        .foregroundStyle(.primary)
+                    Text(body)
+                } else {
+                    Text(errorDetails.title ?? "Something went wrong.")
+                        .font(.title3.bold())
+                        .foregroundStyle(.primary)
+                }
```
EricBAndrews, line 66: *nit*

Duplicate code, can be refactored to:

```
Text(title)
    .font
    .foreground

if let body = errorDetails.body { Text(body) }
```

Mlem/Views/Shared/Error View.swift
```diff
@@ -6,16 +6,131 @@
...
+                if let refresh = errorDetails.refresh {
+                    if !errorDetails.autoRefresh {
```
EricBAndrews, line 68: *nit*

Reduce nested clause depth by combining conditionals:

`if let refresh = errorDetails.refresh, !errorDetails.autoRefresh`

EricBAndrews (approved):
A couple really tiny nits, looks good overall

PR #551, opened 2023-08-26
fix - ensure consistent instance group order

mormaer:
# Checklist
      - Noticed while playing around with how we manage accounts for other work...

# Pull Request Information

## About this Pull Request
A tiny change that ensures the keys we use when rendering the list of accounts are always in a consistent order.

The groups are being stored as a map/dictionary whose keys are unordered, so each time we ask for them we may receive them in a different order.

This applies the default `.sorted()` method to them which will ensure they are always alphabetically ordered.


boscojwho (approved):
👍
Probably could move this to a `get var` on accounts tracker in the future.

PR #554, opened 2023-08-27
Refactor - `SavedAccountsTracker`

mormaer:
# Checklist
      - just some refactoring 🧹 

# Pull Request Information

## About this Pull Request
This PR does two things;
 

fa2fd454df4a58b96721f790abcce03d4ecededf refactors the `SavedAccountsTracker` to remove the need for us to manually keep the `.savedAccounts` and `.accountsByInstance` values in sync with each other.

Manually doing this may lead to issues if we forget to update one when the other changes.

The changes in this commit add an observer to the `.savedAccounts` and then use the standard library dictionary initialiser to re-build the map in response to the change. This allows us to work solely on the accounts list knowing that any change we make will be automatically reflected in the instance map, it becomes the source of truth.

This does mean that we're recreating the map rather than adding/removing keys, however re-creating the map has no performance trade-off unless a user was to have 1000s of accounts spread over 100s of instances... which I don't think is a concern.

28da91ccb56a3229c6143e2e7c8891621aaf8e3a changes the `SavedAccountsTracker` to be used as an `@Dependency` as opposed to an `@EnvironmentObject`. As mentioned in previous PRs this means we can access the value outside of view contexts which will prove useful as we continue to separate logic from the layout code.

### DIFF
--- a/Mlem/Views/Tabs/Feeds/Community List/Community List View.swift
+++ b/Mlem/Views/Tabs/Feeds/Community List/Community List View.swift
@@ -99,11 +99,6 @@ struct CommunityListView: View {
 // MARK: - Previews
 
 struct CommunityListViewPreview: PreviewProvider {
-    static var appState = AppState(
-        defaultAccount: .mock(),
-        selectedAccount: .constant(nil)
-    )
-    
     static var previews: some View {
         Group {
             NavigationStack {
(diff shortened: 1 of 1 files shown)


boscojwho (approved):
👍

PR #559, opened 2023-08-31
fix - make instance metadata parsing handle (some) changes

mormaer:
# Checklist
      - No issue raised, but it was broken

# Pull Request Information

## About this Pull Request
There was a change made on the `awesome-lemmy-instances` side yesterday(?) which added new field (`MO`) to the data which tracks the number of months online.

Unfortunately our parser was reliant on the structure and order of the fields meaning that we could no longer parse the response.

I've refactored the parser so that it now builds it's own set of indexes from the CSV which should avoid things breaking for this reason in the future.

## Additional Context
It's worth noting that this will mean new fields being added should be fine, but parsing a CSV like this on the front-end is likely to fail for other reasons, if any of the headers change name for example we'd break again. I have another branch which adds a bundled fall-back file to use if that happens which I'll put up for review when I get a moment.

### DIFF
--- a/Mlem/Logic/InstanceMetadataParser.swift
+++ b/Mlem/Logic/InstanceMetadataParser.swift
@@ -68,7 +68,7 @@ struct InstanceMetadataParser {
             throw ParsingError.requiredHeaderMissing
         }
         
-        let metadata = lines.compactMap { try? parse($0, using: indexes) }
+        let metadata = lines.compactMap { try? parseLine($0, using: indexes) }
         
         guard !metadata.isEmpty else {
             throw ParsingError.noInstancesFound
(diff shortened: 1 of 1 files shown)


Mlem/Logic/InstanceMetadataParser.swift
```diff
@@ -0,0 +1,142 @@
...
+    
+    private static func index(of field: Field, in fields: [String]) -> Int? {
+        guard let index = fields.firstIndex(of: field.rawValue) else {
+            return nil
+        }
+        
+        return Int(index)
+    }
+    
+    private static func parse(_ line: Substring, using indexes: IndexContainer) throws -> InstanceMetadata? {
```
EricBAndrews, line 103: *nit*

Rename to `parseLine` so we don't have two methods named `parse` that do different things

EricBAndrews (approved):

Issue #328, opened 2023-07-19
Share post content


elias-fox:
In Apollo, a very common workflow for me was to long press on an image or video that was posted and share to another app. Currently in Mlem this shares the post URL, whereas it would be nice to just grab and share the content itself.

EricBAndrews:
Child of #31, should be closed when that PR comes in

Sjmarf:
Closed in #768

Issue #256, opened 2023-07-11
Share as image/select text


NathanielIR:
There were two great sharing features in Apollo:

Select text - when you haptic touched on a comment, there was a select text button which was super handy! If a comment wasn’t in English, you could just select the text and use the built in translate feature. Or if you didn’t know what a word meant, you could look it up! Or you could just copy the text. It was great!

Share as image - Another handy sharing feature that I used to use alllll the time was share as image. You could Haptic Touch one a comment or post and tap share as image and it would render it out into a nice well formatted image to share with anyone!

Issue #244, opened 2023-07-09
Add long-press/3D Touch context menu for images


raptormama:
One of the most commonly-used features of Apollo and similar clients is/was to save images posted by other users.

When viewing an image, whether it is as part of a post or standalone (after tapping on the image in a list of posts or from within a single post), the user should be able to press and hold (or 3D Touch, on older devices) to see a context menu that allows the user to save the image to the camera roll (and other relevant share sheet options) or to share it (email it, print it, forward it, et cetera).

einsteinx2:
If my suggestion should be a separate Github issue, please let me know, but it would be nice to be able to directly share images rather than the Lemmy post link when the image is opened via 3D Touch/long press. Especially for memes, often I want to share just the image to either a family iMessage group or work Slack, and currently I have to do the following dance:

- Share post
- Use Opener to open in Safari
- Long press image
- Save to Photos
- Share Photo
- Delete Photo

Presumably the share menu on the post itself should always share the post and not just the post image, so I figured this would be a good place for direct image sharing.

einsteinx2:
Whoops, I'm an idiot who didn't finish reading the post I replied to... I see you also mention direct sharing. Anyway I'll leave my reply up for the extra explanation of the use case for direct image sharing...

EricBAndrews:
Child issue of #31

EricBAndrews:
Closed in #767

PR #561, opened 2023-09-01
New Media view

tht7:
# Checklist
      - #328
      - #256
      - #244

# Pull Request Information

## About this Pull Request
Brand new media viewer!
Now GIFs and MP4s (and a bunch more) play inline!
Additionally there's a brand new Fullscreen View, fully customisable (not a black box from Apple)
This new fullscreen view allows for zooming, video scrubbing, swipe-to-dismiss, and double-tap to upvote

## Screenshots and Videos
IMG_7619
IMG_7617
IMG_7618
IMG_7620

## Additional Context
This is very much a work in progress, there's lots of features left to writ and some of the code is really dirty looking, but for V1 of this it's shippable, and I think it would benefit us on the long run to ship it asap so the entire team can start building new features on top of this

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -258,6 +258,9 @@
 		B104A6DC2A59BF3C00B3E725 /* NukeUI in Frameworks */ = {isa = PBXBuildFile; productRef = B104A6DB2A59BF3C00B3E725 /* NukeUI */; };
 		B104A6DE2A59BF3C00B3E725 /* NukeVideo in Frameworks */ = {isa = PBXBuildFile; productRef = B104A6DD2A59BF3C00B3E725 /* NukeVideo */; };
 		B104A6E02A59C19400B3E725 /* OperationQueue - Easy init.swift in Sources */ = {isa = PBXBuildFile; fileRef = B104A6DF2A59C19400B3E725 /* OperationQueue - Easy init.swift */; };
+		B1149F0D2AA21400009E3EB9 /* MediaCoordinator.swift in Sources */ = {isa = PBXBuildFile; fileRef = B1149F0C2AA21400009E3EB9 /* MediaCoordinator.swift */; };
+		B1149F0F2AA249F8009E3EB9 /* UpvoteView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B1149F0E2AA249F8009E3EB9 /* UpvoteView.swift */; };
+		B1149F112AA2508B009E3EB9 /* ScrubberHint.swift in Sources */ = {isa = PBXBuildFile; fileRef = B1149F102AA2508B009E3EB9 /* ScrubberHint.swift */; };
 		B11A1A782A4EFF2B00520DB4 /* Feed Root.swift in Sources */ = {isa = PBXBuildFile; fileRef = B11A1A772A4EFF2B00520DB4 /* Feed Root.swift */; };
 		B11D72832A49FAA7009DC22F /* Cached Image.swift in Sources */ = {isa = PBXBuildFile; fileRef = B11D72822A49FAA7009DC22F /* Cached Image.swift */; };
 		B14E93C02A45CA3400D6DA93 /* Post Link.swift in Sources */ = {isa = PBXBuildFile; fileRef = B14E93BF2A45CA3400D6DA93 /* Post Link.swift */; };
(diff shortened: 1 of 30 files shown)


boscojwho:
Haven't deep-dived into this PR yet:
- Feed scrolling started to lag significantly after scrolling down in Feed by several pages (I haven't reproduced this, and wasn't connected to debugger, so I'm not sure what caused this).
- On my iPhone 13 mini, the CPU ran consistently at 3-7% after scrolling down (say, 3-5 pages?) in the Feed.  Haven't checked what the cause is, but the latest `/dev` build doesn't do this.

tht7:
@boscojwho rendering GIFs (and videos) takes significantly more memory and CPU, so I am aware,
plus I'm still working on a few optimisations to bring stuff back to silky  smooth this is very much an initial feedback round

Sjmarf:
Looks nice :)

The back button in the corner extends into the safe area for me.



I also feel like the padding on the bottom bar could be adjusted slightly. The title appears to sit further left than the interaction bar. Also, the interaction bar is shorter here than shown in the feed view - if the user chooses to put too many items in the interaction bar, it could appear squished. Maybe we could fix this by making it wider, or adding a fixed (non-customisable) setup instead.

tht7:
@boscojwho thanks for the feedback will fix those!

tht7:
@boscojwho @Sjmarf Here's what got addressed so far:

- Took some basic steps to keep the CPU and Battery in check, so things should be smoother now (might still need some work, feedback me)
- `I feel like the padding on the bottom bar could be adjusted slightly`: not perfect but adjusted!
- `The dismissing animation is different on the main feed (just disappears) to the profile tab (slides down)`: now no one gets needlessly animated
- `If you open an image and immediately zoom in, it 'jumps' to the left`: couldn't quite repro, but cancelled the most sus lines
- `I think that the overlay fade in/out on tap animation could be a little faster`: much faster now! from 1s to 0.2s
- `Sometimes, clicking on an image in the post doesn't open the image viewer. Then, when I open the post in the expanded view, the image viewer appears`: SwiftUI weirdness; eliminated

tht7:
Added Live Text support
IMG_7638

boscojwho:
Question: Would it make more sense to separate the inline GIF view and the media viewer into separate PRs?
1. Inline GIFs appears much closer (if not already) ready, so we can start beta testing it on private or public builds earlier.
2. Inline GIFs and Media Viewer are really two separate features, so doesn't make sense to keep inline GIFs stuck in PR if it's ready.
3. It's pretty simple to revert to using the existing QuickLookUI while we get the media viewer into shape (Inline GIFs work seamlessly with the existing QuickLookUI).

boscojwho:
Re. Media Viewer:

1. Is there a way to customize the double-tap action? I think double-tap should default to system behaviour (i.e. zooming), especially for one-handed use.

2. Long pressing on a photo triggers a context menu, but that action makes it a bit difficult to trigger the system copy-and-paste feature for photo subjects. Any thoughts?

3. The current system QuickLookUI allows users to fluidly edit a photo right in place (think making memes =P), whereas the new media viewer would require users to save the photo, then leave the app to go to Photos.app to make edits there. QuickLookUI also has built-in search results for photo subjects. Any thoughts on this?
- Bigger context discussion that I thought about while working on swipey actions, but for user interactions where there is a system-wide UI/UX pattern, would it make sense for users to be able to have the option to use the system UI/UX pattern?

boscojwho:
Did some investigation on performance issues on `/LPLinkView` branch on my iPhone 13 mini:
- In c/gif (lemmy.world), I scrolled all the way to the bottom and loaded a total of 44 posts.
- In Instruments, it's showing that 123 instances of MediaCoordinator in memory.
- System is calling CoreMediaViewer's `dismantleUIView(...)`, so theoretically MediaCoordinator shouldn't be in memory soon after that.

I compared the # of instances in memory for `WebsiteIconComplex` (which is also a UIViewRepresentable) and `CoreMediaViewer` in FeedView with the assumption that there should at most be the same # of instances of both in memory (in reality, less of CoreMediaViewer, since not every post has media?). Scrolling down a few pages, I got these results:
- 3 instances of `WebsiteIconComplex.Coordinator`.
- 77 instances of `MediaCoordinator`.

Instruments doesn't register this as a leak, but pretty sure it is.
Screenshot 2023-09-18 at 1 06 40 AM
- The responsible caller is the MediaCoordinator's initializer, which takes in a SwiftUI view as a parameter, but it isn't an `@ViewBuilder`.
- I'm not sure if passing in that parent `view` like that might be causing some issues?

This one's a bit hard to debug because either the app crashes or Instruments hangs/crashes before we can get enough data =((

I'll try to find some more time to investigate =)

tht7:
@boscojwho loved your comments here are my thoughts:
We can separate the inline-gifs from the media viewer PRs since the code is kinda modular so we can cherry-pick only the CoreMediaViewer and CachedImage and be done with it, but I'm afraid this will lead to this PR being abandoned

Let's talk about QuickLookUI, I love it, but I think the whole point of building this viewer is that we'll have full control of the media experience, so we can do more then the native allows, granted it will take a few versions and more effort to get to feature parity but for more cool features it might be worth suffering thro it

So more to the point:
Re. Re. Media Viewer:
 1. Is there a way to customise the double-tap action: Yes! this is just a view modifier it's called `onFullScreenDoubleTap` there's a bunch more of these customisation points in the `Environment - FullScreen Helper.swift` file
 2. The long press is kinda confusing I admit but it was long requested by users and that's how most social apps handle it, remember Apollo. But to make it less confusing I should probably disable that context menu when LiveText is on (btw to set the context menu a parent view can use `onFullscreenContextMenu` modifier and the child view - CoreMediaViewer - can use `fullscreenContextMenuDonationCollector` from the env)
 3.  I think we can build our own editor if there's demand, or maybe try to use PencilKit, and about the subject lifting, there's the Vision API that can probably help us with that (https://developer.apple.com/wwdc23/10176)

The last point is more of an open discussion, should we roll our own or use the system?.. I think it's a delicate balance, I think going custom has some fantastic long term benefits but some pretty awful short term disadvantages:

 - On one hand going custom will allow us to provide a much more customisable experience to the user (like a toggle for what happens when you double tap, a toggle to remove LiveText, a crazy cool gallery view, or **swipe to scrub** which I think was the original inspiration).
 
 - On the other hand it's hard to nail down the System level polish and will probably take a few versions until we get there

about the leak, I think I see why it happened, I keep a circular reference to the parent view...

Sjmarf:
I don't think PencilKit support is a huge concern - it's not too much trouble to open the photos app and do it there imo. I agree that a custom image viewer is necessary to get the level of control that we want. We could absolutely leave an option for the system image viewer, though, at least whilst the custom viewer isn't quite as polished

boscojwho:
@tht7 Don't worry too much about the system behaviour option for now, I just wanted to write that down somewhere before I forget lmao.

Took another look at the latest commit on `/LPLinkView` branch. Basically, I see three separate technical and UI changes:
1. LPLinkView
2. Inline GIF viewer
3. Media Viewer

TL;DR: We might want to chain this PR like this: LPLinkView -> Inline Gif Viewer -> Media Viewer

The LPLinkView (1) changes are very much isolated to the our current UI, touches very few files, so we can probably get this in as-is (I wouldn't mind helping with this =D)

I might have missed something, but (2/3) are only really tied together in `CachedImage`, so if we revert some of the changes there and fix some of the performance regressions (might already be fixed, haven't had to chance to test yet), we could merge that in right after LPLinkView. I can also help with this (pretty sure we don't need to cherry pick, since the full screen media viewer code won't be called at all if it's removed from `CachedImage`).

Then we can focus on just the media viewer (and all the technical and UX stuff).

PR #562, opened 2023-09-01
fix - identify decoding errors in APIClient

mormaer:
# Checklist
      - came up in dev chat with @Sjmarf, just adding support on the client side for these errors to make displaying it on the UI simpler

# Pull Request Information

## About this Pull Request
A small QOL improvement on the `APIClient` that now returns a `.decoding(Data)` error when we don't encounter a properly formatted `APIError` object, and so attempt to decode the expected response and fail. As we're now returned the original `Data` object we can map that into a `String` for display if we wish.

Also added some better support for displaying the error in the log/console that makes use of it.

## Screenshots and Videos
| VIDEO |
| --- |
| <video src="https://github.com/mlemgroup/mlem/assets/5231793/53df363c-deef-4a75-a782-cd2f2507862a"> |

### DIFF
--- a/Mlem/API/APIClient.swift
+++ b/Mlem/API/APIClient.swift
@@ -18,6 +18,22 @@ enum APIClientError: Error {
     case response(APIErrorResponse, Int?)
     case cancelled
     case invalidSession
+    case decoding(Data)
+}
+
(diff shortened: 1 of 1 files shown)


boscojwho (approved):
Nice QoL!

PR #566, opened 2023-09-02
Improved Search

Sjmarf:
Changed the search tab to allow searching of users and posts. This isn't quite done yet, but I wanted to gather some feedback on what I've got so far.

<hr>

<hr>

The time results take to load seems to vary; sometimes it can be a little slow. If anyone knows of a way I can speed this up (I'm not too familiar with async stuff), that would be greatly appreciated :)

Things I still need to do:
- Right now, only 50 results are loaded and it won't load more when you scroll down.
- Reimplement recent searches.
- Fix a couple of bugs.
- Searching comments (may not be in the initial PR, we'll see)
- Searching instances (definitely won't be in this PR, we need an instance screen first anyway)


Mlem/Views/Tabs/Search/Components/SearchSectionHeaderView.swift
```diff
@@ -0,0 +1,65 @@
...
+//                                Divider()
+//                            }
+//                    }
+//                }
+//            )
+            .foregroundStyle(.primary)
+//            .onPreferenceChange(ViewOffsetKey.self) {
+//                // verify if position is zero (pinned) in container coordinates
+//                self.pinned = Int($0) <= 0
+//            }
```
EricBAndrews, line 62: Bunch of dead code in here

Mlem/Views/Tabs/Search/SearchModel.swift
```diff
@@ -0,0 +1,320 @@
...
+    func setUsers(_ users: [APIPersonView]) async {
+        self.users = users
+    }
+    
+    func setPosts(_ posts: [APIPostView]) async {
+        self.posts = posts
+    }
+}
+
+class SearchModel: ObservableObject {
```
EricBAndrews, line 46: *nit*

This is a model, so it shouldn't live under the `Views` tree

Mlem/Views/Tabs/Search/SearchView.swift
```diff
@@ -0,0 +1,244 @@
...
+        @AppStorage("internetSpeed") var internetSpeed: InternetSpeed = .fast
+        self._postTracker = StateObject(wrappedValue: .init(internetSpeed: internetSpeed))
+    }
+
+    var body: some View {
+        NavigationStack(path: $navigationPath) {
+            content
+                .background(Color(UIColor.systemGroupedBackground))
+                .handleLemmyViews()
+                // .navigationBarColor()
```
EricBAndrews, line 47: *nit*

Dead code? Forgot to re-enable?

Mlem/Views/Tabs/Search/SearchView.swift
```diff
@@ -0,0 +1,244 @@
...
+                    
+                    if !searchModel.suggestedFilters.isEmpty {
+                        SearchFilterListView(
+                            filters: searchModel.suggestedFilters,
+                            active: false,
+                            shouldAnimate: !searchModel.activeFilters.isEmpty
+                        )
+                        .padding(.bottom, 6)
+                    }
+                    LazyVStack(spacing: 10) { // pinnedViews: [.sectionHeaders]
```
EricBAndrews, line 140: *nit*

Dead code

Mlem/Views/Tabs/Search/SearchView.swift
```diff
@@ -0,0 +1,244 @@
...
+//                        }
+//                        .task(id: searchModel.page, priority: .userInitiated) {
+//                            print("TASK", searchModel.page, searchModel.loadedPage)
+//                            do {
+//                                try await searchModel.loadMore()
+//                            } catch {
+//                                errorHandler.handle(error)
+//                            }
+//                        }
+                }
```
EricBAndrews, line 188: Dead code

Mlem/Views/Tabs/Search/SearchView.swift
```diff
@@ -0,0 +1,244 @@
...
+                .coordinateSpace(name: "searchArea")
+                .fancyTabScrollCompatible()
+                .safeAreaInset(edge: .top, spacing: 0) {
+                    VStack {
+                        searchBar
+                            .background(.bar.opacity(atTopOfScrollView ? 0 : 1))
+                        //                if !atTopOfScrollView {
+                        //                    Divider()
+                        //                }
+                    }
```
EricBAndrews, line 198: Dead code

EricBAndrews (changes requested):
Bunch of dead code that can be removed. It would also be nice for maintainability to have a couple comments, especially in the model section

EricBAndrews:
Scrolling performance on searches is also pretty wonky. If you get a bunch of results and scroll, it gets very inconsistent in what posts get rendered--you can scroll for a long time down, then back up super fast.

Overall looks really good though!

Sjmarf:
> Bunch of dead code that can be removed.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               

Yep, I'm aware :) As I mentioned, there's still some stuff I needed to fix. Some of that is code that sort-of works for loading more pages, but is too buggy atm.

> It would also be nice for maintainability to have a couple comments, especially in the model section 

Sure 👍

Sjmarf:
> If you get a bunch of results and scroll, it gets very inconsistent in what posts get rendered--you can scroll for a long time down, then back up super fast.

Yeah. I have a feeling this is caused by the fancy highlighted-text, which does its calculations using .onAppear. Maybe it's not worth it.

Sjmarf:
> I haven't had time to review this yet but I notice in the description there is mention of searching instances - would instance search make sense in this area?
> 
> The user has selected an instance by this stage, we are looking to add some search functionality on instances but I believe this would be restricted to the onboarding / guest journeys before entering the app at which point they have access to _this_ search which is using the current instances API to perform searches for communities/posts/users/comments etc?

Of course, it's the least likely category to want to search. It would be all the way at the bottom below every other category. When we implement instance pages, it would make sense to include information about how many users the instance has, as well as potentially an uptime graph if we can find a third-party API for it. If we end up having functionality like that, it's reasonable that some people might want to see this information for instances that they aren't a part of. Hence the search-ability. But yeah, not a priority.

Sjmarf:
> I couldn't picture what would happen when the user tapped an instance result, but general information about it makes sense... once guest mode is in we _could_ let people enter guest mode for the instance from that page, too 😱 

Indeed! Maybe a "sign up" button could go on the instance page if the user is in guest mode too?

The instance sidebar is another bit of information that we don't currently display anywhere, so obviously that'd go in the instance page too. Also an admin list

Sjmarf:
Will be replaced by an upcoming PR

PR #571, opened 2023-09-04
Allow Duplicate Interaction Bar Widgets

Sjmarf:
Still need to make the save system better, though... it still saves only when you press "back", and not when you switch tab


EricBAndrews (approved):
🚀

PR #572, opened 2023-09-05
made info stack less broken with many widgets

EricBAndrews:
Two small changes to how the interaction bar works:

- Widgets are now assigned id based on offset (prevents id conflicts with multi-widgets)
- InfoStack now bleeds behind widgets if out of space
Screenshot 2023-09-04 at 9 11 44 PM


boscojwho (approved):
⬆️ ⬆️ ⬆️

PR #577, opened 2023-09-05
Community & User avatar refactor

Sjmarf:
Moved avatar-rendering functionality to new `CommunityAvatarView` and `UserAvatarView` structs. This makes it possible to display the avatars without displaying the whole `CommunityLabelView` / `UserLabelView` in future.

Also added options to clip avatars to squares instead of circles under Settings. This may be preferable to some users because it allows you to see more of the image, and matches the appearance of lemmy-ui. Adding this option required rebuilding the default avatar system to support both circles and squares, meaning that the default avatars look different now. Let me know if you'd like me to make any changes to the way these look.


Mlem/Views/Shared/Links/User/UserAvatarView.swift
```diff
@@ -0,0 +1,73 @@
+//
+//  UserAvatarView.swift
+//  Mlem
+//
+//  Created by Sjmarf on 05/09/2023.
+//
+
+import SwiftUI
+
+struct UserAvatarView: View {
```
EricBAndrews, line 10: While we're refactoring, do we need this at all? Can we achieve what we need with a single `AvatarView`? It would have to take in a couple parameters to handle things like default avatar icon and clipping, but overall it would cut down on code duplication by a lot and ensure that our avatars are always styled the same way

EricBAndrews (changes requested):
One comment about further refactoring, also some design questions I've raised in the Slack

Mlem/Views/Shared/Links/User/UserAvatarView.swift
Sjmarf: Sure, I can do it that way 👍

Mlem/Views/Shared/Links/Community/CommunityAvatarView.swift
```diff
@@ -0,0 +1,91 @@
...
+    return !clipOptOut.contains(hostString)
+}
+
+func shouldClipAvatar(url: URL?) -> Bool {
+    guard let hostString = url?.host else {
+        return true
+    }
+
+    return !clipOptOut.contains(hostString)
+}
```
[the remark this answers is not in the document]
Sjmarf: Done 👍

Sjmarf:
I've made the requested changes - let me know what you think. At some point, it would be nice to use `AvatarView` for the avatars on user profiles and the community sidebar. However, doing so would require a lot more refactoring because of the way those views work, so I think I'll leave it for now and come back to it later.

EricBAndrews (approved):

PR #584, opened 2023-09-08
fix - access accounts tracker via dependency

mormaer:
# Checklist
      - No issue, just noticed it and it'd crash if we ever access it 🙈 

# Pull Request Information

## About this Pull Request

This (tiny) PR just updates the `DeleteAccountView` to access the accounts tracker via `@Dependency` instead of `@EnvironmentObject`. The change made in #554 means that there is no accounts tracker in the environment so attempting to access it this way will crash the app.

This has snuck through as the work in #554 happened on `dev` before the delete account changes from `master` were re-integrated the other day in #573.


JakeShirley (approved):
Good catch! Wish this could be found at compile time some how.

PR #587, opened 2023-09-09
introduce `AppFlow` to support different application flows

mormaer:
# Checklist
      - Related to #21
      - Resolves a separate issue where multiple instance of `AppState` could be in memory 😱 

# Pull Request Information

## About this Pull Request
At present the application start-up logic is based around the concept of us either having a default account or not, and so decisions about what to render were based on the presence of a non-optional account. The changes in this PR swap out this logic to instead work on an `AppFlow` enumeration which currently only has `.account(SavedAccount)` or `.onboarding` as its cases.

**These two cases continue to represent the previous state of having an account, or _not_ having an account. So our behaviour remains much as it was.**

However, there are a few things that these changes now allow:

1. We can add additional application flow states, eg `.guest(URL)`, which will (soon) link up with the work done in #567 so that we can create `.unauthenticated` sessions when the app enters guest mode.
2. The `AppState` is now back to having no arguments passed to its `.init` and is now stored as an `@StateObject`. This resolves a separate issue I found on another branch where we were ending up with more than one of our app states in memory... which is _less than ideal_ 😭 
3. Allows the app to work without the presence of an account, as the `.currentActiveAccount` is now an optional value there are various changes in this PR to handle that optionality. Note - there will need to be more work done here, I've done the bare minimum required in this PR as although the account now supports optionality we have no way to enter the application without an account... that is still to come with the work for #21.

## Screenshots and Videos
There are no major changes to the UI in this PR, however I noticed that our current App Store build shows an icon next to the add account button which had disappeared on `dev`, so I've added that back.

One noticeable thing that the new flow does introduce is a bit of _jank_ when you add a new account while already inside the app, the dismiss animation can misbehave at times. There are some existing `DispatchQueue` tricks being done to reduce this elsewhere so I will see if I can make it _pretty_ as part of the upcoming onboarding/guest mode work.

## Additional Context
One other thing I noticed while working on this PR is that the way we determine if a post/comment etc belongs to the active account needs some work. We make a comparison on the `.id` of the account, however there is a (small?) risk that we incorrectly identify something as belonging to us when it doesn't since `.id` will only be unique on our home instance - not across the entire Fediverse.

We could/should potentially be using the `.actorId` or such, but I've not tried to fix this as part of this PR to avoid it growing any larger than it needs to be. I've left a `TODO` in the code for it.

### DIFF
--- a/Mlem/AppDelegate.swift
+++ b/Mlem/AppDelegate.swift
@@ -11,7 +11,7 @@ import UIKit
 
 // TODO: we need to do a bit of work to ensure we also switch tab when responding to these
 // as currently it launches you into the app, but if the app was already running you're left
-// on the tab/screen you were on - despite the shortcuts being design to take you to the feeds
+// on the tab/screen you were on - despite the shortcuts being designed to take you to the "Feeds" tab
 var shortcutItemToProcess: UIApplicationShortcutItem?
 
 class AppDelegate: UIResponder, UIApplicationDelegate, UIWindowSceneDelegate {
--- a/Mlem/AppFlow.swift
+++ b/Mlem/AppFlow.swift
@@ -12,6 +12,6 @@ import Foundation
 enum AppFlow: Equatable {
     /// The onboarding flow
     case onboarding
-    /// An 'signed in'' session with the users `SavedAccount` as an associated value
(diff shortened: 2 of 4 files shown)


Sjmarf (approved):
LGTM 👍

PR #593, opened 2023-09-10
Allow display of user avatar in tab bar

mormaer:
# Checklist
      - just a nice QoL improvement

# Pull Request Information

## About this Pull Request
This PR adds logic to store the users avatar URL in the `SavedAccount` object and displays it on the tab bar. If the image fails to download it will fall back to the default symbol. There is also some logic that will update the stored value whenever we fetch the users profile... I've left a `TODO` to look at moving this into the repository layer longer term, but that is a larger project where we look to make the repositories capable of powering things in a more reactive way.

I've added a new option in the related settings screen so the user can choose to disable it displaying, and if they have set their display style to `.anonymous` it is always disabled.

## Screenshots and Videos
| DARK | LIGHT |
| --- | --- |
| Screenshot 2023-09-10 at 21 09 20 | Screenshot 2023-09-10 at 21 08 39 |

| STANDARD | ANONYMOUS |
| --- | --- |
| Screenshot 2023-09-10 at 21 09 56 | Screenshot 2023-09-10 at 21 10 15 |

## Additional Context
I noticed while making sure the avatar on the tab bar updates that when refreshing the `UserView` if the users avatar has been removed on web, or a new one added it does not update. I know you're doing some work on that view current @JakeShirley so it could be worth checking that? 🤞

### DIFF
--- a/Mlem/App State.swift
+++ b/Mlem/App State.swift
@@ -15,6 +15,7 @@ class AppState: ObservableObject {
     
     @AppStorage("defaultAccountId") var defaultAccountId: Int?
     @AppStorage("profileTabLabel") var profileTabLabel: ProfileTabLabel = .username
+    @AppStorage("showUserAvatarOnProfileTab") var showUserAvatar: Bool = true
     
     @Published private(set) var currentActiveAccount: SavedAccount?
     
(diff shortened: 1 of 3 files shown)


JakeShirley (approved):
Thanks for the call out, I'll take a look!

PR #604, opened 2023-09-14
SiteVersion enum

Sjmarf:
Implements a `SiteVersion` enum to store the version number of a site. `SiteVersion` can be initialised from a String, and is Comparable. I've added an instance of `SiteVersion` to `SiteInformationTracker`, but it isn't accessed anywhere yet. `SiteVersion` supports the semver `"x.x.x"` format, as well as the `"x.x.x-abc.x"` format used for betas and release candidates. If a version number doesn't conform to this format, an `other` case will be used.

## Why do we need this?

If Lemmy implements some new feature, we might want to check that the user's site version supports that feature. Or, different backend versions may have differences in the API. The `SiteVersion` enum is Comparable, making it very easy for us to guard against possible API errors caused by instances running on older versions.

## Example usage

```swift
// To check that the user's instance is running 0.18.2-beta.1 or later
if siteTracker.version >= SiteVersion.suffixed(major: 0, minor: 18, patch: 2, suffix: .beta(1) {
    // ...
}
```
```swift
// This is equivalent to the previous example but shorter
if siteTracker.version >= SiteVersion("0.18.2-beta.1") {
    // ...
}
```
```swift
// To render the version number as text in a View
Text(String(describing: siteTracker.version))
```


JakeShirley (commented):
Is it worth using a known-good SemVer library such as this? https://github.com/ddddxxx/Semver

Mlem/API/Internal/SiteVersion.swift
```diff
@@ -0,0 +1,200 @@
...
+            return false
+        case (.beta, .other), (.releaseCandidate, .other):
+            return true
+        default:
+            return false
+        }
+    }
+}
+
+enum SiteVersion: Equatable {
```
JakeShirley, line 58: Semantic version (semver.org) may be a better name here

Mlem/API/Internal/SiteVersion.swift
```diff
@@ -0,0 +1,200 @@
...
+            if let int = int {
+                return "\(string).\(int)"
+            }
+            return string
+        }
+    }
+}
+
+extension SiteVersionSuffix: Comparable {
+    static func < (lhs: SiteVersionSuffix, rhs: SiteVersionSuffix) -> Bool {
```
JakeShirley, line 32: FWIW semver just does alphanumeric sorting here https://semver.org/#spec-item-9

JakeShirley (approved):
Is it worth using a known-good SemVer library such as this? https://github.com/ddddxxx/Semver

Mlem/API/Internal/SiteVersion.swift
```diff
@@ -0,0 +1,200 @@
...
+
+enum SiteVersion: Equatable {
+    
+    /// Format 0.18.2
+    case release(major: Int, minor: Int, patch: Int)
+    /// Format 0.18.2-beta.5, 0.18.2-rc.2
+    case suffixed(major: Int, minor: Int, patch: Int, suffix: SiteVersionSuffix)
+    case other(String)
+    case zero
+    case infinity
```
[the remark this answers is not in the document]
Sjmarf: That is correct. They're designed to be used outside of the SiteVersion in situations such as this 

```swift 
    var minimumVersion: APISiteVersion {
        switch self {
        case .topThreeMonth, .topSixMonth, .topNineMonth:
            return SiteVersion("0.18.1")
        default:
            return .zero
        }
```

Of course, there are are other ways to avoid this problem, but adding zero and infinity seemed like the most logical to me

Mlem/API/Internal/SiteVersion.swift
Sjmarf: Yeah, perhaps. I wasn't sure about lemmy.blahaj.zone's `-kt` identifier though - I'm not sure what it is, so I'm assuming it's blahaj's custom modified version of Lemmy? If that's the case, it should be considered higher than the main version. In contrast, Betas and RCs are considered lower than the main version, since they are pre-releases

JakeShirley: Yeah this is one of the examples they have:
`Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.`

Sjmarf:
> If we focussed only on the `major.minor.patch` aspect of the semver, it'd reduce the complexity a fair bit, and would likely give us what we need to target differences between stable releases?

Yeah, I could do. So would we count all suffixed `x.x.x-beta.1` versions as being later than the base version?

Mlem/API/Internal/SiteVersion.swift
Sjmarf: `1.0.0-kt.1` should be greater than `1.0.0` in our case - but using the server sorting system, it would be less than `1.0.0`.

Mlem/API/Internal/SiteVersion.swift
```diff
@@ -0,0 +1,200 @@
...
+//  APISiteVersionNumber.swift
+//  Mlem
+//
+//  Created by Sjmarf on 09/09/2023.
+//
+import Foundation
+
+enum SiteVersionSuffix: Equatable {
+    /// The only type other than beta "-beta.x" and releaseCandidate "-rc.x" that I've seen
+    /// is lemmy.blahaj.zone's "-kt.x". I'm unable to determine the meaning of this.
```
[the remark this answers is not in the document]
Sjmarf: Ah, i missed those The system will be able to decode all of them aside from the `+` one. I'll look into fixing that if we keep the suffix system

Sjmarf:
> Generally I think we'd want to stick to targeting stable versions only, even if that meant something new like the sorting examples above was released a few weeks after it first became available 🤷 

Ok, I'll strip it back to only use the base `x.x.x` format 👍

Mlem/API/Internal/SiteVersion.swift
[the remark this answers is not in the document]
Sjmarf: `.infinity` would be useful if we need to check if a server is running a version less than a set version - the opposite of the example code I posted above.

Sjmarf:
Done. The suffix is ignored now 👍

Mlem/Views/Tabs/Feeds/Feed View.swift
```diff
@@ -151,7 +151,7 @@ struct FeedView: View {
         } else {
             VStack(alignment: .center, spacing: 5) {
                 Image(systemName: "text.bubble")
-                
+                Text(String(describing: SiteVersion("0.18.2")))
```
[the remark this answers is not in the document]
Sjmarf: Yep! good catch

Issue #607, opened 2023-09-14
* feeds don’t load with latest lemmy version (world et al)


th3dogcow:
**Describe the bug**
Lemmy servers upgraded

**Expected Behavior**
Fred’s (local, all, subscribed) to load

**Actual Behavior**
“no posts found”

**To Reproduce**
What steps do I need to take to trigger the bug?

**Mlem Information:**
1.0.2 (605)

EricBAndrews:
Fixed in PRs mentioned above

PR #612, opened 2023-09-15
fix - handle unencoded urls coming from API

mormaer:
# Checklist
      - partially addresses #607 

# Pull Request Information

## About this Pull Request
This PR resolves an issue found on `lemmy.world` (but would happen on any instance) that caused our modelling to fail when the API returns a URL which is not percent encoded.

There is a stickied post at the top of the `lemmy.world` all feed including a link to their Matrix space (`https://matrix.to/#/#space:lemmy.world`) The `#`'s here are not percent encoded and so trying to form it into a `URL` makes Swift unhappy 🙈 

The change in this PR introduces a new `LemmyURL` struct which will attempt to form a `URL` from the supplied string without any changes, if that process fails it will apply percent encoding and re-try. If the second step fails, it will fail the decoding as before.

I've gone through and changed the places where we were previously modelling values as `URL/URL?` in the API to now decode the raw `String/String?` values and then added computed properties which use the new struct to create the url when requested.

I have not adjusted the `.actorId` values and have left those as `URL` for now. We could alter those to work in the same way, but I do not expect (famous last words) Lemmy to return urls that require percent encoding within that value.

Given the above about the `.actorId`, applying this fix across all of the various URLs is likely being over-cautious but this way we are protected should any unencoded URLs sneak through in other calls.

## Screenshots and Videos

This shows the result before and after the fix is applied.

| BORKED | UNBORKED |
| --- | --- |
| <video src="https://github.com/mlemgroup/mlem/assets/5231793/0f0d4486-5437-45bc-9df6-78539500b9d5"> | <video src="https://github.com/mlemgroup/mlem/assets/5231793/90ea8785-8ff7-404b-8a49-f5597c1f60d0"> |


Mlem/API/Models/Posts/APIPost.swift
```diff
@@ -28,10 +28,15 @@ struct APIPost: Decodable {
     let nsfw: Bool
     let published: Date
     let removed: Bool
-    let thumbnailUrl: URL?
+    let thumbnailUrl: String?
     let updated: Date?
 }
 
+extension APIPost {
+    var linkUrl: URL? { LemmyURL(string: url)?.url }
```
JakeShirley, line 36: This is a clever solution!

JakeShirley (approved):

Issue #534, opened 2023-08-21
Onboarding Enhancements


EricBAndrews:
Onboarding flow can be improved:

- Move all steps into the app itself (remove redirection out to web)--captcha can be fetched from the instance!
- Add search to instance browser
- Add timed refresh to instance browser (~1 week? instances don't change much)
- Add robustness against awesome-Lemmy-instances changing format
- Add general robustness (https://github.com/mlemgroup/mlem/pull/533#discussion_r1302093586, https://github.com/mlemgroup/mlem/pull/533#discussion_r1302095419)

PR #613, opened 2023-09-16
Restructured onboarding

mormaer:
# Checklist
      - Partially addresses #534 

# Pull Request Information

## About this Pull Request
This PR makes some changes to our onboarding to split the new and existing user flows into separate navigation stacks rather than a single tab view. This was discussed in the dev chat the other week and agreed.

In addition to this there is a basic _transition_ flow now shown when the application changes its flow. This is mostly to hide the janky-jank that happens when we're trying to do things like dismissing modals but at the same time we want to rebuild the applications views from the root.

The view itself is basic and just displays a spinner and some text, we can iterate on this and make it _fancy_ if we like, however it will only be shown momentarily so I've not gone any further with it in this PR.

I've also removed the _What is Lemmy?_ link from the add account screen in this PR. With onboarding now present the two ways this screen is accessed is either from inside the application where the user already has an account (so knows what Lemmy is) or via the new user flow which has screens dedicated to explaining Lemmy/Federation.

## Screenshots and Videos
| NEW FLOW |
| --- |
| <video src="https://github.com/mlemgroup/mlem/assets/5231793/5a762ff8-d42f-49d8-a54c-b6018b44f2e4"> |

## Additional Context
There is still an issue present in onboarding due to the use of our `LazyVStack` inside the instances view, which causes problems if you scroll up/down the list. We have discussed limiting the number of instances shown to ~30, allowing us to use a `VStack` which resolves this problem.

I'll do that, and add basic search functionality to the instances in my next PR.

### DIFF
--- a/Mlem/Views/Shared/Accounts/Accounts Page.swift
+++ b/Mlem/Views/Shared/Accounts/Accounts Page.swift
@@ -33,15 +33,14 @@ struct AccountsPage: View {
                         Section(header: Text(instance)) {
                             ForEach(accountsTracker.accountsByInstance[instance] ?? []) { account in
                                 Button(account.nickname) {
-                                    dismiss()
                                     setFlow(using: account)
+                                    dismiss()
                                 }
                                 .disabled(isActiveAccount(account))
                                 .swipeActions {
                                     Button("Remove", role: .destructive) {
-                                        dismiss()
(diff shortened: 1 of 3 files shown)


Sjmarf (approved):
LGTM 👍

PR #624, opened 2023-09-18
bumped version

EricBAndrews:
Bumped version to 1.1 to reflect current working RC


JakeShirley (approved):
A huge PR.  In the future can we please break this out into multiple PRs?

:shipit:

Issue #602, opened 2023-09-14
New post composer destroys composed post without confirmation on down-swipe


mbs:
**Describe the bug**
The "New Post" composer destroys the post without confirmation on down-swipe.

**Expected Behavior**
Either nothing, or at least confirm that the draft post should be lost.

**Actual Behavior**
Post was lost.

**To Reproduce**
Feeds -> click a subscribed community -> Three dots [...] -> New Post -> painstakingly compose a longpost (actually any inputs will do) -> barely swipe down anywhere in the post popup/tab (e.g. while attempting to reposition the carat in the body, or on the titlebar, etc)

**Mlem Information:**
1.0.1 (483)

**Additional Context**
n/a

boscojwho:
Improved in #633
- Can no longer swipe to dismiss if there's content in post composer.

PR #633, opened 2023-09-21
Editor sheet detents

boscojwho:
# Checklist
      - #602 
      - User feedback: Want to read what they're responding to while create/editing.

# Pull Request Information

## About this Pull Request
This PR includes two UI changes to Post editor and Response editor when presented as sheets:
1. Adds `.medium` presentation detent, while allowing users to interact with the view behind it while in `.medium` detent.
- Note: Interaction with view behind is only available on iOS 16.4 or later.
2. Disable interactive dismissal when editor has content.

## Additional Notes
Sheet initially presents at `.medium` detent before moving to `.large` on initial presentation:
- To immediately present sheet in `.large`, we should pass a `selection` binding to `.presentationDetents(...)`, but doing so causes the content view to call into its body, resulting in visible lag when switching between detents.  
- To workaround this animation hitch, we could move the detent binding into each of the composer views, but we would also need to move the sheet's NavigationStack into the composers view, as well, since `.sheet(...)` expects presentation detent modifiers to be applied on the root view.

## Screenshots and Videos

https://github.com/mlemgroup/mlem/assets/2549615/d7a846fa-5d18-42f9-ba49-8bfe3d9d1769

https://github.com/mlemgroup/mlem/assets/2549615/ac67551f-fcea-44df-9d25-5468a0085248


EricBAndrews (approved):

EricBAndrews:
Merging per 12hr principle

PR #634, opened 2023-09-21
Improved Use of SFSymbols

### LINKED ISSUE #48: Unified symbol handling
All our symbols should be stored in AppConstants (or somewhere like that) so we can guarantee consistency and change them at will.

Symbols should also be stored only once, and use `.symbolVariant(.fill)`, `.symbolVariant(.circle)`, etc. to render variants.

EricBAndrews:
# Checklist
      - closes #48 

# Pull Request Information

This PR does three main things:
- Moves all symbol name constants from `AppConstants` to `Icons`
- Defines constants for (hopefully?) all icons used and replaces raw strings with icon constants
- Renames icon constants to be simpler, more consistent, and more informative

It additionally does some minor stuff:
- Simplifies overcomplicated profile and community avatar defaults to use appropriate SFSymbols
- Removes final "symbol map constant" from Icons


boscojwho:
Profile icon look different, ignore if intentional.

| Before  | After |
| ------------- | ------------- |
|  |  |

EricBAndrews:
The user icon is an intentional change--the old one had scaling issues when the instance was hidden and was just a *lot* of code to do basically the same thing as a single symbol. 

Jump button was a mistake, though, I'll have that fixed momentarily

Sjmarf:
> The user icon is an intentional change--the old one had scaling issues when the instance was hidden and was just a lot of code to do basically the same thing as a single symbol.

Hmm. I much prefer the one we have currently - it more closely matches apple's own user profile image.

Mlem/Enums/FeedType.swift
```diff
@@ -26,17 +26,17 @@ enum FeedType: String, Encodable, SettingsOptions {
 extension FeedType: AssociatedIcon {
     var iconName: String {
         switch self {
-        case .all: return AppConstants.federatedFeedSymbolName
-        case .local: return AppConstants.localFeedSymbolName
-        case .subscribed: return AppConstants.subscribedFeedSymbolName
+        case .all: return Icons.federatedFeed
```
Sjmarf, line 29: Is this meant to be `federatedFeedFill`? Or no?

Sjmarf (approved):
Otherwise, lgtm 👍

PR #636, opened 2023-09-22
fix - missing limit on post tracker refresh

mormaer:
# Checklist
      - No issue raised for this in GH

# Pull Request Information

## About this Pull Request
This fixes issues resulting in missing posts after the user applies a different sorting option or pulls to refresh on the main/community feeds.

Credit to @Sjmarf for spotting it 🙌 

## Screenshots and Videos
No UI changes in this PR.

## Additional Context
I've made the `limit` value non-optional as part of this PR even though it is an optional argument on the API. The reason being that when it comes to the post tracker, we will always want to supply a `limit` as forgetting to do so breaks the pagination.

I also removed a few default arguments from the request layer so that they are required to be passed down from the repositories, again just to increase visibility of the values so they're not inadvertently ignored in the future.

### DIFF
--- a/Mlem/API/APIClient/APIClient+Post.swift
+++ b/Mlem/API/APIClient/APIClient+Post.swift
@@ -8,14 +8,15 @@
 import Foundation
 
 extension APIClient {
+    // swiftlint:disable function_parameter_count
     func loadPosts(
         communityId: Int?,
         page: Int,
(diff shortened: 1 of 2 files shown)


Sjmarf (approved):
Nice 👍

EricBAndrews (approved):
🚀

PR #649, opened 2023-09-24
Editor sheet detents animation fix

Sjmarf:
Fixed a bug where the "new post" sheet and comment editor sheet would appear at the `.medium` presentation detent initially before bouncing up to `.large`. Now, it's `.large` straight away.


boscojwho:
Of course, just pass in a constant 🤦‍♂️

PR #680, opened 2023-09-30
Fixed swipey action bad behavior with context menus

### LINKED ISSUE #621: Mask swipey actions bad behavior in comments
Swipe actions sometimes try to also trigger context menu. On posts this renders as lag, in comments you see the duplication from the two animations--comments should also just "lag"

EricBAndrews:
# Checklist
      - closes #621 

# Pull Request Information

This PR addresses swipey actions and context menus conflicting. Previously, an aborted context menu pop that triggered the animation would invoke the drag handler of the swipey action, leading to the swipey action either jumping to the drag state or, worse, duplicating the view and showing them one atop the other. All .contextMenu view modifiers have been moved to occur *after* swipey action view modifiers, ensuring that the context menu takes priority.

### DIFF
--- a/Mlem/Extensions/Swipey Actions.swift
+++ b/Mlem/Extensions/Swipey Actions.swift
@@ -348,6 +348,12 @@ struct SwipeyView: ViewModifier {
 // swiftlint:enable function_body_length
 
 extension View {
+    /// Adds swipey actions to a view.
+    ///
+    /// NOTE: if the view you are attaching this to also has a context menu, add the context menu view modifier AFTER the swipey actions modifier! This will prevent the swipey action from triggering and appearing bugged on an aborted context menu pop if the context menu animation initiates.
+    /// - Parameters:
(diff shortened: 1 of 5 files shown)


boscojwho:
Attempt to review, but not sure how to get the context menu to abort in a way that causes this bug =/

Also, was this an issue on iOS 16 as well? I've heard of quite a few view-modifier related regressions on iOS 17.

EricBAndrews:
Ah, yes, I suppose I should have included better repro instructions. You have to hold the post for just a moment so it *thinks* you want the context menu, then drag to trigger the swipey action. On the current dev build this will snap the post to your finger after a short delay; this PR (in theory) disables that behavior. 

It's actually much better on 17 than 16--16 still does the snap, but that one is more or less unfixable. This PR at least makes the behavior consistent across all swipey views on 16.

EricBAndrews:
Video for reference 

https://github.com/mlemgroup/mlem/assets/44140166/ca08f5bd-cc8d-4659-841a-d536c05109ab

PR #687, opened 2023-10-02
Fixed profile refresh problems

### LINKED ISSUE #327: Pulling down to refresh profile tab does not refresh post upvote counts
You have to restart the app to see updated upvote counts. I'm not sure whether this also applies to comments.

### LINKED ISSUE #686: Cannot refresh own profile
Per title. Refreshing own profile causes error.

EricBAndrews:
# Checklist
      - closes #686 
      - closes #327 

# Pull Request Information

This PR fixes two problems with the profile:
1. Refreshing own profile always throws an error. This was caused by the refresh function opportunistically trying to update the app state with the new profile information, which was in turn causing a view rerender and therefore cancelling the saved items load task. I have moved that logic to happen after saved items are loaded and added a comment re: long-term opportunistic profile refresh
2. Upvote counts don't update when refreshing. This was caused by the profile simply trying to *add* the new (refreshed) items to the trackers, which led to them all being rejected because they were dupes. I changed the behavior to *reset* the trackers instead, which is fine because we never perform any load action other than reload (though that should also be changed later). I have also made the items `Hashable` based on the underlying content hash and given them a `ContentModelIdentifier` uid, which allows them to be rerendered appropriately.

For cleanliness, I also moved the loading calls into the person repository.

### DIFF
--- a/Mlem/API/Internal/HierarchicalComment.swift
+++ b/Mlem/API/Internal/HierarchicalComment.swift
@@ -39,6 +39,16 @@ extension HierarchicalComment: Equatable {
     }
 }
 
+extension HierarchicalComment: Hashable {
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(commentView.id)
+        hasher.combine(commentView.comment.updated)
+        hasher.combine(commentView.counts.upvotes)
+        hasher.combine(commentView.counts.downvotes)
+        hasher.combine(commentView.myVote)
+    }
+}
+
 extension [HierarchicalComment] {
     /// A method to insert an updated `APICommentView` into this array of `HierarchicalComment`
     /// - Parameter commentView: The `APICommentView` you wish to insert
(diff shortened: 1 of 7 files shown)


boscojwho:
Tested using test account on our test server on iPhone simulator:
- Nothing loads in `Comments` "tab".
- Switch to latest `/dev` build on same account and comments load.
- Switch back to this PR, and nothing loads.
- Pull-to-refresh succeeds, but nothing loads.

boscojwho:
It appears logic in `UserFeedView.generateCommentFeed(...)` is filtering out all comments.

EricBAndrews:
> It appears logic in `UserFeedView.generateCommentFeed(...)` is filtering out all comments.

It was actually a layer deeper--the method that was building the comment tracker was simply overwriting the authored comments with the saved comments. I've tidied that all up now, should be nice and functional (if badly in need of a proper rewrite)

boscojwho (approved):
Looks good:
- Comments now load.
- Votes update when refreshing.
Worth revisiting later on:
- Only `Posts` list automatically re-renders when votes change (e.g. tap into post from profile, change vote, go back, list already updated with new vote count). The other lists don't do this yet.

PR #691, opened 2023-10-03
New Sort Options

### LINKED ISSUE #321: Add new sort options
Lemmy have added yet more sort options:

image

In my opinion, it is a bad idea to simply add these to the current list or add another set of nested drop-downs. The menu is already a little big and cumbersome. Maybe a redesign is in order? Related: #140

One option is to add the Hour and Month sub-options horizontally within a menu row, sort-of like what Notes does here. I haven't prototyped this though so it might not work well in practice. 

image

What do people think the best way to do this is?

### LINKED ISSUE #673: [0.19.0] - Scaled sort option
## Introduction
As discussed in the recent Lemmy announcement a new sorting option `scaled` has been added which becomes available from version `0.19.0` and above.

> A new scaled sort option has been added. This sort is identical to the Hot sort, but also takes into account the number of each community’s active monthly users, and so helps to boost posts from less active communities to the top.

> https://github.com/LemmyNet/lemmy/pull/3907

## Requirements
We should look to add support for this ASAP, the option will need to be gated behind a version check as it will not be available on previous versions.

Sjmarf:
Adds five new sort options:

- `scaled`
- `controversial`
- `top 3 months`
- `top 6 months`
- `top 9 months`

`scaled` and `controversial` are 0.19.0+ whilst the new `top` modes are available on the latest release version. The 0.19.0+ sort modes are hidden on instances that aren't running 0.19.0 or above.

Potential problem with this implementation:

You can still set the default sort mode to a mode that isn't available on your instance. If the user chooses an unsupported sort mode as the default, it would obviously be selected in the feed and they would see an error. I _could_ filter out the modes that aren't available from the default sort-mode picker - this doesn't remove the problem though, because the user could set the default sort mode on a 1.19.0 instance and then switch accounts to an older instance, and they would still see an error. 

Closes #673
Closes #321

### DIFF
--- a/Mlem/Enums/Settings/PostSortType.swift
+++ b/Mlem/Enums/Settings/PostSortType.swift
@@ -6,12 +6,17 @@
 //
 
 import Foundation
+import Dependencies
 
 enum PostSortType: String, Codable, CaseIterable, Identifiable {
+    @Dependency(\.siteInformation) static var siteInformation
+    
     case hot = "Hot"
     case active = "Active"
     case new = "New"
     case old = "Old"
+    case scaled = "Scaled"
+    case controversial = "Controversial"
     case newComments = "NewComments"
     case mostComments = "MostComments"
     case topHour = "TopHour"
@@ -20,13 +25,55 @@ enum PostSortType: String, Codable, CaseIterable, Identifiable {
     case topDay = "TopDay"
     case topWeek = "TopWeek"
     case topMonth = "TopMonth"
+    case topThreeMonth = "TopThreeMonth"
+    case topSixMonth = "TopSixMonth"
+    case topNineMonth = "TopNineMonth"
     case topYear = "TopYear"
     case topAll = "TopAll"
     
     var id: Self { self }
     
-    static var outerTypes: [PostSortType] { [.hot, .active, .new, .old, .newComments, .mostComments] }
-    static var topTypes: [PostSortType] { [.topHour, .topSixHour, .topTwelveHour, .topDay, .topWeek, .topMonth, .topYear, .topAll] }
+    static var outerTypes: [PostSortType] { [
+        .hot,
+        .scaled,
+        .active,
+        .new,
+        .old,
+        .newComments,
+        .mostComments,
+        .controversial
+    ] }
+    
+    static var topTypes: [PostSortType] { [
+        .topHour,
+        .topSixHour,
+        .topTwelveHour,
+        .topDay,
+        .topWeek,
+        .topMonth,
+        .topThreeMonth,
+        .topSixMonth,
+        .topNineMonth,
+        .topYear,
+        .topAll
+    ] }
+    
+    static var availableOuterTypes: [PostSortType] { filterTypes(outerTypes) }
+    static var availableTopTypes: [PostSortType] { filterTypes(topTypes) }
+    
+    private static func filterTypes(_ types: [PostSortType]) -> [PostSortType] {
+        guard let siteVersion = siteInformation.version else { return types }
+        return types.filter { siteVersion >= $0.minimumVersion }
+    }
+    
+    var minimumVersion: SiteVersion {
+        switch self {
+        case .controversial, .scaled:
+            return .init("0.19.0")
+        default:
+            return .zero
+        }
+    }
     
     var description: String {
         switch self {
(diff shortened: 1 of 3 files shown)


Sjmarf:
> What version is the latest release version here? is it worth limiting the new top modes, the current instance list still shows a fair amount of instances on 0.18.1-0.18.3?

Couldn't remember when I was writing this, but I implemented it a filter for it in my other sort-mode PR and it seems they're new in 0.18.1. I can add a filter for it, sure

Sjmarf:
> We have a few weeks till this needs to be in so we can have a think of how to handle it... this will be a common problem going forward for other versioned values so we should try and come up with a nice solution that will solve it for future problems too 🙏

The best solution I could think of is to select the similar sort mode to the default if the default is unavailable. Scaled would become hot, controversial would become most comments, etc

EricBAndrews (changes requested):
Code looks good--per Slack, let's go with offering a "fallback sort" option if the user selects a sorting option that is not supported by all instances.

Sjmarf:
Sure 👍

Sjmarf:
Done

Screenshot 2023-10-21 at 12 42 22

EricBAndrews (changes requested):
Looks good! I really like moving sorting into its own menu section.

I noticed one bug around initializing the fallback state. When I first launch the app, it takes the default value and never the fallback. It works on subsequent loads--if I switch accounts or navigate back to the community menu and back into the feed it picks up the fallback, but it looks like the fallback is computed after the FeedView is already rendered and it does not reload on change.

Sjmarf:
> I noticed one bug around initializing the fallback state. When I first launch the app, it takes the default value and never the fallback. It works on subsequent loads--if I switch accounts or navigate back to the community menu and back into the feed it picks up the fallback, but it looks like the fallback is computed after the FeedView is already rendered and it does not reload on change.

So, I tried to fix it and now it works 90% of the time... but occasionally it just won't for some reason and I can't work out why :(

Mlem/Views/Tabs/Feeds/Feed View.swift
```diff
@@ -39,30 +41,47 @@ struct FeedView: View {
...
         community: CommunityModel?,
         feedType: FeedType,
-        sortType: PostSortType,
         showLoading: Bool = false
     ) {
         // need to grab some stuff from app storage to initialize post tracker with
         @AppStorage("internetSpeed") var internetSpeed: InternetSpeed = .fast
         @AppStorage("upvoteOnSave") var upvoteOnSave = false
         
+        self.showLoading = showLoading
```
EricBAndrews, line 50: Is this needed? It doesn't look like it's read anywhere

Mlem/Views/Tabs/Feeds/Feed View.swift
```diff
@@ -39,30 +41,47 @@ struct FeedView: View {
...
+            if siteVersion >= defaultPostSorting.minimumVersion {
+                _postSortType = .init(initialValue: defaultPostSorting)
+            } else {
+                _postSortType = .init(initialValue: fallbackDefaultPostSorting)
+            }
+            isWaitingForSiteInformation = false
+        } else {
+            _postSortType = .init(initialValue: .hot)
+            isWaitingForSiteInformation = true
+        }
```
EricBAndrews, line 76:
```suggestion
        _postSortType = .init(initialValue: fallbackDefaultPostSorting)
        isWaitingForSiteInformation = true
```

Occasionally the feed ends up set to `.hot` and never updates thereafter. I think this is caused by siteInformation updating after the check here but before the view renders, leading to it not getting picked up by the .onChange handler. I've developed a fix that seems to work--see below for the other half of it

Mlem/Views/Tabs/Feeds/Feed View.swift
```diff
@@ -93,7 +112,26 @@ struct FeedView: View {
...
             .navigationBarColor(visibility: .visible)
             .environmentObject(postTracker)
-            .task(priority: .userInitiated) { await initFeed() }
+            .task(priority: .userInitiated) {
+                if !isWaitingForSiteInformation {
+                    await initFeed()
+                }
+            }
+            .onChange(of: siteInformation.version) { newValue in
+                if let siteVersion = newValue {
```
EricBAndrews, line 121:
```suggestion
            .task(id: siteInformation.version) {
                if let siteVersion = siteInformation.version {
```

Making this a .task ensures that it performs the first check once the view is properly rendered, ensuring that a published update can't slip through the cracks

EricBAndrews (changes requested):
Commented a suggested fix for those 10% of cases where `siteInformation` doesn't properly trigger updates

Sjmarf:
Thanks! Looks like it works consistently now.

EricBAndrews (approved):
Nice 👍

PR #703, opened 2023-10-06
Swipe to delete recent search

Sjmarf:
Swipe on a recent search to delete just that item from the list.


Mlem/Views/Tabs/Search/RecentSearchesView.swift
```diff
@@ -62,11 +72,16 @@ struct RecentSearchesView: View {
...
+                    CommunityResultView(
+                        community: community,
+                        showTypeLabel: true,
+                        hasSwipeActions: false
+                    )
                 } else if let user = contentModel.wrappedValue as? UserModel {
                     UserResultView(user: user, showTypeLabel: true)
                 }
             }
+            .addSwipeyActions(trailing: [deleteSwipeAction(contentModel)])
```
EricBAndrews, line 84: Can we just attach all the swipe actions up at this level? That would get rid of the need for the hasSwipeActions toggle and make the swipe action logic much more readable. `subscribe()` would have to get pulled up to here for that, but it appears that all of its required info is already present at this level.

That would also let us attach both a subscribe *and* a clear action to a recent search view, which feels like how it should be.

Mlem/Views/Tabs/Search/Results/CommunityResultView.swift
```diff
@@ -50,6 +52,14 @@ struct CommunityResultView: View {
     }
     
     var body: some View {
+        if hasSwipeActions {
+            content
+                .addSwipeyActions(trailing: [subscribeSwipeAction])
+        } else {
+            content
+        }
+    }
```
EricBAndrews, line 61: See comment above--it would be nice for this view to not come bundled with swipe actions

EricBAndrews (changes requested):

Mlem/Views/Tabs/Search/RecentSearchesView.swift
Sjmarf: My logic for doing it this way is that I'd need to do the same for posts in future, when we add post searching. 

Right now (I think?) the post's swipe actions are attached inside of `FeedPost`. Moving them out of `FeedPost` would mean implementing that logic in more than one place, since we'd need to attach the same swipe actions to `FeedPost` in both the search results and main feed. 

But, we'd need to disable these swipe actions to allow for swipe-to-delete in the recent searches list. My solution to this would be to add a `hasSwipeActions` parameter to `FeedPost` too, in much the same way that I've done here.

The user can still subscribe/unsubscribe to a community via the recent searches list using the context menu, by the way. I chose to replace the swipe action in this case because Apple does this in all of their recent search lists (spotlight, Apple Music, etc). I don't mind adding a dual swipe action (swipe a little to subscribe, swipe further to remove) but I feel like it would get too complicated for posts which already have 2 swipe actions on the trailing edge.

EricBAndrews: That makes sense--you're right that `FeedPost` comes preloaded with swipe actions and that that's a desired architecture, and that we are going to want a way to modify that behavior.

What if we instead keep the actual `.addSwipeyActions` view modifier within `FeedPost` and `CommunityResultView`, but replace the boolean `hasSwipeActions` with an optional `overrideSwipeActions: SwipeActionConfiguration?`, where `SwipeActionConfiguration` is a struct with a `leading` and `trailing` field--that makes it much more readable (the entire behavior is contained on a single line that's clearly "overriding the swipe actions"), consistent (the swipe action is always attached at the exact same level in the view hierarchy), and a little cleaner imo

Sjmarf: Agreed, that's a much cleaner solution. I'll implement that tomorrow 👍

Sjmarf: Done 👍

EricBAndrews (approved):
Nice! 👍

PR #704, opened 2023-10-06
Avoid search rate limit

Sjmarf:
Got a rate limit whilst searching. This PR reduces the number of API requests we make by refreshing the search results less often.

## Before this PR

Search was refreshed every time the search field content is changed (ie, when the user types another letter)

## After this PR

Search is refreshed if the user stops typing for 0.2 seconds (thoughts on this delay? I type faster than 0.2 seconds per key so it works for me, but this may not be true for everyone)


EricBAndrews (approved):
Nice! Very clean solution.

PR #705, opened 2023-10-07
Search context menus

Sjmarf:
Lets you share and block communities and users from a context menu on each search result.
IMG_9555


boscojwho (changes requested):
Looks good, works as expected:
- Would it be possible to move the destructive actions (block/unsubscribe) into its own section in the menu?

Sjmarf:
> Would it be possible to move the destructive actions (block/unsubscribe) into its own section in the menu?

When you say 'section', do you mean adding a divider or adding a separate sub-menu as used for the 'top' sort modes in the feed?

I don't think misclicks should be a huge issue - the app asks for confirmation for both unsubscribing and blocking anyways

EricBAndrews (approved):
LGTM 👍

Sjmarf:
I've had to rewrite most of this - it was quite old haha. Some extra changes that weren't present in the previous version:

- The subscribe swipe action now has a haptic (like the post swipe actions)
- Unsubscribing from a community via the swipe action now shows the 'are you sure?' popup
- Doubled the thumbnail resolution in the search tab (64 -> 128)

boscojwho (approved):

boscojwho:
Looks good:
- Actions work via swipe or context menu, as expected.
- Drag and drop works (tested by dragging link into Safari).
- Share menu generates expected items.
- Tested on iOS 17 simulator (so couldn't test swipe haptics).

PR #707, opened 2023-10-07
UserModel middleware and multiple flairs

Sjmarf:
# UserModel

Fleshed out the `UserModel` middleware, and changed the  `creator` property of `PostModel` to be a `UserModel` instead of an `APIPerson`. 

I've changed quite a bit of the related code to use `UserModel` instead. There are still some places where reaching into the `APIPerson` inside of `UserModel` is still needed. I've marked the `person` attribute of `UserModel` as deprecated to discourage it's use in future whilst we phase it out.

In future I'll be applying similar treatment for `CommunityModel`.

# Flairs

- It's now possible for a user to have more than one flair. If a user has more than one flair, their username will be uncolored as if they had no flairs.
- Added a 'banned' flair.
- Changed the 'admin' flair from red to teal to avoid having the same color as the red banned flair.
- Flairs are now visible in user search.
- Removed the flair for the mlemofficial account, which was commented out?
- Added myself to the developer flair list (if anyone else who isn't on the list wants to be, lmk). Side note: should we show this flair globally as we do now on only whilst on mlemapp?

# Screenshots
IMG_9567
IMG_9568
IMG_9569


Mlem/Models/Content/User/UserModel+ContentModel.swift
```diff
@@ -0,0 +1,26 @@
...
+    var uid: ContentModelIdentifier { .init(contentType: .user, contentId: userId) }
+    
+    var imageUrls: [URL] {
+        if let url = avatar {
+            return [url.withIcon64Parameters]
+        }
+        return []
+    }
+    var searchResultScore: Int {
+        if let commentCount = commentCount, let postCount = postCount {
```
EricBAndrews, line 20: *nit*

This can be simplified to `let commentCount, let postCount`

EricBAndrews (changes requested):
Nice! One obscenely small nit, looks great otherwise

PR #714, opened 2023-10-10
Better 'no posts to be found' message

Sjmarf:
Centered the loading graphic and replace "no posts to be found" message (see images).
<p float="left">

</p>
<p float="left">

</p>

### DIFF
--- a/Mlem/Views/Tabs/Feeds/Components/NoPostsView.swift
+++ b/Mlem/Views/Tabs/Feeds/Components/NoPostsView.swift
@@ -1,0 +1,82 @@
+//
+//  NoPostsView.swift
+//  Mlem
+//
+//  Created by Sjmarf on 10/10/2023.
+//
+
+import SwiftUI
+
+struct NoPostsView: View {
+    @EnvironmentObject var postTracker: PostTracker
+    
+    @Binding var isLoading: Bool
+    @Binding var postSortType: PostSortType
+    @Binding var showReadPosts: Bool
+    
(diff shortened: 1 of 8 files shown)


Sjmarf:
Added the fancy error-message from `ExpandedPostView` too

Sjmarf:
In future we could do something similar for the inbox tab, offering a button to show all messages if the user is only viewing unread messages

Mlem/Views/Tabs/Feeds/Components/NoPostsView.swift
```diff
@@ -0,0 +1,83 @@
...
+                        .aspectRatio(contentMode: .fit)
+                        .frame(width: unreadItems == 0 ? 35 : 50)
+                        .padding(.bottom, unreadItems == 0 ? 8: 12)
+                    Text(title)
+                    
+                    if unreadItems != 0 {
+                        Text(
+                            (unreadItems == 1 ? "1 available post was" : "\(unreadItems) available posts were")
+                            + " hidden because you have 'hide read' enabled."
+                        )
```
EricBAndrews, line 35: Could be more concise:

`\(unreadItems) read post \(unreadItems == 1 ? " has" : "s have") been hidden`

Along with the "show read," I think it's clear enough what's going on

EricBAndrews (changes requested):
Nice! Minor wording comment, but fantastic work as always

EricBAndrews (approved):

PR #715, opened 2023-10-11
Bosco/always show search bar in search

boscojwho:
# Checklist
      - *list issue(s) here*

# Pull Request Information

## About this Pull Request
Makes search bar in `Search` always visible, regardless of scroll position.
- Previous implementation would destroy home/recent/results whenever page switches in Search.
- New implementation keeps all three pages alive in a `ZStack`.
- Slight change to animation logic to workaround some jankiness resulting from altered view hierarchy.

## Screenshots and Videos

https://github.com/mlemgroup/mlem/assets/2549615/81294381-808e-4a3d-a526-8eab98d4524e


Sjmarf (changes requested):
Very nice! Works well. I noticed a small bug:

- Search for something
- Scroll down the list
- Press cancel
- Search for something again
- You're still scrolled down the page from the previous search

It would be nice if you're brought back to the top in this situation instead of remaining halfway down the page. This might be possible to do by clearing the content tracker for the search results when the user cancels the search, maybe? That should remove all of the results from the view, which would force the scroll back to the top.

boscojwho:
I'll take a look tonight, should be easy...I think XD

Sjmarf (approved):
LGTM 👍

PR #719, opened 2023-10-14
Eric/inbox middleware semaphore

EricBAndrews:
New inbox middleware making use of a stream-based multi-tracker approach

Still lots of work to do but the bones are now here


Mlem/Models/Trackers/Feed/Generic Multi-Trackers/ChildTracker.swift
```diff
@@ -0,0 +1,47 @@
...
+//  ChildTracker.swift
+//  Mlem
+//
+//  Created by Eric Andrews on 2023-10-16.
+//
+
+import Foundation
+
+class ChildTracker<Item: ChildTrackerItem>: BasicTracker<Item>, ChildTrackerProtocol {
+    private var parentTracker: ParentTracker<Item.ParentType>?
```
Sjmarf, line 11: Does a reference cycle happen here? The swift compiler doesn't check for reference cycles iirc, so this might cause a memory leak? Perhaps a weak reference is needed?

```suggestion
    private weak var parentTracker: ParentTracker<Item.ParentType>?
```

EricBAndrews: Nice catch, that's absolutely a reference cycle

Issue #569, opened 2023-09-03
Profile Tab Middleware


EricBAndrews:
Profile tab needs better middleware--instant upvote feedback is now broken with the new post tracker, since dirty status is no longer tracked per-post.

PR #726, opened 2023-10-20
Generic Multi-Trackers, Inbox Middleware

EricBAndrews:
# Checklist
      - progress on #73 and #569

# Pull Request Information

## About this Pull Request
This PR replaces the existing inbox backend with a new, middleware-model based implementation. To support this, it adds a set of new generic multi-feed trackers. In short, here's how those work:

All of the loading and filtering operations are performed by the child trackers. These look more or less identical to the existing trackers; they extend `BasicTracker`, which is much like `FeedTracking` but built with generic trackers in mind. Child trackers augment this normal tracker functionality with the ability to treat their arrays as streams via two new methods: `nextItemSortVal`, which returns the sorting value of the next item in the stream, and `consumeNextItem`, which returns the next item in the stream and increments the cursor.

To the consumer, the streams abstract loading operations; the consumer simply awaits the next item and it appears, regardless of whether data needed to be fetched. To avoid the case where multiple threads invoke loading at the same time, `loadNextPage` has been reworked: it now requires the caller to specify a page, and using a semaphore it ensures that only one thread accesses the function at a time. If the page is not the expected next page, the call is ignored. This has the consequence that threads can effectively await a `loadNextPage` called by a different thread.

`ParentTracker` then uses stream system this to efficiently aggregate child trackers: provided all child trackers are sorted on the same criterion, the parent tracker can "fetch a new page" by effectively just merge sorting the streams of its child trackers. This ensures that loading behavior is only invoked when needed, and handles cases where one child tracker has many more items than the others. Since the child trackers are themselves just trackers, they can be used as normal trackers in single-stream views while at the same time feeding a multi-feed.

This implementation is pushing the limits of what Swift really wants to do with generics--the `Item` associated with the parent and the child must be related, such that items of the child type are convertable to items of the parent type and all can be sorted on the same criteria. This latter condition unfortunately proved impossible (at least for me) to pull off in a fully compiler-safe way; there is therefore only a single `TrackerSortType`, which must enumerate *all* possible sorts for a tracker. If a given `TrackerItem` does not implement a given `TrackerSortType`, that should be expressed in assertions.

To use the generic trackers, implement:
- `ParentTrackerItem` conformance for the parent type, which should probably be an enum of child types
- `ChildTrackerItem` conformance for all child types
- Create trackers for each child type, extending `ChildTracker` and using the appropriate `ChildTrackerItem`. These should be very quick and easy, since only `fetchPage` needs to be overridden and everything else is handled by generics
- Create tracker for the parent type. Usually this can be a simple extension of `ParentTracker` with the `ParentTrackerType`

This PR also changes the structure of how the inbox is organized. The trackers are all held in `InboxView`; they are passed to the per-item views as `ObservedObject`s. The per-stream views are now their own views rather than functions on the main view, and some of the structure of the individual items has been lightly refactored. None of the changes here are major.

## A note to reviewers

The line count on this is simply obscene. It's not as bad as it looks, though, since a ton of those changes are just moving logic around, notably splitting `Inbox View Logic` into the models and moving some things between repositories.

The core new code is limited to `BasicTracker`, `ChildTracker`, and `ParentTracker`; almost everything else is refactoring or very thin (protocols, extending classes, etc.)

## Behavior Changes

Blocking a user no longer removes them from the feed. The code is there, but not used. Since those users just show up again when the inbox is refreshed, I disabled that behavior in order to present a more consistent experience; long-term, we should look into filtering blocked users more robustly, but that's out of scope here.

There is one issue I am aware of: refreshing the inbox does not refresh the read status of replies or mentions, though it does refresh messages. I'm losing sanity trying to debug that one, and it's so minor that I'm willing to call it acceptable for the sake of getting this PR in, but if anybody has any insight into why I'd love to hear it.


Mlem/Models/Trackers/Inbox/InboxItem.swift
```diff
@@ -0,0 +1,82 @@
...
+//  Mlem
+//
+//  Created by Eric Andrews on 2023-06-26.
+//
+
+import Foundation
+import SwiftUI
+
+/// Enumeration of content models that can appear in the inbox
+enum InboxItem {
```
Sjmarf, line 12: Do we need this type? There seems to be a lot of logic surrounding this. Perhaps just using the same type for the parent and child type would be better? Something like this? It does require using base classes rather than protocols though

```swift
class ContentIdentifiable {
    var uid: ContentModelIdentifier
}

class InboxItem: ContentIdentifiable {
    var published: Date
    // etc
}

class MessageModel: InboxItem {
}

class ParentTracker<Item: ContentIdentifiable>: ObservableObject, ParentTrackerProtocol {
}

class InboxTracker: ParentTracker<InboxItem> {
}
```

Instead of using a switch statement to determine types view-side, we could use `as?` statements

EricBAndrews: Updated per Slack

Sjmarf:
I'll do a proper review of this today 👍

Sjmarf:
It appears that tapping on a reply in the inbox feed no longer opens the post

Sjmarf (changes requested):
It appears that tapping on a reply in the inbox feed no longer opens the post

Mlem/Models/Trackers/Feed/TrackerItem.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+//  Created by Eric Andrews on 2023-10-15.
+//
+import Foundation
+
+protocol TrackerSortable {
+    func shouldSortBefore(_ other: Self?) -> Bool
+}
+
+protocol TrackerItem: Equatable {
```
Sjmarf, line 13: How would this system work with a model that needs to be used in multiple different parent trackers? For example, `PostModel` would be used in both the user profile and search trackers.

Something like this could work perhaps?

```swift
import Foundation


protocol SortType: RawRepresentable where RawValue: StringProtocol { }
                        
protocol ParentTrackerItem {
    associatedtype Sort: SortType
}

enum ProfileSortType: String, SortType {
    case published
}

enum SearchSortType: String, SortType {
    case searchScore
}

// This is the equivalent of AnyInboxItem
enum AnyProfileItem: ParentTrackerItem {
    // ...
    typealias Sort = ProfileSortType
}

class PostModel {
    // ...
    func sortValue(sortType: any SortType) -> Int {
        switch sortType {
        case ProfileSortType.published:
            return self.published
        case SearchSortType.searchScore:
            return self.score
        default:
            fatalError("postModel does not support this sort type!")
        }
    }
}
```

Mlem/Views/Tabs/Inbox/Feed/AllItemsFeedView.swift
```diff
@@ -39,21 +40,31 @@ extension InboxView {
...
-                        inboxReplyViewWithInteraction(reply: reply)
-                    }
-                }
+                inboxItemView(item: item)
                 
                 Divider()
             }
         }
+        
+        EndOfFeedView(loadingState: inboxTracker.loadingState, viewType: .cartoon)
```
Sjmarf, line 51: I see this on my `test-mlem` account which only has one notification, but not on my main account that has many pages of notifications

Mlem/Views/Tabs/Inbox/Feed/InteractionSwipeAndMenuHelpers.swift
```diff
@@ -1,348 +1,342 @@
+////
+////  Messages Feed View Logic.swift
```
Sjmarf, line 2: Dead code?

Sjmarf (changes requested):

Mlem/Models/Trackers/Feed/ChildTrackerProtocol.swift
```diff
@@ -0,0 +1,30 @@
...
+
+protocol ChildTrackerProtocol {
+    associatedtype Item: TrackerItem
+    associatedtype ParentItem: TrackerItem
+
+    // stream support methods
+    
+    mutating func setParentTracker(_ newParent: any ParentTrackerProtocol)
+
+    mutating func consumeNextItem() -> ParentItem?
```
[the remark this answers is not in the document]
EricBAndrews: Agreed, that's a good idea

Mlem/Models/Trackers/Feed/ChildTracker.swift
```diff
@@ -0,0 +1,90 @@
...
+
+class ChildTracker<Item: TrackerItem, ParentItem: TrackerItem>: BasicTracker<Item>, ChildTrackerProtocol, ObservableObject {
+    private weak var parentTracker: (any ParentTrackerProtocol)?
+    private var cursor: Int = 0
+
+    func toParent(item: Item) -> ParentItem {
+        assertionFailure("This method must be implemented by the inheriting class")
+        // swiftlint:disable force_cast
+        return item as! ParentItem // this cursed and wretched line is only here to make the compiler happy
+        // swiftlint:enable force_cast
```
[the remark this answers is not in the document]
EricBAndrews: Ooh nice, didn't know about that

Mlem/Views/Tabs/Inbox/Inbox View.swift
```diff
@@ -119,41 +142,38 @@ struct InboxView: View {
                 }
             }
             .pickerStyle(.segmented)
-            .padding(.horizontal)
+            .padding(.horizontal, AppConstants.postAndCommentSpacing)
+            .padding(.top, AppConstants.postAndCommentSpacing)
             
             ScrollView(showsIndicators: false) {
```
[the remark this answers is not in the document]
EricBAndrews: Interesting design question--raising in Slack

Mlem/API/Models/Messages/APIPrivateMessage.swift
```diff
@@ -18,4 +18,19 @@ struct APIPrivateMessage: Decodable {
...
+        self.id = apiPrivateMessage.id
+        self.content = apiPrivateMessage.content
+        self.creatorId = apiPrivateMessage.creatorId
+        self.recipientId = apiPrivateMessage.recipientId
+        self.local = apiPrivateMessage.local
+        self.read = read ?? apiPrivateMessage.read
+        self.updated = apiPrivateMessage.updated
+        self.published = apiPrivateMessage.published
+        self.deleted = apiPrivateMessage.deleted
+    }
```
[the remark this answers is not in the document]
EricBAndrews: It's a bit of a matter of taste, I think--with the copying initializers you have the fairly ugly initializer itself, but then you can state fake with one line:
`await setPrivateMessage(APIPrivateMessage(from: privateMessage, read: !privateMessage.read))`
compared to three:
```
var copy = private Message
copy.read = !privateMessage.read
await setPrivatemessage(copy)
```

I personally like this style because it removes complexity from the functions where state faking takes place, which is IMO a readability increase that warrants the ugliness in the initializer--we can discuss further on Slack

Mlem/API/Models/ScoringOperation.swift
```diff
@@ -7,11 +7,20 @@
...
 
 import Foundation
+import SwiftUI
 
 enum ScoringOperation: Int, Decodable {
     case upvote = 1
     case downvote = -1
     case resetVote = 0
+    
+    var color: Color? {
```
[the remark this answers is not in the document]
EricBAndrews: Agree that the middleware itself should be concerned with the what not the how; at the same time, this is a switch statement that gets duplicated around a fair amount. I've split the difference with an `AssociatedColor` protocol which `ScoringOperation` conforms to via extension, so the enum itself is pure but we get the benefit of code reuse.

Mlem/Models/Trackers/Feed/TrackerItem.swift
EricBAndrews: Oops, that protocol was an artifact of a previous iteration that snuck through.

Sorting is the biggest bump in the carpet for the whole system--for any given parent/tracker pairing, the associated `Item` of the child needs to implement the same sorting as all of the other children of the parent. I tried a bunch of ways to make this work nicely, but that's a degree of nuance that the generics system just can't quite grasp. 

My solution is more or less what you've suggested, but with only the single `TrackerSortType`, which corresponds to `TrackerSortVal`, which in turn defines a custom comparison operator that performs runtime type safety checks in lieu of compiler-enforced.

In `TrackerItem` conformance extensions, the models implement that sort value switch. I went with that instead of a more granular per-feed sort type system because there's a fair amount of overlap--profile and feed, for example, should both be sortable on all of the common sorting operations between posts and comments, so it seems a bit redundant to define `PostSortType.new` and `CommentSortType.new`

I've gone through and made this cleaner--`TrackerItem` conformance is now its own extension separate from `InboxItem` conformance for the three models, and I've cleaned out the vestigial sorting logic.

Mlem/Views/Tabs/Inbox/Feed/AllItemsFeedView.swift
EricBAndrews: Hmm, that's odd--it works properly on my main, which also has many pages of inbox, and I tested it with an aggressively reduced page size to trigger loading more often. I can't seem to repro.

Sjmarf: Ah, you're right, it's actually a different bug. I had thought I'd reached the end but I hadn't - when you scroll down too fast, it doesn't always trigger loading the next page. The easiest way to repro this is by enabling the scroll indicator and dragging the scroll bar all the way to the bottom. The EndOfFeedView loads properly for me when I scroll slowly enough.

Mlem/Models/Trackers/Feed/ParentTracker.swift
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    func addChildTracker(_ newChild: some ChildTrackerProtocol) {
+        var newChild = newChild
+        newChild.setParentTracker(self)
+    }
+    
+    /// If the given item is the loading threshold item, loads more content
+    /// This should be called as an .onAppear of every item in a feed that should support infinite scrolling
+    func loadIfThreshold(_ item: Item) {
```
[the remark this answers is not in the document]
EricBAndrews: Yeah, that's a good idea. I've implemented `CoreTracker`, which handles all of the common logic. I also renamed `BasicTracker` to `StandardTracker` because `BasicTracker`/`CoreTracker` does not make it clear which one is the core class and which is the inheriting.

Mlem/Views/Tabs/Inbox/Feed/AllItemsFeedView.swift
EricBAndrews: Fixed--I've added a fallback threshold that triggers on the last item in feed, which seems to work nicely.

Mlem/Models/Trackers/Feed/TrackerItem.swift
Sjmarf: > I went with that instead of a more granular per-feed sort type system because there's a fair amount of overlap--profile and feed, for example, should both be sortable on all of the common sorting operations between posts and comments, so it seems a bit redundant to define PostSortType.new and CommentSortType.new

Ah yes, that makes sense. Looks good to me

Mlem/API/Models/Messages/APIPrivateMessage.swift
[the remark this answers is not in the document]
Sjmarf: Sure, we can leave it as-is for now then 👍

PR #729, opened 2023-10-22
CommunityModel improvements

Sjmarf:
Fleshed out `CommunityModel` and changed the type of the `PostModel.community` property from `APICommunity` -> `CommunityModel`.


Mlem/Models/Content/Community/CommunityModel.swift
```diff
@@ -0,0 +1,170 @@
...
+        }
+        do {
+            try await apiClient.followCommunity(id: communityId, shouldFollow: !subscribed)
+        } catch {
+            hapticManager.play(haptic: .failure, priority: .high)
+            let phrase = (self.subscribed ?? false) ? "unsubscribe from" : "subscribe to"
+            errorHandler.handle(
+                .init(title: "Failed to \(phrase) community", style: .toast, underlyingError: error)
+            )
+        }
```
EricBAndrews, line 125: Both of these cases should update the tracker again to ensure it matches the server--if it succeeds, we get the latest community info, so we might as well update, and if it fails we should revert the state fake

Mlem/Models/Content/Community/CommunityModel.swift
```diff
@@ -0,0 +1,170 @@
...
+            }
+        } catch {
+            hapticManager.play(haptic: .failure, priority: .high)
+            errorHandler.handle(error)
+            let phrase = !blocked ? "block" : "unblock"
+            errorHandler.handle(
+                .init(title: "Failed to \(phrase) community", style: .toast, underlyingError: error)
+            )
+        }
+    }
```
EricBAndrews, line 150: Ditto re: call callback again

Mlem/Models/Content/User/UserModel.swift
```diff
@@ -114,7 +114,7 @@ struct UserModel {
         }
         if let comment = commentContext, comment.distinguished {
             ret.append(.moderator)
-        } else if let community = communityContext, community.moderators.contains(where: { $0.moderator.id == userId }) {
+        } else if let community = communityContext, let moderators = community.moderators, moderators.contains(where: { $0.moderator.id == userId }) {
```
EricBAndrews, line 117: Lint failure

Mlem/Models/Content/Community/CommunityModel.swift
```diff
@@ -0,0 +1,170 @@
...
+        case noData
+    }
+    
+    @available(*, deprecated, message: "Use attributes of the CommunityModel directly instead.")
+    var community: APICommunity
+    
+    // Ids
+    let communityId: Int
+    let instanceId: Int
+
```
EricBAndrews, line 27: This should also get a `ContentModelIdentifier` id for consistency and to play nice in mixed feeds

Mlem/Views/Tabs/Feeds/Components/Sidebar View.swift
```diff
@@ -46,9 +36,10 @@ struct CommunitySidebarView: View {
     private func loadCommunity() async {
         do {
             errorMessage = nil
-            communityDetails = try await communityRepository.loadDetails(for: community.id)
+            let communityDetails: GetCommunityResponse = try await communityRepository.loadDetails(for: community.communityId)
+            community = .init(from: communityDetails)
         } catch {
-            errorMessage = "We were unable to load this communities details, please try again."
+            errorMessage = "We were unable to load this community's details, please try again."
```
EricBAndrews, line 42:
```suggestion
            errorMessage = "Unable to load community details, please try again."
```

More succinct

Mlem/Models/Content/Community/CommunityModel.swift
```diff
@@ -0,0 +1,170 @@
...
+    var local: Bool
+    var removed: Bool
+    var deleted: Bool
+    var hidden: Bool
+    var postingRestrictedToMods: Bool
+    
+    // From APICommunityView
+    var blocked: Bool?
+    var subscribed: Bool?
+    var subscriberCount: Int?
```
EricBAndrews, line 48: These values get nil-checked a lot but don't have meaningful nil states--probably nicer to remove the optional and default a `nil` from the API model to false/0

Mlem/Models/Content/Community/CommunityModel.swift
```diff
@@ -0,0 +1,170 @@
...
+    let creationDate: Date
+    let updatedDate: Date?
+    
+    // URLs
+    let communityUrl: URL
+    
+    // These values are only available via GetCommunityResponse
+    var site: APISite?
+    var moderators: [APICommunityModeratorView]?
+    var discussionLanguages: [Int]?
```
EricBAndrews, line 60: Two more optionals with a non-meaningful nil state--default to empty arrays?

EricBAndrews (changes requested):
Nice!

Couple small change requests, largest one is around updating state-faked values after the API call either succeeds or returns. Per Slack and Inbox Middleware PR, that logic might look a little nicer if we convert this to an ObservableObject class, but should be super easy to port in the future if we decide to do that

Mlem/Models/Content/Community/CommunityModel.swift
Sjmarf: It's got one in `CommunityModel+ContentModel.swift`

Mlem/Models/Content/Community/CommunityModel.swift
Sjmarf: Hmm, I'm not sure what to do about this one. How do we know if a value is loaded or not? Setting `subscribed` to `false` when we don't actually know whether the user is subscribed or not would cut down on the amount of null coalescing we have to do View-side, but could be misleading.

Notably, I'm doing a check to test if one of these values is `nil` in `SidebarView`. If the value is `nil`, it will fetch the extra information. If we make these values non-optional, we'd have to add a `hasLoaded` property or something.

EricBAndrews: Ah, yes, that's a good point--I didn't catch that these also store loading state. We can leave as-is for now.

EricBAndrews (approved):

PR #743, opened 2023-11-02
Fixed some inbox problems, tweaked haptics

EricBAndrews:
# Checklist
      - Follow-up to #726 

# Pull Request Information

This PR addresses an issue created by #726 where haptics were playing twice on votes.

It also fixes an issue where marking mentions and replies as read was causing the view to hitch while the `unreadTracker` updates; `unreadTracker` is now not updated until after a successful API call. This introduces a little bit of lag between the state fake and the unread badge state, but is substantially smoother on the swipe.

It also performs some minor haptic adjustments:
- The haptics for voting and saving have been switched: voting now plays a single tap, and saving plays two taps
- Removed the "Success" haptic and replaced it with what was previously "Gentle Success"
- Replaced "Gentle Success" with "Light Success," which is the old "Firmer Info"
- Created a new "Firm Info" haptic that replaces "Firmer Info." It is slightly less heavy than the old "Firmer Info"--Gentle Info -> Firm Info -> Light Success now creates a pleasing sequence of increasing intensity


Sjmarf (approved):
LGTM! I much prefer the new swipe action haptics

PR #753, opened 2023-11-09
Eric/0.19 account deletion

### LINKED ISSUE #676: [0.19.0] Allow deletion of user content during account removal
## Introduction
As detailed in the recent Lemmy announcement when a user opts to delete their account from an instance they now have the **option** to also remove all content.

> When a user deletes their own account using POST /api/v3/user/delete_account, there is a new parameter called delete_content. If it is true, all posts, comments and other content created by the user are deleted (this is the previous default behaviour). If it is false, only the user profile will be marked as deleted.

> https://github.com/LemmyNet/lemmy/pull/3817

## Requirements
As we already support account deletion we will want to update the requests to support this new parameter. We should also update the UI to allow the user to select if they would like this done, with the default being `true`, which is the original behaviour.

EricBAndrews:
# Checklist
      - closes #676

# Pull Request Information

This PR enables deletion of content during account deletion on instances running 0.19.0 or above.

It also makes the account deletion flow scarier and harder to accidentally execute:

Screenshot 2023-11-09 at 4 10 19 PM
Screenshot 2023-11-09 at 4 10 23 PM

### DIFF
--- a/Mlem/API/APIClient/APIClient.swift
+++ b/Mlem/API/APIClient/APIClient.swift
@@ -75,8 +75,6 @@ class APIClient {
     @discardableResult
     func perform<Request: APIRequest>(request: Request) async throws -> Request.Response {
         let urlRequest = try urlRequest(from: request)
-        
-        print(String(data: urlRequest.httpBody ?? Data(), encoding: .utf8))
 
         let (data, response) = try await execute(urlRequest)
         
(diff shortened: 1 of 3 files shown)


Mlem/API/APIClient/APIClient.swift
```diff
@@ -259,8 +259,13 @@ extension APIClient {
     
     // MARK: - User requests
 
-    func deleteUser(user: SavedAccount, password: String) async throws {
-        let request = DeleteAccountRequest(account: user, password: password)
+    func legacyDeleteUser(user: SavedAccount, password: String, deleteContent: Bool) async throws {
```
Sjmarf, line 262: We don't need the `deleteContent` parameter here

Sjmarf (approved):
One nit, otherwise LGTM

Mlem/API/APIClient/APIClient.swift
EricBAndrews: Nice catch

PR #774, opened 2023-11-22
Fix top 3/6/9 month sort modes

Sjmarf:
Seems like I got the wrong name for the top 3/6/9 month sort modes last time. It annoys me that it isn't consistent haha


EricBAndrews (approved):
Nice 👍

PR #779, opened 2023-11-28
Eric/0.19 compatibility

EricBAndrews:
# Checklist
      - 0.19 compatibility

# Pull Request Information

This PR adds some missing optional flags on `id` fields in API aggregate objects that have been removed in 0.19, and the new `isAdmin` field on `APIPersonView`.

It also adds some prettier debug logging for errors.

It introduces a bizarre regression, which is that the feed doesn't load immediately on 0.19 instances--the site version change is never picked up by the `.task(id: siteInformation.version`. Works perfectly on 0.18.


Sjmarf (approved):
Looks good, aside from that regression you mentioned... a temporary "fix" would be to check the site version and attempt to load the feed every second after the view appears, perhaps?

PR #808, opened 2023-12-13
Tab Bar Navigation

boscojwho:
# Pull Request Information

### Tab Bar Navigation
- Tapping on tab now causes view to scroll to top (except for some document views in Settings).
- Tapping again causes the view to be removed ("popped") from its navigation stack.

### Overview
- Tab re-selection can now be configured to perform a dismiss action and, optionally, auxiliary action(s).
- If configured, the auxiliary action(s) is performed first, before the dismiss action is called.
- The auxiliary action can be configured to run multiple times.
- Views have to opt-in to this system (see `README` for more info).

### Auxiliary Action(s)
- In this PR, the auxiliary action is primarily used to scroll to top.
- Auxiliary actions can be run multiple times until the view decides it's had enough.
- Auxiliary actions can be configured to do pretty much anything, for example:
-- In root views that have a search bar, the auxiliary action can be set to scroll to top first, and then focus on a search bar.
-- Or it can be configured to scroll up to each parent comment.

### Dismiss Action
- This is the primary action, and causes the current view to be removed ("popped") from the navigation stack.
- The only exception is in `FeedView`, where it uses the `rootDetails` binding to pop to sidebar (the system dismiss action was not very reliable in this scenario for some reason). In this corner case setup, the auxiliary action causes `FeedView` to be removed ("popped").

### iPad:
- In `FeedView`, tapping on tab causes sidebar to switch between visible and hidden.

### What's Missing:
- Tap search tab to focus search field.
- In `DocumentView` in Settings, tapping tab never causes view to scroll to top (didn't seem particularly useful).

## Screenshots and Videos

https://github.com/mlemgroup/mlem/assets/2549615/62db20f9-e4df-4c2f-9e61-d4ad6282d366

## Additional Context
- See `README` in `/Navigation` directory for more details.
- This PR removes the current tap to scroll to top logic.
- This branch was re-created from `tab-bar-navigation-2` by following the instructions in the `README` file.

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -508,6 +508,8 @@
 		E453A1D02A81C2140004BB8A /* QuickLookPreviewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = E453A1CF2A81C2140004BB8A /* QuickLookPreviewController.swift */; };
 		E46AF98E2B29A4AA0087FDF3 /* DismissAction.swift in Sources */ = {isa = PBXBuildFile; fileRef = E46AF98D2B29A4AA0087FDF3 /* DismissAction.swift */; };
 		E46AF9902B29A50A0087FDF3 /* README.md in Resources */ = {isa = PBXBuildFile; fileRef = E46AF98F2B29A50A0087FDF3 /* README.md */; };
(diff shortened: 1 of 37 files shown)


EricBAndrews (approved):
Looks good to me--tested and everything appears to be working as intended.

PR #811, opened 2023-12-14
Accounts tab navigation

boscojwho:
# Pull Request Information

## About this Pull Request
+ Opts in the new accounts views into tab bar navigation (only dismiss, no scroll to top).
+ Moves account discussion languages view into separate view, instead of nesting it inside NavigationLink. This is necessary because SwiftUI has trouble figuring out which dismiss action to get when there are two full-screen views inside a single view definition.


EricBAndrews (approved):
Looks good to me

PR #814, opened 2023-12-17
Extension cleanup

EricBAndrews:
This PR cleans up the extensions folder, making the following changes:

- All extensions are now named `Base+Extension`
- All extension files now contain just one extension or variable. Exceptions are made for extensions that group static members (e.g., `Color+Colors`) and extensions that define the same function in multiple contexts (e.g., `Error+ToEquatableError`)
- All extensions are now grouped in folders according to what they extend, with some exceptions:
  - Mocks all live in the `Mock` group
  - View modifiers all live in the `View Modifiers` group
  - Tracker item conformance extensions all live in the `Tracker Items` group
- Removed some unused extensions


boscojwho (approved):
Improvement over current state:
It builds, it runs, it heckin` passes 🐈

PR #828, opened 2023-12-23
Account Switcher Redesign

Sjmarf:
Redesigned the Account switcher to show profile pictures and the time since you last used the account. I merged #824 into this branch so that I can work on the Quick switcher - that PR should be merged before this one

## Behaviour Changes

- I've disabled the quick-switcher and accounts page when the user only has one account. In this case, an "Add Another Account" button is shown in-place of the "Accounts" `NavigationLink` in Settings. Tapping this button opens the sign-in sheet. 
- When the user has exactly two accounts, holding down the Profile tab icon will simply toggle between the two accounts rather than opening the sheet. 
- Signing out of an account via the swipe action now asks for confirmation before signing out of the account.

## Sorting

You can now sort the account list by "Name", "Instance", or "Most Recent". When sorting by "Most Recent", the active account is only shown in Settings and the Profile switcher - it is hidden in the quick switcher. As the user won't want to switch to the account they already have active, I think it just takes up space unnecessarily. Let me know if you disagree with this design choice, though. For the other two sort modes the active account is still shown in the quick switcher, because otherwise it would mess up muscle memory.

You can also choose to group the accounts into sections via a "Grouped" toggle.
  - For the "Name" sort, this groups accounts into sections by the first letter of the account name.
  - For the "Instance" sort, accounts are grouped by instance or placed in an "Other" section if there is only one account with the instance name. 
  - For the "Most Recent" sort, accounts are split into sections for "Today", "Last 30 Days" and "Older".
- The sort mode you choose applies to everywhere the accounts list can be accessed (settings, profile and the quick switcher). 

The sort mode is saved in `@AppStorage`. The default sort mode is "Name", ungrouped. The sort-mode selector isn't shown when the user has only two accounts, and the 'grouped' option is hidden unless they have at least four accounts.

Due to an annoying bug introduced in iOS 17, we can only have the sort mode drop-down in the accounts page in Settings and in the Profile page sheet. Placing it in the `.presentationDetents` sheet causes an issue where the sheet can only display at the `.large` detent. This isn't an issue on iOS 16.


Mlem/Window.swift
```diff
@@ -26,7 +27,17 @@ struct Window: View {
...
-            .onChange(of: flow) { _ in flowDidChange() }
+            .onChange(of: flow) { [flow] _ in
+                switch flow {
+                case let .account(account):
+                    if accountsTracker.savedAccounts.contains(account) {
+                        accountsTracker.update(with: SavedAccount(from: account, lastUsed: .now))
+                    }
+                default:
+                    break
+                }
```
EricBAndrews, line 38: Can this just be rolled into `flowDidChange()`? That already has a `switch` on `flow`

EricBAndrews (changes requested):
Looks good! A couple design considerations raised in Slack.

Mlem/Window.swift
Sjmarf: This code sets the `lastUsed` time for the account that you're *leaving* when you switch accounts. `flowDidChange` is called when the app opens, too (which I suppose is a little misleading), and this code doesn't need to run in that case. Hence why the code is here. 

Putting it in `flowDidChange` would require us to add an `oldValue` parameter to the function, which would have to be `nil` when `flowDidChange` is called from `onAppear`. It could be done, if you'd prefer it that way. An alternative would be to rearrange the logic such that `flowDidChange` is *only* called when the flow changes, and logic that is shared between opening the app and switching accounts can be moved to a `setupFlow` function or something.

EricBAndrews: Ah, gotcha. This seems fine then.

Mlem/Views/Tabs/Settings/Components/AccountButtonView.swift
```diff
@@ -0,0 +1,139 @@
...
+                Spacer()
+                if appState.currentActiveAccount == account {
+                    Text("Active")
+                        .foregroundStyle(.secondary)
+                } else {
+                    Image(systemName: "chevron.right")
+                        .foregroundStyle(.tertiary)
+                        .fontWeight(.semibold)
+                        .imageScale(.small)
+                }
```
EricBAndrews, line 97:
```suggestion
                if appState.currentActiveAccount == account {
                    Image(systemName: Icons.present)
                        .foregroundStyle(.green)
                        .font(.system(size: 10.0))
                }
```

EricBAndrews (changes requested):

Sjmarf:
- Added the 'Add Account' button back to the quick switcher
- Made the active account always show in quick switcher
- Fixed missing `@` issue
- For the grouped "most recent" sort, I've moved the active account from it's own section into the "today" section to save space in the quick switcher

EricBAndrews (approved):
Nice!

PR #830, opened 2023-12-26
Embedded Post Link Multiline Fix

Sjmarf:
Fixed the embedded post link title alignment so that it's aligned to the leading edge rather than the center.

Before:

<br/>
After:
<br/>


EricBAndrews (approved):
Nice

PR #831, opened 2023-12-26
Site Version Check Rework

Sjmarf:
Rewrote the logic that decides which default sort mode to use when a feed is opened.

The new system caches the site version in `SavedAccount`. If an account becomes active and the site version is cached from a previous use, the cached site version is assumed to be true and the corresponding default sort is used. 99% of the time the cached sort mode is correct, and the feed loads correctly without having to wait for the site version to load. This results in a loading speed increase in those 99% of cases.

If there is no site version cached in `SavedAccount`, it will assume that the instance is able to handle the default sort mode. If it receives a response from the server informing it that it can't handle that mode, it will set the sort mode to the fallback and try again. This also happens if the instance proves to be running a different version as the one stored in `SavedAccount`, which would happen if the instance updates to a newer Lemmy version.
 
This system means that the page loads twice as slowly as it would otherwise when the user first logs in, or when the instance changes version.
 
 To ensure that the cached site version saves correctly, I've had to make some modifications to ensure that the accounts json file is written to only once when the user switches account. Previously, it was written to in several places - because each file-saving task was running asynchronously, this caused issues where one save operation would revert changes made in the previous save operation. We'll have to be careful not to accidentally resurface this issue in future - it seems like something that could be easily broken. Ideally, we'd make `SavedAccount` a `class` rather than a `struct`, which would allow us to safely save the accounts however many times we like. This could be difficult, though, considering that we're observing the value of `AppState.currentActiveAccount` via `.onChange`. It might be possible using `@StateObject` or iOS 17's `@Observable`.


EricBAndrews (changes requested):
Introduces a bug where navigating back to the feed overrides selected sort type:


https://github.com/mlemgroup/mlem/assets/44140166/445bd23f-3d8a-4acb-9129-47f57891fa9e

I think all that's missing is a "site information resolved" flag check

Sjmarf:
Fixed 👍

EricBAndrews (approved):
Looks good to me!

PR #834, opened 2023-12-29
New Profile Page

Sjmarf:
Redesigned the profile page.


Mlem/Enums/FeedType.swift
```diff
@@ -5,16 +5,34 @@
 //  Created by Jonathan de Jong on 12.06.2023.
 //
 
-import Foundation
+import SwiftUI
 
 enum FeedType: String, Encodable, SettingsOptions {
```
EricBAndrews, line 10: *nit*

We have a protocol for associated colors, might as well use it

```suggestion
enum FeedType: String, Encodable, SettingsOptions, AssociatedColor {
```

Mlem/Enums/FeedType.swift
```diff
@@ -5,16 +5,34 @@
...
+        case .all:
+            return "Subscribed communities from all instances"
+        case .local:
+            return "Local communities from your server"
+        case .subscribed:
+            return "All communities that federate with your server"
+        }
+    }
+    
+    var color: Color {
```
EricBAndrews, line 28: `AssociatedColor` conformance

```suggestion
    var color: Color? {
```

Mlem/Views/Tabs/Profile/UserView+Logic.swift
```diff
@@ -0,0 +1,94 @@
...
+            privatePostTracker.reset(with: newPosts)
+            
+            self.isLoadingContent = false
+//
+//            errorDetails = nil
+        } catch {
+//                errorDetails = ErrorDetails(error: error, refresh: {
+//                    await tryReloadUser()
+//                    return userDetails != nil
+//                })
```
EricBAndrews, line 84: Dead code?

Mlem/Views/Tabs/Profile/UserView.swift
```diff
@@ -0,0 +1,282 @@
...
+                await tryReloadUser()
+            }
+        }
+        .onChange(of: user.userId) { _ in
+            Task {
+                await tryReloadUser()
+            }
+        }
+        .refreshable {
+            await tryReloadUser()
```
EricBAndrews, line 191: This call gets cancelled by the view redraw--either state changes within the function need to be delayed until after all calls complete or the whole thing can just be wrapped in a task (value awaited so that the `refreshable` indicator waits properly)

```suggestion
            await Task {
                await tryReloadUser()
            }.value
```

EricBAndrews (changes requested):
A couple minor comments but looks good overall!

EricBAndrews (approved):
Nice work! Looks great

PR #836, opened 2023-12-30
Accout Switch error bug fix

Sjmarf:
When switching accounts, there was a chance of a `fatalError` occurring. The `fatalError` did not have a line number, and had the following message:

```
Fatal Error: Duplicate keys of type 'SavedAccount' were found in a dictionary.
```

This occurred when switching accounts using the quick-switcher, settings page or profile switcher. I was able to reproduce the problem roughly every 5-10 account switches via the quick switcher. I was logged into 3 accounts, if that makes any difference.

To be honest, I'm not exactly sure what was causing this. We don't have a dictionary with `SavedAccount` keys anywhere in the program, so maybe it's an internal SwiftUI issue? I did some testing, and discovered that it didn't occur if I switched accounts not via the quick switcher view (I added a temporary button that switched to a different account to test this). I also added debug `print` statements to find that the issue didn't occur anywhere in the `flowDidChange` logic. So I reckon it's probably something to do with SwiftUI.

I've "fixed" the bug by immediately hiding the accounts list as soon as one of the buttons is clicked, which seems to fix the issue. Since the bug happens seemingly randomly, there's a chance I just haven't encountered it yet and it's still there haha, but I've tried a good 100 or so times and haven't got it again thus far. So I'm hoping this is a working fix

### DIFF
--- a/Mlem/Views/Tabs/Settings/Components/AccountButtonView.swift
+++ b/Mlem/Views/Tabs/Settings/Components/AccountButtonView.swift
@@ -15,6 +15,7 @@ struct AccountButtonView: View {
     @Environment(\.setAppFlow) private var setFlow
     
     @State var showingSignOutConfirmation: Bool = false
+    @Binding var isSwitching: Bool
     
     enum CaptionState {
         case instanceOnly, timeOnly, instanceAndTime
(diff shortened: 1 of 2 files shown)


EricBAndrews (commented):
Seems to work as intended 👍

EricBAndrews (approved):

PR #837, opened 2024-01-01
Feed Headers

Sjmarf:
>[!IMPORTANT]
> I merged #834 into this branch, so that PR should be merged into `dev` before this one. 

Apologies for the large PR 😅  At least some of the line count is probably changes from #834.

## New Headers

Added headers to both the Community and the Feed. The Community view has tabs that display all of the community sidebar information, and the dedicated sidebar view has been removed. To facilitate this, I've split the current `FeedView` into `CommunityView` and `FeedView`, which both contain a `PostFeedView` that implements most of the feed logic. I pretty much rewrote it from scratch, so if I missed anything or you find a bug LMK.

Tap the button in the top-right of `CommunityView` to subscribe, long press to favourite.

## Other changes

- Added a 'Copy Name' action to communities that copies the fully qualified name of the community. You can also do this by tapping the community name in the header of the feed.
- The 'favorite' toolbar action in `CommunityView` now properly creates a view update (it didn't before).
- `CommunityResultView` now has "New Post" and "Favorite" context menu actions. I added a second trailing swipe action for favoriting the community. A blue star is displayed instead of a green checkmark if the community is favorited.
- You now can't favorite a community without subscribing to it. The "Favorite" option is still shown when you aren't subscribed; tapping it will both subscribe to the community and favorite it.
- Swapped the `ellipsis` icon in the top-right of the feed for `ellipsis.circle`. It's more consistent with Apple's own apps, but that's not the reason it's changed - SwiftUI forces me to use `ellipsis.circle` because of the way I've set up the toolbar actions. If you prefer `ellipsis` there's probably a different way we can do the toolbar, though it might not be as convenient.


Mlem/Views/Tabs/Feeds/FeedView.swift
```diff
@@ -0,0 +1,225 @@
...
+                if !postTracker.items.isEmpty {
+                    Divider()
+                }
+                PostFeedView(postTracker: postTracker, postSortType: $postSortType)
+                    .background(Color.secondarySystemBackground)
+            }
+        }
+        .refreshable {
+            await postTracker.refresh(clearBeforeFetch: true)
+        }
```
EricBAndrews, line 93: Should be wrapped in a `Task` to prevent view redraw cancellation problems:

```suggestion
        .refreshable {
            await Task {
                _ = await postTracker.refresh(clearBeforeFetch: true)
            }.value
        }
```

EricBAndrews: Also not sure how I feel about clearBeforeFetch being on here, it causes a weird little flicker if the refresh is fast that isn't super great

Mlem/Views/Tabs/Profile/AvatarBannerView.swift
```diff
@@ -0,0 +1,108 @@
...
+                                }
+                                .compositingGroup()
+                            }
+                            
+                        }
+                        Spacer()
+                    }
+                    if showAvatar {
+                        avatarView
+                    }
```
EricBAndrews, line 97: *nit*

Probably nicer to rip out the `ZStack` and put this in an `.overlay`--I think SwiftUI handles that a little more efficiently because it saves a size computation

```suggestion
                    .overlay {
                        if showAvatar {
                            avatarView
                                .frame(maxHeight: .infinity, alignment: .bottom)
                        }
                    }
```

EricBAndrews (changes requested):
Nice! Noticed a bug:

- Navigate to community info page
- Go to the Moderators tab
- Subscribe or unsubscribe
- Moderator list vanishes!

Mlem/Views/Tabs/Feeds/FeedView.swift
Sjmarf: Oops, I didn't mean to put that here haha. Nice catch.

Sjmarf:
Fixed 👍

EricBAndrews (approved):
LGTM!

PR #841, opened 2024-01-03
Automatic collapse child comments

schmeat:
# Checklist
      - *list issue(s) here*

# Pull Request Information

## About this Pull Request

Adding a feature that automatically collapses child comments when a user loads a page. A feature toggle is added to the settings with the default being off.

## Screenshots and Videos

| Screenshot | Screenshot |
|--------|--------|
| Simulator Screenshot - iPhone 15 Pro - 2024-01-05 at 17 50 09 | Simulator Screenshot - iPhone 15 Pro - 2024-01-03 at 15 32 59 | 
| Simulator Screenshot - iPhone 15 Pro - 2024-01-05 at 17 52 15 | Simulator Screenshot - iPhone 15 Pro - 2024-01-05 at 17 52 25 | 

| Video |
|--------|
| https://streamable.com/b0h7s2  | 

## Additional Context

Other apps such as Apollo have this feature and thought it would be a good addition. Please let me know if I should change naming conventions. The icon in the settings page is missing too, was not sure what the best one is available for that.

Apollo did have different UI for this feature, it would say [X amount of replies] under each parent comments and would load comments if a user clicked that. I can add that functionality but just did core functionality for now while I wait for feedback.


Sjmarf (changes requested):
Thanks for the PR! This is a useful feature. 

Regarding the settings icon, I think `"arrow.down.and.line.horizontal.and.arrow.up"` would be a good choice. This is what we use elsewhere in the app to represent collapsing a comment.

I did notice a small issue. When a comment is collapsed, its children should be hidden completely. However, when a post is first opened with the new setting enabled, all of the children are visible, even those whose parents are collapsed. 

Here's what it look like currently:

Screenshot 2024-01-03 at 18 23 39


Here's what it _should_ look like:

Screenshot 2024-01-03 at 18 23 53

The correct appearance can be reached by expanding and re-collapsing each top-layer (orange-colored) comment. 

The way to resolve this issue might involve calling `setCollapsed` on each reply instead of setting the initial `isCollapsed` value

schmeat:
Good catch @Sjmarf , I'll check to see why that collapse behaviour is happening. I'll also update the icon too. Thanks

schmeat:
Adding the feature flag to `isParentCollapsed` produces this result:

Simulator Screenshot - iPhone 15 Pro - 2024-01-03 at 10 44 09


I do find that it looks better this way and the user can see amount of replies in the bottom portion of the comment box. However, clicking on it collapses the comment which means the user has to press the comment twice to see child comments. Looking at ways to work around that right now, any suggestions?

I also need to put `isParentCollapsed`to `true` on both of the available comment blocks in order to produce this result, doing it on only one seems to just produce the same behaviour as before.

Sjmarf:
Hmm yeah that's a little tricky. 

I think we'd need to have `isParentCollapsed` set to `true` for all comments further down than the orange-colored reply, but set to `false` for the top-level comment and the orange-colored reply.

The `comment` object doesn't store the depth of the comment, so we can't find the comment depth directly from there. However, as `populateChildren` is a recursive function, it should be possible to add a `depth` parameter to the function that tracks the current depth of the comment. Then, we could set `isParentCollapsed` to `true` for all comments with a depth of 2 or more.

schmeat:
Ok updated logic:

https://streamable.com/fevk7o

Mlem/Icons.swift
```diff
@@ -9,7 +9,7 @@ import Foundation
 import SwiftUI
 
 /// SFSymbol names for icons
-struct Icons {
+enum Icons {
```
schmeat, line 12: I'm not sure why this changed, I think xcode did it automatically? Can revert if needed

Sjmarf: Hmm, possibly. I don't think it makes a difference code-wise, but I'd rather keep it as a struct

Mlem/API/Internal/HierarchicalComment.swift
```diff
@@ -22,11 +22,14 @@ class HierarchicalComment: ObservableObject {
...
+        let depth = max(0, comment.comment.path.split(separator: ".").count - 2)
+        
         self.commentView = comment
         self.children = children
-        self.depth = max(0, commentView.comment.path.split(separator: ".").count - 2)
-        self.isParentCollapsed = parentCollapsed
-        self.isCollapsed = collapsed
+        self.depth = depth
+        self.isParentCollapsed = collapseChildCommentsFlag && depth > 1 || parentCollapsed
+        self.isCollapsed = collapseChildCommentsFlag && depth == 1 || collapsed
```
Sjmarf, line 32: What do you think about collapsing all comments, not just the top layer? That way, when you expand a reply, that reply's replies are also collapsed until tapped.

```suggestion
        self.isCollapsed = collapseChildCommentsFlag && depth >= 1 || collapsed
```

Sjmarf (changes requested):
Looks good! I like the new settings location.

Mlem/API/Internal/HierarchicalComment.swift
schmeat: I was considering that but this is closer to how Apollo had it where all the comments would load after you uncollapse one. In the future I'll submit something closer, where it just shows "x amount of replies" under the parent comment and it uncollapses all of the children when you click that. 


Up to you though, I can make it collapse everything if you want.

schmeat:
For some reason every time I try to revert the enum, but it discards the change and commits a blank file. Unsure why that's happening, maybe due to git hooks? Not sure

Mlem/API/Internal/HierarchicalComment.swift
EricBAndrews: I agree with the way it currently is (expanding one reply expands its whole sub-thread)--that feels like a more natural flow

Sjmarf: Sure, that's fine 👍 

I agree that in the future an "expand n comments" button, reminiscent of what Apollo had, would be a better experience.

EricBAndrews:
> For some reason every time I try to revert the enum, it discards the change and makes a blank commit. Unsure why that's happening, maybe due to git hooks? Not sure

This is due to the Swiftformat hook. `Icons` contains nothing but static members--it's basically a namespace, except that Swift doesn't have namespaces. It therefore doesn't make sense for it to be instantiable, hence declaring it as a caseless `enum`.

schmeat:
Okay I've updated the UI and the logic technically works right now. Was briefly trying to figure out why depth 1 comments were not unwrapping when using `commentTracker.setCollapsed(false, comment: hierarchicalComment)` on the top comment, but just iterating through the depth 1 comments for now. Will revisit when I wake up.

| Dark | Light |
|--------|--------|
| Simulator Screenshot - iPhone 15 Pro - 2024-01-05 at 01 59 06 | Simulator Screenshot - iPhone 15 Pro - 2024-01-05 at 01 59 12 |

schmeat:
I think I'm good to go with this PR now, let me know if anything needs to change.

Sjmarf:
Not a huge issue, but if you turn off "Collapse comments" whilst a post is open, the "Show n comments" buttons disappear but the collapsed comments don't appear. I think that either it should update immediately to reflect the new setting, or remain on the old setting until the user refreshes or re-enters the post.

schmeat:
Should be fixed now! Another thing I did notice is that the app doesn't display deleted comments but they're counted in the childComment count, so in controversial threads it'll show a bunch of replies to show but none actually display when un-collapsing.

schmeat:
Second issue should be resolved now too @Sjmarf 

I made it so that if a user views all child comments, then collapses top level comment, the children will be uncollapsed if the uncollapse the top level comment again.

Mlem/Views/Shared/Comments/Components/CollapsedCommentReplies.swift
```diff
@@ -0,0 +1,33 @@
...
+                .border(width: lineWidth, edges: [.leading], color: .accentColor)
+                .frame(width: lineWidth)
+            Image(systemName: Icons.replies)
+            Text("show \(numberOfReplies) \(replyText)")
+                .foregroundStyle(.blue)
+        }
+        .padding(.top, 5)
+        .padding(.bottom, 10)
+        .frame(maxHeight: 50)
+    }
```
Sjmarf, line 28: What's your opinion on these small design tweaks for the "show n replies" button? 




```suggestion
        HStack {
            Rectangle()
                .border(width: lineWidth, edges: [.leading], color: .accentColor)
                .frame(width: lineWidth)
            Image(systemName: Icons.replies)
            Text("Show ^[\(numberOfReplies) Reply](inflect: true)")
                .foregroundStyle(.blue)
                .padding(.vertical, 10)
        }
        .frame(maxHeight: 50)
        .padding(.leading, 10)
```

Sjmarf (changes requested):
Looks good! One more thing I found - if you navigate to a user profile from the comments page and then return to the post, all of the comments you may have expanded become collapsed again.

schmeat:
That UI update looks good, i'll look into that other bug now

schmeat:
@Sjmarf This one seems a bit tricky since it looks like the comments are being rebuilt when the user goes back to the post, so I'm not sure how I'll retain the state of comments. I'm trying to figure out why `isCommentReplyHidden`is not being reset when going to profile though.

schmeat:
Okay still a bit confused as to why it's collapsing because it's not hitting any of my collapse methods when going back to the thread. But I have worked around that by showing the comment collapse reply button when going back @Sjmarf

Sjmarf:
Hmm yeah I've had a look and I'm not sure why it's happening either. Your solution works fine for me.

Mlem/Views/Shared/Comments/Comment Item.swift
```diff
@@ -184,6 +187,21 @@ struct CommentItem: View {
...
                 }
+                
+                if collapseComments,
+                   !hierarchicalComment.isCollapsed,
+                   hierarchicalComment.depth == 0,
+                   hierarchicalComment.commentView.counts.childCount > 0,
+                   !isCommentReplyHidden {
+                    HStack {
+                        Divider()
+                        CollapsedCommentReplies(numberOfReplies: .constant(hierarchicalComment.commentView.counts.childCount))
```
Sjmarf, line 198: This Divider() looks better outside the HStack, I think

```suggestion
                   !isCommentReplyHidden {
                    Divider()
                    HStack {                        CollapsedCommentReplies(numberOfReplies: .constant(hierarchicalComment.commentView.counts.childCount))
```

Sjmarf (changes requested):
In a user profile, the "Show n replies" button is sometimes shown where it shouldn't be. The dividers between comments also appear to be too thick in some places in the profile comments feed.

IMG_9873

schmeat:
I'll have a look but I do remember running issues into hiding it without knowing the context of which page it's being displayed on

schmeat:
Fixed now, realized I was checking the wrong child count, so it would show children in user profile even though it had no children in the array

Sjmarf (changes requested):
The dividers between comments on a user profile are too thick in some cases.

IMG_9882

schmeat:
@Sjmarf Ok that was a great catch, the reason it was showing thick lines was because it was collapsing any user comment that was depth level > 0. I have added page context variables to stop certain logic from happening in the user profile, such as auto collapsing and manually collapsing. I am also able to remove previous ònDisappear`code because of that.

Let me know what you think.

Mlem/Views/Tabs/Profile/UserFeedView.swift
```diff
@@ -33,11 +33,10 @@ struct UserFeedView: View {
     }
     
     var isOwnProfile: Bool {
-        return siteInformation.myUserInfo?.localUserView.person.id == user.userId
+        siteInformation.myUserInfo?.localUserView.person.id == user.userId
```
schmeat, line 36: this appears to be the formatter automatically doing it

Mlem/Views/Tabs/Profile/UserView+Logic.swift
```diff
@@ -42,8 +42,8 @@ extension UserView {
     func tryReloadUser() async {
```
schmeat, line 42: all of this is the formatter as well

Sjmarf (changes requested):
Yep, that fixed it!

This introduced a bug, though. If you click "show n replies", then go to a user profile, then click "back", the comment tree is collapsed again but the "show n replies" button is hidden. 

Before the most recent fix, the "Show n replies" button would be (correctly) shown in this situation.

schmeat:
That was because I removed the `onDisappear`since it should only be collapsing in certain contexts, but still not sure what's causing it. Anyways I readded that `onDisappear` code and it should work now

Sjmarf (changes requested):
Because the comments section uses a `LazyVStack`, using `.onDisappear` makes the "Show n replies" button appear again on an expanded comment if you scroll down the comments and then back up.

schmeat:
@Sjmarf Not sure what the best solution to this would be. I tried to set `isCommentReplyHidden` in `.onAppear` instead, but it seems like when we navigate back to the post from the user profile, `isCollapsed`is not being set correctly in `HierarchicalComment`. I'm unsure what method is actually collapsing it still, since it's not hitting any of my current collapse methods.

I tried debugging and it's showing as retaining the hidden variable but not resetting collapse.

```
ID: 3350207, isCollapsed: false, isCommentReplyHidden: false, number of children: 8
ID: 3350207, isCollapsed: false, isCommentReplyHidden: true, number of children: 8
```

Would that be an acceptable bug to ship with? Will take me some more time to investigate the root cause of it and I'm not sure how much time I can dedicate this week. If you have a hunch as to why this behaviour is happening please let me know.

Sjmarf:
I'll take a look at it this weekend and see if I can find the issue 👍

Sjmarf:
Looks like it's caused by the `.task` in `ExpandedPost` line 85. Adding an `if` statement fixes it

```swift
.task {
    if commentTracker.comments.isEmpty {
        await loadComments()
    }
}
```

schmeat:
Thanks so much for checking, I've added those changes now!

Sjmarf (approved):
Looks good to go!

schmeat:
Is there something I need to fix to get it merged? I noticed some build failures but logs say something about access

Sjmarf:
Every time you make a change, the checks need to re-run. For organisation members this happens automatically, but for everyone else it has to be approved by an organisation member :( 

Feel free to merge this whenever you like (if you're able to? If not lmk and I'll merge it).

schmeat:
Okay sounds good. I don't think it lets me merge unless all checks are passed.

Sjmarf:
Huh, I see the `Error: Resource not accessible by integration` too now 🤔 

I've not seen that before

schmeat:
Not sure what the build settings for this project are, but maybe this is it:

https://github.com/kishikawakatsumi/xcresulttool/issues/691#issuecomment-1825814516

> We need pull_request_target if the repository is forked.

Sjmarf:
Thanks - Jake's put a PR up to fix it, once that's in I'll rerun the checks 👍

EricBAndrews:
Builds and checks pass locally--we're still getting our CI stuff worked out for forked builds, so once this PR is merge-compatible I'll run the checks again locally and bypass CI to merge

schmeat:
Sorry for the delay, busy week. Should be good to go now.

EricBAndrews:
Build and test work locally. I'm bypassing branch protections to merge without CI; if this breaks `dev` that's on me.

PR #842, opened 2024-01-04
Fix scroll-up feed lag

Sjmarf:
Fixed an issue where the feed would "jump" when scrolling up. This issue occurred in both the posts feed and comments section, or rather anywhere where `MarkdownView` is used inside of a `LazyVStack`. 

The text to be rendered by the `MarkdownView` was stored in an `@State` array attached to the view. The array was empty initially, and then populated when the view appeared. This caused the `MarkdownView` to appear with no content, and then load the content shortly after via a `.task`, which caused a height change in the view.

I've fixed this by populating the array in the `MarkdownView` initialiser instead. Using this instead of `.task` doesn't appear to impact performance.


EricBAndrews (approved):
Nice! This cleans up a lot of bad behavior

PR #847, opened 2024-01-10
Fix Sort Mode Labelling Inconsistency

Sjmarf:
The labels of the "Six Hours" and "6 Months" sort modes used inconsistent naming schemes. I changed the labels so that they all use numerals ("6") instead of words ("Six"). If you prefer they all use "Six", lmk and I'll change it. We should keep it consistent either way.

Also fixed a casing inconsistency with comment context menu actions. I changed "Undo upvote" to "Undo Upvote".


EricBAndrews (approved):
🚀

PR #848, opened 2024-01-10
Fix "Show score in info" setting default value inconsistency

Sjmarf:
The default value for this was inconsistent. I've set it to `false` everywhere, since the score is already shown in the Upvote/downvote widget.

Also added a footer to the "Interactions and info" section in post appearance settings to clarify that the settings only apply to Large & Headline posts. For Compact posts, the user's preferences are ignored and all of the info is shown.


EricBAndrews (approved):
Nice 👍

PR #849, opened 2024-01-10
Fix community "Favorite" action being shown when it shouldn't

Sjmarf:
When the `subscribed` property of `CommunityModel` is `nil`, it won't show the "Subscribe" action in context menus or swipe actions. This occurs when the `CommunityModel` was created from an `APICommunity` and not an `APICommunityView`, such as in the moderated communities list on a user profile. 

This PR fixes a bug where the "Favorite" action was still shown even when the "Subscribed" action was hidden. Now, both actions are hidden if the `subscribed` property is `nil`.


EricBAndrews (approved):
Nice 👍

PR #850, opened 2024-01-11
Fix `flowDidChange` being called extraneously

Sjmarf:
`flowDidChange` was called via `onAppear` on the `Window` view. `onAppear` was triggered when navigating back from an external link, which called `flowDidChange`. I've fixed this by only calling `flowDidChange` the first time `onAppear` is triggered.

This _might_ be related to the app store review about having to re-authenticate when navigating back from an external link?

It might also be related to the error haptic that I've been getting when navigating back from a link, though I'm not sure because I can't reproduce that bug at the moment. It might be dependent on internet speed or something, because I was able to reproduce it consistently yesterday and nothing's changed since then.


Mlem/Window.swift
```diff
@@ -38,11 +39,17 @@ struct Window: View {
...
+                if !loadedInitialFlow {
+                    flowDidChange()
+                    loadedInitialFlow = true
+                }
+            }
             .environment(\.setAppFlow, setFlow)
     }
 
     func flowDidChange() {
+        print("FLOW CHANGE")
```
EricBAndrews, line 52: Intentional?

EricBAndrews (changes requested):
Looks good! Maybe leftover test print?

Mlem/Window.swift
Sjmarf: Removed 👍

EricBAndrews (approved):
Nice 👍

PR #851, opened 2024-01-11
Matrix ID Settings Bug Fixes & UI Tweaks

Sjmarf:
A couple of bug fixes related to the Matrix ID page:

- Matrix ID input field always showed `""` when opening the page, not your existing Matrix ID.
- Pressing "cancel" the first time wouldn't do anything - you'd have to press it twice for the cancel to take effect.

I changed the description text (it seems Lemmy messaging doesn't go through Matrix as I previously thought) and also tweaked the appearance of the navigation link to show your matrix ID if you have one.

This is shown if you have a matrix ID:

IMG_9900

This is shown if you don't:

IMG_9901


EricBAndrews (approved):
👍

PR #852, opened 2024-01-11
Fix toggling "hide read" not immediately updating

Sjmarf:
Toggling "Hide Read" via the ellipsis menu in a feed wouldn't immediately work - you'd have to exit and re-enter the feed to see changes. Now, the feed refreshes immediately with the new setting applied. 

I don't really know why this fixes it, but it works 🤔


JakeShirley (approved):
The best kind of fix!

PR #853, opened 2024-01-12
Fix NSFW Community Search Result title bug

Sjmarf:
Previously, NSFW community names would display like this

```
Optional("name") • NSFW
```

This PR fixes the problem:

```
name • NSFW
```

Also, the avatar would be blurred even if it was the default avatar. Now, the avatar won't be blurred if it's the default image.


Mlem/Views/Tabs/Search/Results/CommunityResultView.swift
```diff
@@ -41,9 +41,9 @@ struct CommunityResultView: View {
     
     var title: String {
         if community.blocked ?? false {
-            return "\(community.name) ∙ Blocked"
+            return "\(community.name!) ∙ Blocked"
         } else if community.nsfw {
-            return "\(community.name) ∙ NSFW"
+            return "\(community.name!) ∙ NSFW"
```
EricBAndrews, line 46: We can avoid force unwrapping altogether, and also capture both blocked and NSFW status:

```
        var suffix = ""
        if community.blocked ?? false {
            suffix.append(" ∙ Blocked")
        }
        if community.nsfw {
            suffix.append("∙ NSFW")
        }
        
        return community.name + suffix
```

EricBAndrews (changes requested):
One nit re: force unwrapping, otherwise looks good!

EricBAndrews (approved):
Nice!

PR #854, opened 2024-01-12
User Settings UI Consistency with User Profile

Sjmarf:
Made the headers of the `UserView` and `AccountSettings` view use consistent styling. If the user has a banner, it will now be shown in account settings. Also made the navigation link for account settings show the account display name rather than the username.


Mlem/Views/Tabs/Settings/Components/SettingsView.swift
```diff
@@ -38,11 +39,11 @@ struct SettingsView: View {
                                 )
                                 .padding(.vertical, -6)
                                 .padding(.leading, 3)
-                                if let account = appState.currentActiveAccount {
+                                if let person = siteInformation.myUserInfo?.localUserView.person {
                                     VStack(alignment: .leading, spacing: 3) {
-                                        Text(account.nickname)
+                                        Text(appState.currentActiveAccount?.nickname ?? info.localUserView.person.name)
```
EricBAndrews, line 44: Build fails here--maybe something like

```suggestion
                                        Text(appState.currentActiveAccount?.nickname ?? person.name)
```

EricBAndrews (changes requested):
Doesn't build, but looks good otherwise!

EricBAndrews (approved):
LGTM 🚀

PR #856, opened 2024-01-14
Instance Page

### LINKED ISSUE #115: Provide a screen with instance details
Instances (under GetSite) provide a lot of details about the current instance, such as discussion languages, description, moderators, and other things. We should expose this in the app, and provide a way for a user to look up this information in-app.

Sjmarf:
Closes #115

This PR implements a basic instance details page. Navigate to the instance page via the ellipsis menu in `CommunityView` or `UserView`. 

The instance page has two tabs. The "About" tab shows the instance's sidebar information, and the "Administrators" tab shows a list of instance administrators.

Only Lemmy instances are supported, and not KBin instances. KBin doesn't appear to have a `GetSiteRequest` equivalent, or if there is, it isn't described in their [API Docs](https://docs.kbin.pub/). If you try to visit a KBin instance, a suitable error is shown.

 I also tweaked the `UserResultView` and `CommunityResultView` code in this PR (no appearance changes though).

## Screenshots

# Future Plans

This PR includes a bare-essentials implementation. In future, I would like to make improvements to this page.

### "Details" tab

Similar to the community "Statistics" tab. Edit: Implemented in #869.

### "Uptime" tab

Fetches uptime data from https://lemmy-status.org.

### "Trust & Safety" tab

Fetches information from https://fediseer.com.

### Instance Search

Edit: I implemented this in #873.

I think the ideal way to do this would be to have a GH action that pulls from lemmyverse every day and creates a file storing the instance names, software type, and icon. We could exclude instances with <5 members or something, since those are single-person instances that people probably aren't searching for. Then, the client would download this list every time the app opens.


Mlem/API/APIClient/APIClient.swift
```diff
@@ -161,8 +161,6 @@ class APIClient {
         
         if let overrideToken {
             urlRequest.setValue("Bearer \(overrideToken)", forHTTPHeaderField: "Authorization")
-        } else if case let .authenticated(_, token) = session {
-            urlRequest.setValue("Bearer \(token)", forHTTPHeaderField: "Authorization")
```
Sjmarf, line 165: This was breaking the `GetSiteRequest(instanceURL: )` initialiser. I don't think we need this?

Sjmarf:
Added pre-emptive support for `GetPersonDetailsResponse.site` (https://github.com/LemmyNet/lemmy/pull/4394)

EricBAndrews (approved):
Nice! Looks really good

PR #857, opened 2024-01-14
Community Favorite Fix

Sjmarf:
Fixed a bug where favoriting a community from the ellipsis menu in `CommunityView` wouldn't update the `CommunityModel` with the new state of `subscribed`, so the community would appear unsubscribed to until the view was exited and re-entered.


EricBAndrews (approved):
Nice

PR #858, opened 2024-01-14
New Comments Indicator

Sjmarf:
Adds an indicator to the info stack that shows the number of unread comments. Let me know what you think of the design - is the green too obtrusive?

I also changed the direction of the speech bubble from right-facing to left-facing across the whole app. Apple uses left-facing bubbles in all of their uses of the symbol, so I believe this is the best option. Again, if you have any thoughts about or disagree with this change, lmk.

IMG_9909


EricBAndrews (approved):
Nice 👍

PR #860, opened 2024-01-15
Image Links & Badges

### LINKED ISSUE #541: Handle images embedded as part of links
Markdown allows for images to be embedded as links, using this syntax:

`[text and image  as link](link-to-content)`

Right now this breaks our Markdown parser just a tiny bit--it displays the image properly, but the rest of the content is not parsed as a link. We need to:
a. Figure out how to elegantly display this in feed--images right now get full width, which doesn't align well with inlined image/link hybrids
b. Parse this case in our Markdown handling
c. Display (a) when we parse (b)

Sjmarf:
Closes #541.

Adds support for image links (`[Alt text](https://google.com/link)`). Inline images now also have a corner radius to match images in posts.

Also adds rendering of `shields.io` badges. In the wild, this only really happens in instance sidebars. I implemented those in #856; that PR is independent to this one. For testing purposes, I've created a couple of posts on `test-mlem.jo.wtf` that contain the badges.

`shields.io` lets the badge creator specify an SVG as an icon. This is problematic, because we can't easily render SVGs from URLs. Instead, I've made it support a set number of images, which we store on-device as SF symbols. If an unsupported image is set, no image is shown. The following logos are supported.

- Github
- Matrix
- Discord
- Lemmy
- Mastodon

This is only a partial implementation of what `shields.io` offers - some `shields.io` badges include metrics that it fetches from other sources. I haven't implemented this. For example, `shields.io` Discord badges show the number of users in the Discord server. Our rendering of it doesn't. 

In addition to `shields.io` badges, some instances use badges from `fediseer.com` or Lemmy uptime sites. Since this data will be available inside of Mlem, I've opted to not render these badges at all.


EricBAndrews (approved):
Nice!

EricBAndrews (approved):
🚀

Sjmarf:
Thanks for the double approval :D

PR #862, opened 2024-01-17
Face ID Lock

schmeat:
# Checklist
      - *list issue(s) here*

# Pull Request Information

## About this Pull Request

This adds a MVP of face ID locking. It will allow the user to lock their app until face ID succeeds.

## Screenshots and Videos

| Lock screen | Lock screen |
|--------|--------|
| Simulator Screenshot - iPhone 15 Pro - 2024-01-17 at 10 38 27 | Simulator Screenshot - iPhone 15 Pro - 2024-01-17 at 10 38 20 | 

| Error | Error |
|--------|--------|
| Simulator Screenshot - iPhone 15 Pro - 2024-01-17 at 11 08 22 | Simulator Screenshot - iPhone 15 Pro - 2024-01-17 at 11 08 39 | 

https://github-production-user-asset-6210df.s3.amazonaws.com/1139817/297490338-b5185149-556d-4932-aa9a-a52d3e7b689a.mp4

## Additional Context

In the settings it is currently a drop down rather than pill box because I plan on extending functionality to only ask to unlock every n minutes. Currently it is set to lock every time you exit and renter the app, or "instant".


Mlem/API/APIClient/APIClient.swift
```diff
@@ -324,10 +323,10 @@ extension APIClient {
                 showReadPosts: localUser.showReadPosts,
                 showScores: localUser.showScores,
                 theme: localUser.theme,
-                auth: try session.token
```
schmeat, line 327: formatter

Mlem/API/APIClient/APIClient.swift
```diff
@@ -324,10 +323,10 @@ extension APIClient {
...
+                auth: session.token
             )
         )
-        return SuccessResponse(from: try await perform(request: request))
+        return try await SuccessResponse(from: perform(request: request))
```
schmeat, line 329: formatter

Mlem/ContentView.swift
```diff
@@ -76,18 +80,18 @@ struct ContentView: View {
                 }
                     
                 ProfileView(user: myUser)
-                .fancyTabItem(tag: TabSelection.profile) {
```
schmeat, line 79: formatter

Sjmarf:
Nice!

- Do we need to show the "Unlock Mlem" button the entire time? What if we only showed it if the initial FaceID unlock fails?
- How does this behave on devices without Face ID? (XCode's Simulator lets you test other types of devices, and there's options to fake a matching or unmatching fingerprint / face in the menu bar).
- When you unlock the app using Face ID, the feed you're looking at gets reloaded. Is there a way to avoid this? Maybe by just showing the view over the top of the app instead of using an `AppState` case?

<br/>
- Maybe it's possible to hide the content of the app whilst in the app switcher? I've seen other apps do this, but I'm not sure how it's done. It would be nice to have at some point, but I don't mind if it isn't in this PR.
- At some point it would be nice to have this option available per-account (the option would be inside of Account Settings instead of inside General). Again, that doesn't need to be in this PR.

schmeat:
> Do we need to show the "Unlock Mlem" button the entire time? What if we only showed it if the initial Face ID unlock fails?

I'll enable the button when it fails only. I originally modelled it after the signal lock screen.


> How does this behave on devices without Face ID? (XCode's Simulator lets you test other types of devices, and there's options to fake a matching or unmatching fingerprint / face in the menu bar).

I'll look into touch ID devices but it might fail if it's a different API. Right now I make it so that when you try to enable the feature in settings, if it fails permission check it turns the flag off.

> When you unlock the app using Face ID, the feed you're looking at gets reloaded. Is there a way to avoid this? Maybe by just showing the view over the top of the app instead of using an AppState case?

I did originally start with having it as a `fullScreenCover` but I was running into certain issues for maintaining state. I will revisit it though because that was a while ago and I forget what specific issues.

> Maybe it's possible to hide the content of the app whilst in the app switcher? I've seen other apps do this, but I'm not sure how it's done. It would be nice to have at some point, but I don't mind if it isn't in this PR.

I would like to know how this works as well, I'll look into it after this PR

> At some point it would be nice to have this option available per-account (the option would be inside of Account Settings instead of inside General). Again, that doesn't need to be in this PR.

Will look into after this PR as well

schmeat:
It does with with touch ID with no changes required

Simulator Screenshot - iPhone SE (3rd generation) - 2024-01-17 at 14 18 09

schmeat:
@Sjmarf Changed back to full screen modal, guess my previous issues I had with it are resolved now.

schmeat:
Made it so the screen will show during app switcher now. Should be all the changes for this PR unless you find bugs.

Mlem/Views/Shared/AppLockView.swift
```diff
@@ -0,0 +1,91 @@
...
+    @State var alertMessage: String = ""
+
+    var body: some View {
+        @State var logoOffset: CGFloat = 0
+
+        VStack(spacing: 50) {
+            LogoView()
+                .frame(width: 150, height: 150)
+                .animation(.spring(), value: isAuthenticating)
+                .offset(y: isAuthenticating ? 0 : 300)
```
Sjmarf, line 32: I'm not a huge fan of the up-and-down animation. What do you think about a simple opacity animation instead?

```suggestion
        ZStack {
            LogoView()
                .frame(width: 150, height: 150)
                .opacity((isAuthenticating || biometricUnlock.isUnlocked) ? 0 : 1)
                .animation(.default, value: isAuthenticating || biometricUnlock.isUnlocked)
                .frame(maxHeight: .infinity, alignment: .center)
```

Mlem/Views/Shared/AppLockView.swift
```diff
@@ -0,0 +1,91 @@
...
+                    .frame(width: 200)
+                    .padding(15)
+                    .background(.blue)
+                    .foregroundColor(.white)
+                    .clipShape(RoundedRectangle(cornerRadius: 12))
+                }
+            }
+            .frame(maxHeight: .infinity, alignment: .bottom)
+            .opacity(isButtonHidden ? 0 : 1)
+        }
```
Sjmarf, line 53: Using a `ZStack` instead of a `VStack` (see previous comment) and adding `.ignoresSafeArea` here ensures that the logo remains centered on all devices

```suggestion
                    .clipShape(RoundedRectangle(cornerRadius: 12))
                    .padding(.bottom, 50)
                }
            }
            .frame(maxHeight: .infinity, alignment: .bottom)
            .opacity(isButtonHidden ? 0 : 1)
        }
        .ignoresSafeArea()
```

Mlem/API/APIClient/APIClient.swift
```diff
@@ -80,8 +80,8 @@ class APIClient {
         switch flow {
         case let .account(account):
             session = .authenticated(account.instanceLink, account.accessToken)
-        case .onboarding:
-            // no calls to our `APIClient` should be made during onboarding
+        default:
+            // no calls to our `APIClient` should be made during onboarding or applock
```
Sjmarf, line 84: I think this can be reverted?

```suggestion
        case .onboarding:
            // no calls to our `APIClient` should be made during onboarding
```

Mlem/ContentView.swift
```diff
@@ -30,13 +30,25 @@ struct ContentView: View {
     @State private var tabSelection: TabSelection = .feeds
     @State private var tabNavigation: any FancyTabBarSelection = TabSelection._tabBarNavigation
     @GestureState private var isDetectingLongPress = false
+    @State private var isAppLockAuthenticated: Bool = true
```
Sjmarf, line 33: This is unused
```suggestion
```

Mlem/ContentView.swift
```diff
@@ -30,13 +30,25 @@ struct ContentView: View {
...
+
+    @StateObject var biometricUnlock = BiometricUnlock()
+    
+    var isAppLocked: Bool {
+        if appLock != .disabled, !biometricUnlock.isUnlocked {
+            return true
+        }
+        
+        return false
+    }
```
Sjmarf, line 51: This can be simplified a little

```suggestion
    var isAppLocked: Bool {
        appLock != .disabled && !biometricUnlock.isUnlocked
    }
```

Mlem/Logic/BiometricUnlock.swift
```diff
@@ -0,0 +1,70 @@
...
+//
+//  Created by Sumeet Gill on 2024-01-16.
+//
+
+import Foundation
+import LocalAuthentication
+import SwiftUI
+
+enum BiometricsError: LocalizedError {
+    case unknown
```
Sjmarf, line 13: `unknown` case isn't used.

```suggestion
````

File: Mlem/Logic/BiometricUnlock.swift [hunk 10]
```diff
@@ -0,0 +1,70 @@
...
+    case permissions
+    
+    var errorDescription: String? {
+        switch self {
+        case .permissions:
+            return "Please check Face ID permissions."
+        case .rejected:
+            return "Please try again."
+        case .unknown:
+            return "An unknown error has occured."
```
Sjmarf, line 24: See above

```suggestion
```

Mlem/MlemApp.swift
```diff
@@ -16,7 +16,8 @@ struct MlemApp: App {
     
     @AppStorage("lightOrDarkMode") var lightOrDarkMode: UIUserInterfaceStyle = .unspecified
     @AppStorage("homeButtonExists") var homeButtonExists: Bool = false
-    
+    @AppStorage("appLock") var appLock: AppLock = .disabled
```
Sjmarf, line 19: Unused

```suggestion
```

Mlem/Window.swift
```diff
@@ -21,6 +21,7 @@ struct Window: View {
     @StateObject var recentSearchesTracker: RecentSearchesTracker = .init()
     @StateObject var layoutWidgetTracker: LayoutWidgetTracker = .init()
     @StateObject var appState: AppState = .init()
+    @StateObject var biometricUnlock = BiometricUnlock()
```
Sjmarf, line 24: Unused

```suggestion
```

Mlem/Views/Tabs/Settings/Components/Views/General/GeneralSettingsView.swift
```diff
@@ -113,6 +114,19 @@ struct GeneralSettingsView: View {
...
+                    settingIconSystemName: Icons.appLockSettings,
+                    settingName: "App Lock",
+                    currentValue: $appLock,
+                    options: AppLock.allCases
+                )
+            } header: {
+                Text("Privacy")
+            } footer: {
+                Text("Locks your app with Face ID")
+            }
```
Sjmarf, line 129: I think we should use a toggle here until more options are added

```suggestion
                SwitchableSettingsItem(
                    settingPictureSystemName: Icons.appLockSettings,
                    settingName: "Lock with Face ID",
                    isTicked: Binding(
                        get: { appLock == .instant },
                        set: { selected in
                            appLock = selected ? .instant : .disabled
                        }
                    )
                )
            } header: {
                Text("Privacy")
            }
```

Sjmarf (commented):
Nice! Works great - I didn't find any bugs :D

Mlem/Views/Shared/AppLockView.swift
schmeat: Simulator Screenshot - iPhone 15 Pro - 2024-01-18 at 13 02 02


Maybe I copy pasted it in wrong but it doesn't seem to show the user anything when face ID is doing its challenge. Is that correct? 

I personally like the animation (obviously) because it lets the user know that they're in the Mlem app while touch/face ID is occuring.

Sjmarf: > Maybe I copy pasted it in wrong but it doesn't seem to show the user anything when face ID is doing its challenge. Is that correct?

That's correct, yes. For me, the FaceID animation is only shown for a second. I think the opacity-only animation looks cleaner, personally. It's more in-line with what I've seen before in other apps

Here's a video comparison:

https://github.com/mlemgroup/mlem/assets/78750526/c1816ba2-493e-4429-b58a-1ad78c9f4901

https://github.com/mlemgroup/mlem/assets/78750526/e7d905c7-d49b-4b8c-bf11-84c64e8c4253


Another thought I had - the animation for FaceID incorporates the dynamic island for devices that have it. This means that the animation isn't in the middle anymore. I think we could always show the logo in the middle whilst FaceID is unlocking in this case?

schmeat: Check out my latest changes, it's a hybrid between both of ours. If it has dynamic island then it does your changes, which is fades logo out. Otherwise it moves the logo up slightly. Does that work?

Sjmarf: Sure, I'll check it out when I get home later. 

> If it has Dynamic Island it does your changes

Do we need to animate the logo at all when there's a Dynamic Island? We don't need to move it out of the way of the animation - I haven't tested it yet, but I think we should just be able to leave the logo at 100% opacity in the center rather than animating it.

> Also something to note: `.ignoresSafeArea()` seems to push the button below the bottom bar for iPhone 15 pro.

I put a bottom padding on the button in my review comment suggestion to mitigate this. Sorry, I should have mentioned it 😄

schmeat: It doesn't move out of the way if there's Dynamic Island right now. I have it fading in and out of opacity but that wouldn't be needed either so I can remove that part. 


I'll add padding to the button.

Sjmarf: Hmm, I still prefer the opacity-only animation over the bouncy one - I think the unlock happens too quickly to justify making the logo move. I've used a few apps that have tried various approaches to this, and I've always preferred the animations that are simple and smooth (like Apollo) vs ones with more animation (like Signal). In my opinion we should just use the simpler opacity-only animation

Mlem.xcodeproj/project.pbxproj
```diff
@@ -63,11 +63,10 @@
 		03A276792AFD903600C0D66B /* CommunityModel+MenuFunctions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03A276782AFD903600C0D66B /* CommunityModel+MenuFunctions.swift */; };
 		03A2767B2AFE560000C0D66B /* CommunityModel+SwipeActions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03A2767A2AFE560000C0D66B /* CommunityModel+SwipeActions.swift */; };
 		03A2767D2AFE656700C0D66B /* UserModel+MenuFunctions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03A2767C2AFE656700C0D66B /* UserModel+MenuFunctions.swift */; };
-		03A40DAD2AD5EA11005F019F /* NoPostsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03A40DAC2AD5EA11005F019F /* NoPostsView.swift */; };
```
schmeat, line 66: Not sure why my rebase did this but I still see both of these files in the project

EricBAndrews:
I agree with Sjmarf on the animation. It doesn't animate smoothly due to mainactor logic, and the face id popup itself is a toast, so it doesn't make good design sense to rearrange underlying content around it.

Other than that, the code looks good! This is a nice feature that I'm sure our users will appreciate. Same as with the collapsing child comments PR, I'll test locally and bypass CI once merge conflicts are resolved.

schmeat:
Ok I'll make the changes and rebase as well. Either tonight or tomorrow.

schmeat:
@EricBAndrews Updated to latest dev and removed animation. Should be good to go now.

PR #863, opened 2024-01-17
Fix Account list sorting issue

Sjmarf:
Fixed an issue where the current account wasn't shown at the top when sorting by "Most Recent" with "Grouped" enabled


EricBAndrews (approved):
Nice 👍

PR #865, opened 2024-01-18
Fixed CI builds for forked PRs

JakeShirley:


Mlem/API/APIRequest.swift
```diff
@@ -9,7 +9,7 @@ import Foundation
 
 // MARK: - APIRequest
 
-enum APIRequestError: Error {
+enum APIRequestError: Error
```
Sjmarf, line 12: Unintentional?

Sjmarf (changes requested):

Mlem/API/APIRequest.swift
JakeShirley: Yes! Trying to make sure Xcode error reporting works.

PR #866, opened 2024-01-18
Edit Profile from "Profile" tab

Sjmarf:
Clicking the ellipsis button in the top right of the profile and then "edit" opens a sheet presenting the profile editor.


EricBAndrews (approved):
LGTM 👍

PR #867, opened 2024-01-20
Slur Filter Warning

Sjmarf:
## Background

Lemmy allows instances to set a slur filter. Users cannot create posts that match the filter regex. The user *can* create a comment that matches the regex - however, all matches contained within the comment are replaced with `*removed*`. 

## Implementation

This PR modifies the `PostComposerView` to show a `"xyz" is disallowed` banner if the post matches the slur filter regex. The regex is checked dynamically every time the post content is modified. 

This improves the user experience - previously, the user would just get an `"error: slur"` notification, with no hint of where in the post the slur is. This could have been confusing, especially if a slur was included accidentally as part of another word, or the user was unaware that their instance blocked a certain word (an example would be `"bitch"` on `lemmy.ml`).

<br/>

## What about comments?

Edit: Added for comments too!

Original message:

Should I implement something similar for comments? Lemmy doesn't actually stop the user from posting a comment containing slurs. I'm not sure what the best approach is here. Do we stop the user from posting their comment? Do we still allow them to post, but give them a warning? Or do we not do anything at all?


Mlem/Views/Shared/Composer/PostComposerView.swift
```diff
@@ -103,17 +92,32 @@ struct PostComposerView: View {
...
+                                    if let regex = siteInformation.slurFilterRegex {
+                                        if let output = try regex.firstMatch(in: newValue.lowercased()) {
+                                            titleSlurMatch = String(newValue[output.range])
+                                        } else {
+                                            titleSlurMatch = nil
+                                        }
+                                    }
+                                } catch {
+                                    print("REGEX FAILED")
+                                }
```
EricBAndrews, line 106: *nit*

This logic is reused for the body and comments so it probably makes sense to extract it to `func slurCheck(_ toCheck: String) -> String?` in `SiteInformationTracker` and just have this do `titleSlurMatch = siteInformation.slurCheck(newValue)`

Mlem/Views/Shared/Composer/PostComposerView.swift
```diff
@@ -186,6 +190,19 @@ struct PostComposerView: View {
...
+                                if let regex = siteInformation.slurFilterRegex {
+                                    if let output = try regex.firstMatch(in: newValue.lowercased()) {
+                                        bodySlurMatch = String(newValue[output.range])
+                                    } else {
+                                        bodySlurMatch = nil
+                                    }
+                                }
+                            } catch {
+                                print("REGEX FAILED")
+                            }
```
EricBAndrews, line 204: See above re: duplicate logic

Mlem/Views/Shared/Composer/PostComposerView.swift
```diff
@@ -261,4 +278,36 @@ struct PostComposerView: View {
...
+                )
+                .environment(\.layoutDirection, layoutDirection == .leftToRight ? .rightToLeft : .leftToRight)
+            }
+        }
+        .overlay {
+            if let slurMatch = titleSlurMatch == nil ? bodySlurMatch : titleSlurMatch {
+                ZStack {
+                    Capsule()
+                        .fill(.red)
+                    Text("\"\(slurMatch)\" is disallowed.")
```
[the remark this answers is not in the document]
EricBAndrews: Maybe we even want to centralize this into a struct? It seems a little silly for a single line of text but it would guarantee consistency if we ever change it down the line

Mlem/Views/Shared/Composer/ResponseEditorView.swift
```diff
@@ -58,11 +53,42 @@ struct ResponseEditorView: View {
...
+                            if let regex = siteInformation.slurFilterRegex {
+                                if let output = try regex.firstMatch(in: newValue.lowercased()) {
+                                    slurMatch = String(newValue[output.range])
+                                } else {
+                                    slurMatch = nil
+                                }
+                            }
+                        } catch {
+                            print("REGEX FAILED")
+                        }
```
EricBAndrews, line 68: See above re: duplicate logic

EricBAndrews (changes requested):
Looks good! I've left a couple minor suggestions. It also appears that the slur filter does not get reset when switching to an instance with no slur filter:
- Switch to a `lemmy.ml` account (or an account on any instance with slur filter)
- Switch to `test-mlem` (or an account on any instance without a slur filter)
- Try to compose a post with a slur
The slur filter from the previous instance will still be applied

Mlem/Views/Shared/Composer/PostComposerView.swift
Sjmarf: Sure, good idea 👍

Sjmarf:
> also appears that the slur filter does not get reset when switching to an instance with no slur filter

Fixed!

Mlem/Views/Shared/Composer/PostComposerView.swift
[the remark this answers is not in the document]
Sjmarf: Hmm, I'd rather use a static property of `AppConstants` like we do for some destructive actions if we want to keep the text the same. If we every get around to localisation, we can just use the same localisation key in both places.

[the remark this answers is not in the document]
EricBAndrews: Yeah, that makes sense to mee

EricBAndrews (approved):
LGTM!

PR #868, opened 2024-01-20
Fix Post Subscribed Indicator

Sjmarf:
Wasn't showing at all - probably because of my `CommunityView` rewrite


Sjmarf:
Fixed in #874, closing

PR #869, opened 2024-01-21
Instance Statistics

Sjmarf:
Added a "Details" tab to the Instances page, which is similar to the community Statistics tab. In addition to the instance stats, some of the instance's preferences are also shown. Which preferences are shown is dependent on whether Developer Mode is enabled or not. Let me know if you disagree with any of these choices.

<table>
  <tr>
    <th>Statistics</th>
    <th>Dev Mode Disabled</th>
    <th>Dev Mode Enabled</th>
  </tr>
   <tr>
      <td></td>
      <td></td>
      <td></td>
  </tr>
</table>

## Other changes

- Added community creation date to Community Stats, and refactored `CommunityStatsView`
- Refactored `UserView` a little (no UI changes)


Mlem/Views/Shared/Instance/InstanceStatsView.swift
```diff
@@ -0,0 +1,270 @@
...
+                        "Reports Email Admins",
+                        systemImage: Icons.moderationReport,
+                        value: instance.reportsEmailAdmins ?? false
+                    )
+                }
+                .frame(maxWidth: .infinity)
+                .background(Color(uiColor: .secondarySystemGroupedBackground))
+                .cornerRadius(AppConstants.largeItemCornerRadius)
+            }
+
```
EricBAndrews, line 205: Currently this view leaves a block of white space at the bottom:

Screenshot 2024-01-27 at 8 18 33 PM

I like having a little bit of overscroll, but imo that space should definitely be gray as well

EricBAndrews (changes requested):
Looks good! A couple minor notes:

- I think the detailed instance info should be visible even when dev mode is off--I think anybody who's actually browsing through an instance info page is going to want to see all the info we can give them
- The instance administrators page is currently not visually consistent with community moderators. Maybe we can centralize that view?

Mlem/Views/Shared/Instance/InstanceStatsView.swift
```diff
@@ -0,0 +1,270 @@
...
+                )
+                if developerMode, let regex = instance.slurFilterString {
+                    Divider()
+                    Text(showingSlurRegex ? regex : "Tap to show slur filter regex")
+                        .textSelection(.enabled)
+                        .font(.footnote)
+                        .foregroundStyle(.secondary)
+                        .padding(12)
+                        .onTapGesture {
+                            showingSlurRegex.toggle()
```
EricBAndrews, line 165: This might want a `.withAnimation` just to smooth it out a little

EricBAndrews (changes requested):

Mlem/Views/Shared/Instance/InstanceStatsView.swift
Sjmarf: Making that space gray means adding a duo-tone background, which we had issues with in the Community View?

Mlem/Views/Shared/Instance/InstanceStatsView.swift
Sjmarf: Sure

Sjmarf:
> I do think the detailed instance info should be visible even when dev mode is off--I think anybody who's actually browsing through an instance info page is going to want to see all the info we can give them

Here's my reasoning for why I hid certain settings:

- "Default Feed" could confuse the average user, because it only applies to Lemmy-ui and not to Mlem. Instead of hiding "Default Feed", I suppose another option would be to add an `info.circle` tooltip that clarifies it?
- "Federation Signed Fetch" is a technical option - I barely know what it does, but some instances have it enabled and some don't. 99% of our users won't care about this, so I hid it.
- I hid the slur filter itself for non-dev accounts because some instances use it to block what I assume is illegal content. `lemmy.world` has URLs in there, and a string of base-64 that's probably a code for something. What's your opinion on this?

I agree it makes sense to always show the other currently hidden settings though ("Show Mod names in Modlog", "Applications email admins" and "Reports email admins")

Mlem/Views/Shared/Instance/InstanceStatsView.swift
Sjmarf: Also, I switched from a gray background to a hard cut-off for the Community stats page in #879. I think the styiling on this pages should match, ideally.

EricBAndrews: Yeah, this comment's outdated--reviewed this PR before 879, edited my original review but forgot to delete this comment 🤦

EricBAndrews:
We could maybe hide those in an "advanced" disclosure group--an instance admin or someone might still want to see them, but I agree that it doesn't make sense to put them top-level for the average user. Alternatively we can rename "developer mode" to "power user mode" or something similar, since "developer" carries the implication that the option will only reveal things that are salient to Mlem developers.

Re: default feed, what if we just label it "Default Feed (Desktop)"?

Sjmarf:
> Re: default feed, what if we just label it "Default Feed (Desktop)"?

Sure, sounds good 👍 

To be honest we could just not show the Federation Signed Fetch at all... I'm yet to find an instance that has it enabled. That just leaves the question of what to do with the slur filter. Just show the option to reveal it all of the time?

EricBAndrews:
Slur filter is publicly available information, so I don't think it's really on us to decide whether it's something we should expose or not. It might be worth putting some clickwrap on it to the effect of "this is a regex for detecting slurs, so it's probably going to contain some pretty nasty language" just to be safe.

Sjmarf:
Done 👍

Mlem/Views/Shared/Instance/InstanceDetailsView.swift
```diff
@@ -0,0 +1,272 @@
...
+                )
+                if let regex = instance.slurFilterString {
+                    Divider()
+                    VStack(alignment: .leading, spacing: 2) {
+                        if showingSlurRegex {
+                            Text(regex)
+                            .foregroundStyle(.secondary)
+                            .textSelection(.enabled)
+                        } else {
+                            Text("Tap to slow slur filter regex.")
```
EricBAndrews, line 155:
```suggestion
                            Text("Tap to show slur filter regex.")
```

EricBAndrews (changes requested):
One typo, other than that LGTM 👍

EricBAndrews (approved):
Nice!

PR #870, opened 2024-01-21
Custom Account Sort Order

Sjmarf:
Adds a "Custom" sort order to the accounts list. Tap and hold to rearrange. Reordering only works in Settings and the profile, not in the quick switcher. You can't have "Grouped" enabled when using the "Custom" sort mode.

I've changed the default sort mode from "Name" to "Custom". "Custom" is also now used instead of "Name" if you have only 2 accounts.


EricBAndrews (approved):
Nice 👍

PR #872, opened 2024-01-21
Rearrange Settings

Sjmarf:
This PR moves some settings to different menus. It also includes some minor design Settings design touch-ups. 

## Account -> Local Options

This new section contains the settings that are stored per-account locally within Mlem. Currently, this constitutes the Account Nickname setting (which has been moved from Appearance -> Tab Bar) and the "Delete Community Favorites" button (which has been moved from General). 

Because the account nickname is visible in the account switcher in addition to the tab bar, I think it fits better in Account Settings than Tab Bar settings.

In future, when we add an option for per-account Face ID Lock, that setting will go here as well. 

## Accounts -> Quick Switcher

I've moved the "Swipe up for Account Switcher" option from Appearance -> Tab Bar to a new "Quick Switcher" submenu inside of the Accounts page. It's not an appearance setting, so it doesn't make sense for it to be under Appearance (I knew this when I added it, but couldn't be bothered to make a new menu at the time lol). This page lets you enable/disable the long press and swipe activation methods, and plays short animations showing how to use each gesture.

<table>
  <tr>
    <td width=50%></td>
    <td><video src="https://github.com/mlemgroup/mlem/assets/78750526/9ab25d78-c6c9-46bd-a810-ab8e736b4a2c.mp4"></td>
   </tr> 
</table>

This has the added benefit of letting the user know that the Long Press gesture exists - we didn't tell the user about this anywhere previously.

If the user only has two accounts, the "Quick Switcher" `NavigationLink` is replaced by a tip that tells you about the long-press gesture:

https://github.com/mlemgroup/mlem/assets/78750526/2f04470a-da06-4ad8-ac35-dd997568516c

I'm aware that these animations are overkill lol, but I wanted to play with the new iOS 17 `PhaseAnimator`. This does have the unfortunate side effect of the animations not being visible pre-iOS 17 though.

## Tab Bar Appearance Settings

As mentioned above, the nickname field has been moved from this page to the new "Local Options" page, and the "Swipe up for Account Switcher" option has been moved to Accounts.

I've redesigned the Profile Icon Appearance setting here, too:

You may notice that I've removed the "username" option, and set "nickname" as the new default. My rationale for this is that if the user sets a nickname, they probably want it to be consistent everywhere in Mlem. So, the "username" setting didn't make much sense to me - why hide a nickname that you've chosen to set? Lmk if you disagree with this change.


EricBAndrews (approved):
LGTM 👍

PR #873, opened 2024-01-22
Search Instances

Sjmarf:
> [!note]
> This PR is built ontop of #869. Drafting until that PR is merged.

## Instance Search

Adds an "instances" tab to search. When the app opens, the instance list is downloaded from https://github.com/mlemgroup/mlem-stats. That repo has a GH action set up to refresh the data every 24 hours. So if an instance changes their avatar or version number, it could take 48 hours for it to be updated in Mlem (lemmyverse's 24h maximum, followed by our 24h maximum). Instances that have a negative score, have less than 20 users or are censured on fediseer are excluded. This cuts the list down from ~850 to ~160 instances.

The search tab code is a little clunky, and will need to be re-done as a whole at some point. I'll do that post-iOS 17, at the same time I do post searching probably.

What do we do about NSFW instances? There's currently no way to tell if an instance is NSFW-only, although [there's a PR open for this](https://github.com/LemmyNet/lemmy/pull/4393).


Mlem/Views/Tabs/Search/SearchResultsView.swift
```diff
@@ -44,14 +44,14 @@ struct SearchResultsView: View {
...
-            }
-            .animation(.default, value: contentTracker.isLoading)
+//            Group {
+//                if contentTracker.isLoading && contentTracker.page == 1 && !shouldLoad {
+//                    ProgressView()
+//                        .padding(.trailing)
+//                        .transition(.opacity)
+//                }
+//            }
+//            .animation(.default, value: contentTracker.isLoading)
```
EricBAndrews, line 54: Dead code?

Mlem/Views/Tabs/Search/SearchView.swift
```diff
@@ -75,12 +77,26 @@ struct SearchView: View {
             .autocorrectionDisabled(true)
             .textInputAutocapitalization(.never)
             .onAppear {
```
EricBAndrews, line 79: *nit*

This can be a `.task` since it's all async

EricBAndrews (changes requested):
Looks good! A couple nits, and I also noticed some inconsistent behavior around searching:

https://github.com/mlemgroup/mlem/assets/44140166/2e341ec6-c495-4d07-b102-e5c1265c1f60

It's minor enough that I don't see a need to hold up the merge to resolve it, but if we merge with this behavior still present it's worth opening an issue.

Sjmarf:
Nice catch! Should be fixed 👍

EricBAndrews (approved):
Nice! 🚀

PR #874, opened 2024-01-22
New Post Tracker

EricBAndrews:
# Checklist
      - Progress towards #73 

# Pull Request Information

This PR rewrites the `PostTracker` to inherit from the generic `StandardTracker` and moves the saved feed out of `UserView` and into `FeedsView`. It also refactors `FeedsView` to have a cleaner navigation pattern.

## Note To Reviewers

This PR touches a massive amount of the codebase, but doesn't change much behavior. Almost all of the significant new or modified code is contained within `StandardPostTracker`, `PostFeedView`, `FeedsView`, and the generic trackers. Many of the other diffs are simple name or type changes.

### Trackers

Generic trackers now have support for cursor-based pagination and have been lightly refactored to support this and improve readability.

### PostModel

PostModel is now a class, and all of its logic has been moved into it. This also allows `PostTracker` to be removed from many views where it was passed in exclusively to handle post interactions.

### Filtering

Previously, the `PostTracker`'s filter was a single function created at initialization that returned filter reasons for bookkeeping. Now, the `PostTracker` maintains a list of filters to apply that can be dynamically modified; newly fetched posts are evaluated against every tracked filter, the number of filtered items is recorded, and the filtered list is returned from the `fetch` methods.

### FeedType

`FeedType` has been fully decoupled from `APIListingType`. `FeedType` now exclusively refers to an internal feed type, and `APIListingType` matches the API specification.

### Navigation

Several unused navigation routes have been removed.

## Todo

The saved feed currently does not contain saved comments. I will open a followup PR soon to address this.

Tap tab to navigate has been removed per Slack. I have opened issue #875 to track this.


boscojwho:
Looks like community list view (`CommunityFeedView`?) isn't properly insetted under the tab bar.

EricBAndrews:
> Nice! I'll do a proper review of this tomorrow. A couple of things I noticed during some quick testing:
> 
>     * The feed refreshes when switching tabs in `CommunityView`.
> 
>     * If you set Saved as your default feed in Settings and then re-launch the app, the Saved feed appears but the ellipsis menu in the top-right isn't shown. I also had this happen once when navigating to Saved from the community list, but haven't been able to repro it.



> Nice! I'll do a proper review of this tomorrow. A couple of things I noticed during some quick testing:
> 
>     * The feed refreshes when switching tabs in `CommunityView`.
> 
>     * If you set Saved as your default feed in Settings and then re-launch the app, the Saved feed appears but the ellipsis menu in the top-right isn't shown. I also had this happen once when navigating to Saved from the community list, but haven't been able to repro it.

Looks like those both came from the same bug in initialization. Should work now.

Sjmarf:
Now when I open the app, I get an error haptic and it crashes immediately. I'm still using Saved as the default feed. Not sure if that affects it

It only does it about  half of the time on WiFi, but happens every single time when I use (pretty bad) mobile data 🤔 Possibly related to internet speed?

Mlem/API/Models/ListingType.swift
```diff
@@ -11,42 +11,5 @@ enum APIListingType: String, Codable {
     case all = "All"
     case local = "Local"
     case subscribed = "Subscribed"
-    
-    // Pre 0.18.0 it appears that they used integers instead of strings here. We can remove this intialiser once we drop support for old versions. To fully support both systems, we'd also need to *encode* back into the correct integer or string format. I'd rather not go through the effort for instance versions that most people don't use any more, so I've disabled the option to edit account settings on instances running <0.18.0
-    // - sjmarf
-    
-    // TODO: 0.17 deprecation remove this initialiser
-    
-    init(from decoder: Decoder) throws {
```
Sjmarf, line 20: We can't remove this decoder just yet. 

Inside of `MyUserView` we have to decode a `ListingType` in inteher form on pre-0.18.0 versions. This decoder handles that. We can't just exclude the property entirely, unfortunately - due to a bug in 0.18.0 we need to send the `ListingType` when we save the user's settings - even if they aren't modifying that setting. This bug is fixed in 0.19.0.

If we remove this a decoding error is thrown and the feed won't load on test-mlem.jo.wtf (which runs 0.17.3)

Mlem/Models/Content/Post Model.swift
```diff
@@ -5,26 +5,35 @@
...
+    @Dependency(\.errorHandler) var errorHandler
+    @Dependency(\.postRepository) var postRepository
+    
+    var postId: Int
+    var post: APIPost
+    var creator: UserModel
+    var community: CommunityModel
+    @Published var votes: VotesModel
+    var commentCount: Int
+    var unreadCommentCount: Int
```
Sjmarf, line 24: Do we want the unread comment count to disappear once you've clicked on a post? `unreadCommentCount` needs to be `@Published` if so

Mlem/Views/Tabs/Feeds/Components/NoPostsView.swift
Sjmarf, line 1: In the "Saved" feed, sort-mode-related prompts can still be shown. We shouldn't do this. Repro:

- Switch to "Top of today" and enable "hide read"
- Go to saved posts
- Observe no posts message

Sjmarf, line 1: Also, this view flashes briefly before the loading animation appears when switching sort modes in the feed.

Mlem/Views/Tabs/Feeds/Components/PostFeedView+Logic.swift
```diff
@@ -0,0 +1,22 @@
...
+//  Mlem
+//
+//  Created by Eric Andrews on 2024-01-21.
+//
+
+import Dependencies
+import SwiftUI
+
+extension PostFeedView {
+    func setDefaultSortMode() async {
```
Sjmarf, line 12: The default-sort-mode-picking logic appears to be broken - Mlem incorrectly uses the default sort mode rather than the fallback on <0.19 instances.

Sjmarf (changes requested):
Code looks good :D I found a couple of bugs

Sjmarf:
We had a fade-in transition on the posts in the feed before this PR. Would it be possible to bring that back? We still have it in `UserView`, but not in the feeds

Mlem/Views/Tabs/Feeds/Components/NoPostsView.swift
EricBAndrews: Saved feed needs its own view--what's in there right now is a stub for navigation and such. As soon as the bugs are squashed on this PR I'm going to get to work on that, so until then as long as saved is stable and the navigation works that's all that's required

Mlem/Views/Tabs/Feeds/Components/PostFeedView+Logic.swift
EricBAndrews: That *should* be fixed now

Mlem/Views/Tabs/Feeds/Components/NoPostsView.swift
EricBAndrews: Fixed

EricBAndrews:
> We had a fade-in transition on the posts in the feed before this PR. Would it be possible to bring that back? We still have it in `UserView`, but not in the feeds

Nice catch! Added it back in

boscojwho:
Screenshot 2024-01-23 at 6 53 31 PM

There appears to be a missing call to `.handleLemmyLinkResolution(...)`.

Mlem/Views/Tabs/Feeds/Components/PostFeedView.swift
```diff
@@ -0,0 +1,167 @@
...
+                        }
+                    }
+                }
+            }
+    }
+    
+    var content: some View {
+        LazyVStack(spacing: 0) {
+            if postTracker.items.isEmpty || versionSafePostSort == nil {
+                noPostsView()
```
Sjmarf, line 99: Thoughts on adding a little more padding here?

```suggestion
                noPostsView()
                    .padding(.top, 50)
```

Mlem/Views/Tabs/Feeds/Components/PostFeedView+Logic.swift
Sjmarf: Nice, looks like it's working now 👍

Mlem/Views/Tabs/Feeds/Feed Types/AggregateFeedView.swift
```diff
@@ -0,0 +1,169 @@
...
+                            
+                        Text(subtitle)
+                            .font(.footnote)
+                            .foregroundStyle(.secondary)
+                    }
+                    .frame(height: 44)
+                    .frame(maxWidth: .infinity, alignment: .leading)
+                }
+                .padding(.vertical, 5)
+
```
Sjmarf, line 141: Can we pad this a little more to make it consistent with the community view? 
```suggestion
                        .padding(.bottom, 3)
```

Sjmarf (changes requested):
I get a small flicker when loading feeds in certain situations, now. I think this was introduced since last review. See videos

https://github.com/mlemgroup/mlem/assets/78750526/b84e9bf3-44e7-466c-b920-3a549958d977

https://github.com/mlemgroup/mlem/assets/78750526/aa6afafc-c084-453c-ba80-89a88dad2529


I still get that single-tap haptic whenever I switch accounts, or when first opening the app.

Mlem/Views/Tabs/Feeds/Components/PostFeedView+Logic.swift
Sjmarf: Nice, looks like it's working now 👍

Sjmarf:
The moderator badge doesn't appear to be shown in the post feed anymore

Mlem/Views/Tabs/Feeds/Feed Types/SavedFeedView.swift
Sjmarf, line 1: Switching to "Saved" via the drop-down in the feed header causes the `noPostsView` to appear briefly whilst the posts are transitioning in. This only happens on the Saved feed.

Mlem/Views/Tabs/Feeds/Feed Types/CommunityFeedView.swift
Sjmarf, line 1: Posts aren't refreshed when switching to "Hot" sort mode via the "Switch to hot" `noPostsView` button. 

Repro:

- Set "top of hour" as default
- Visit a community with no posts in last hour
- Tap "switch to hot"
- See "no posts" message. This shouldn't happen - instead, the posts should be shown
- Switch to "About" tab
- Switch back to "Posts" tab
- Posts load correctly

Sjmarf (changes requested):

EricBAndrews:
> The moderator badge doesn't appear to be shown in the post feed anymore
> 
> And does the large "Communities" navigation title on the feeds list make sense here anymore? "Saved" isn't a community. Does "Feeds" make more sense?

"Feeds" definitely makes more sense, changed.

Also fixed the moderator flair. Turns out the CommunityModel bundled in UserModel doesn't have that info!

Sjmarf (approved):
Looks good! I was testing it today and didn't find any other issues.

PR #877, opened 2024-01-26
iOS 16 symbol fix

Sjmarf:
The bubble symbol alias is new in iOS 17. Switched to bubble.left (different name for the same icon) which is supported on both iOS 16 and iOS 17.


EricBAndrews (approved):
🚀

PR #878, opened 2024-01-27
Ban Users

Sjmarf:
Adds the option for administrators to ban users. Requested on mlemapp [here](https://lemmy.ml/post/11013684). In future, the same sheet will be used to allow moderators to ban users from communities. 

Here's the equivalent modal in lemmy-ui:


EricBAndrews (approved):
Nice! Great work on the sheet, very easy to use.

PR #879, opened 2024-01-27
Community View Tweaks

Sjmarf:
Replaced the duo-tone background with a single-color background in all of the Community tabs. The duo-tone background had visual issues when scrolling too far up/down, and was broken visually on iPad


EricBAndrews (approved):
Nice!

PR #882, opened 2024-01-27
Link context actions

Sjmarf:
Thoughts on this? It makes it easier to copy the link, but reduces the area for which a long press opens the _post's_ context menu options. Another option would be to put these buttons in the post's context menu, but I think that would make the menu too long.

If you like the idea, would it also be worth doing something similar for "save" and "share" actions on images in posts?


EricBAndrews (approved):
I like it! The tappable links are distinct enough elements that I don't think the UX will be unintuitive.

PR #883, opened 2024-01-28
Swipe actions activation angle

Sjmarf:
Previously, a post's swipe actions could be activated by swiping at 45° from the horizontal. In my opinion, this made it too easy to accidentally activate the swipe actions whilst scrolling. I've changed this such that the swipe actions now only activate when within ~30° from the horizontal. Thoughts?


EricBAndrews (approved):
Nice! Feels about right to me

PR #884, opened 2024-01-29
Instance Uptime

Sjmarf:
Adds an "Uptime" tab to instance pages. Uptime data is fetched from https://lemmy-status.org. Lemmy-status watches a set list of instances - if an instance isn't on that list, the tab won't be shown. I'm storing a list of supported instances as a static property within Mlem - if lemmy-status ever changes their list of supported instances, we'd have to update ours too. 

The data refreshes automatically every 30 seconds.

When the instance page is opened, Mlem attempts to contact the instance for site information. If this request fails and the instance is supported by lemmy-status, it'll attempt to show the uptime view instead of `ErrorView`.


Mlem/Views/Shared/Instance/InstanceUptimeView.swift
```diff
@@ -0,0 +1,248 @@
...
+                if uptimeData.results.filter(\.success).count < 15 {
+                    if mostRecentOutage.endTime == nil {
+                        HStack(spacing: 5) {
+                            (Text("\(instance.name) is ") + Text("offline").foregroundColor(.red))
+                                .font(.title2)
+                                .fontWeight(.semibold)
+                            Image(systemName: Icons.uptimeOffline)
+                                .foregroundStyle(.red)
+                        }
+                        footnote("Outage started \(mostRecentOutage.startTime.getRelativeTime()).")
```
EricBAndrews, line 94: *nit*

This block gets duplicated three times--maybe move to a helper function?

Mlem/Views/Shared/Instance/InstanceUptimeView.swift
```diff
@@ -0,0 +1,248 @@
...
+            .padding(.top, 20)
+            footnote("Mlem will refresh this automatically every 30 seconds.")
+            .padding(.top, 8)
+            .padding(.leading, 6)
+            .padding(.bottom, 30)
+            section("Response Time") {
+                VStack(alignment: .leading, spacing: 4) {
+                    responseTimeChart
+                        .padding(.horizontal, 20)
+                    footnote("Mean average: \(uptimeData.results.map(\.durationMs).reduce(0, +) / uptimeData.results.count)ms")
```
EricBAndrews, line 44:
```suggestion
                    footnote("Average: \(uptimeData.results.map(\.durationMs).reduce(0, +) / uptimeData.results.count)ms")
```

Mlem/Views/Shared/Instance/InstanceUptimeView.swift
```diff
@@ -0,0 +1,248 @@
...
+                        .padding(.leading, 20)
+                }
+                .padding(.top, 17)
+                .padding(.bottom, 8)
+            }
+            section("Incidents", spacing: 0) {
+                ForEach(uptimeData.downtimes) { event in
```
EricBAndrews, line 51: Maybe only show outages from the past day, and hide the rest in a disclosure group? As-is it's a super long stack of low-impact data. It might also be valuable to add a filter to show only medium to high severity incidents?

Mlem/Icons.swift
```diff
@@ -149,6 +149,12 @@ enum Icons {
     static let close: String = "multiply"
     static let cakeDay: String = "birthday.cake"
     
+    // uptime
+    static let uptimeOffline: String = "xmark.circle.fill"
+    static let uptimeOnline: String = "checkmark.circle.fill"
+    static let uptimeShortOutage: String = "ellipsis.circle.fill"
```
EricBAndrews, line 155: This symbol choice feels odd. I initially tapped it thinking it would be a button that would show me more info or something like that. Exclamation mark wouldn't be great, though--we could maybe go the Jenkins route and use cloud/rain/thunder for severity?

EricBAndrews (changes requested):
Looks good! Super useful feature for a federated social media app.

Left a couple minor notes on design/wording.

Mlem/Views/Shared/Instance/InstanceUptimeView.swift
Sjmarf: That could work. 

I'm not exactly sure how much I trust the incidents list to be honest - [lemmy-status has been reporting programming.dev as unhealthy for 3 days](https://lemmy-status.org/endpoints/_programming-dev), but it's working just fine. Beehaw has a similar issue 🤔 

What do we do in that situation?

Mlem/Icons.swift
Sjmarf: Sure, I think the exclamation mark would look good 👍

Mlem/Views/Shared/Instance/InstanceUptimeView.swift
EricBAndrews: I think augmenting the footer to read "Uptime data fetched from lemmy-status.org, and may contain inaccuracies" would be sufficient. If we wanted to be a little fancier we could add some logic to detect discrepancies between the most recent incident and the current status, and if one is detected add a footer to that incident reading "Discrepancy detected" with a tappable element to open a popup explaining that the incident and the current active status conflict, and that the user should trust the active status.

Sjmarf: Hmm. Maybe it would be better to just hide the faulty incident if it's in conflict with the actual status?

EricBAndrews: That works for me

Sjmarf: How's this?

EricBAndrews:
Nice 👍

EricBAndrews (approved):

Issue #124, opened 2023-06-22
Unified Logging Solution


EricBAndrews:
Create a better logging solution than just haphazardly printing things

schmeat:
Have you guys considered the official swift logger? 

https://developer.apple.com/documentation/os/logger

PR #888, opened 2024-02-01
Unified Logging

schmeat:
# Checklist
      - #124 

# Pull Request Information

## About this Pull Request

This is just a proof of concept for logging right now, with minimal files edited. Want to get consensus before I modify every file with logging. @Sjmarf @EricBAndrews 

## Additional Context

N/A for now

### DIFF
--- a/Mlem/Logic/MlemLogger.swift
+++ b/Mlem/Logic/MlemLogger.swift
@@ -1,0 +1,21 @@
+//
+//  MlemLogger.swift
+//  Mlem
+//
+//  Created by Sumeet Gill on 2024-01-31.
+//
+
+import Foundation
+import OSLog
+
+class MlemLogger {
+    private let logger: Logger
+    
+    required init(file: String = #fileID) {
+        self.logger = Logger(subsystem: AppConstants.loggerSubSystem, category: file)
+    }
+    
+    func log(level: OSLogType, message: String) {
+        logger.log(level: level, "\(message)")
+    }
+}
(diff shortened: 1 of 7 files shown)


Mlem/Logic/MlemLogger.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+
+import Foundation
+import OSLog
+
+class MlemLogger {
+    private let logger: Logger
+    
+    required init(file: String = #fileID) {
+        self.logger = Logger(subsystem: AppConstants.loggerSubSystem, category: file)
```
schmeat, line 15: I made the category be the file name for now. Not sure if we should have defined categories for logging, such as "user interaction", "networking", etc. It may be a lot to maintain and enforce going forward if we do that.

schmeat:
It would be nice to have a linting rule to warn people using `print` to use `MlemLogger` instead. I'm not sure how to do that in current project since I don't see a swiftlint markup file.

EricBAndrews:
I like the concept! I'm tentatively scoping this for 1.3--pending the face ID PR, we're in full bugfix mode for the 1.2 release.

Re: categorization, I think that would be a good idea--there's a lot of discussion to be had about what those categories should be, but the added dev time categorizing logs has in my experience always paid off eventually.

We might want to look into an architecture like [this one](https://www.avanderlee.com/debugging/oslog-unified-logging/), where categorized loggers are defined as static properties on the Logger class--that saves us from needing to dependency inject or instantiate loggers everywhere

schmeat:
Yeah makes sense with categorization. I didn't make it statically defined because I have the file name coming in through the init. Once we define the categories, I won't need to do that.

PR #890, opened 2024-02-02
Warnings Cleanup

EricBAndrews:
# Checklist
      - *list issue(s) here*

This PR removes as many warnings as can be cleanly fixed without serious refactoring:
- Removes duplicate `CommunityListSection` build file declaration
- Changes unmodified `var`s to `let`s
- Removes unused `vars` and `lets`
- Removes deprecated `.apiPerson` navigation route
- Removes deprecated `APIPerson` in `UserModel`
- Migrates inbox models to use `UserModel` and `CommunityModel` instead of `APIPerson` and `APICommunity`
- Removes unused `UserModeratorView`

The outstanding warnings are:
- `UserView+Logic` uses deprecated `PersonRepository` methods. These will be removed when I update that view to use the shiny new `UserContentTracker`.
- `UserFeedView` has a manual `Hashable` conformance. This will also be removed when I update that view to use `UserContentTracker`
- `UserView` uses `StandardPostTracker.reset`. Ditto above.
- `PersonRepository.loadUserDetails` from URL uses its own deprecated `loadUserDetails` method. This should be replaced an analagous `loadUser` and `loadUserContent`, perhaps with a `loadUserAndContent` that returns a `UserModel`, `UserContentModel` tuple.
- `CommentBodyView` makes a deprecated call to `UserLinkView` using `APIPerson` instead of `UserModel`. This will require a middleware `CommentModel` that replaces `APICommentView` internally to fix.
- `CommunityModel` still contains an `APICommunity` which is needed for the `FavoritesTracker`--some care will need to be taken to update this without losing stored favorites.
- `PostComposerView` uses the deprecated `APIPerson` initializer for `UserLabelView`. This will require a middleware model replacing `APIMyUserInfo` in the `SiteInformationTracker` to resolve.


Sjmarf (approved):
Nice! 😄

PR #894, opened 2024-02-03
Fix blocking communities from posts

### LINKED ISSUE #889: Blocking communities from posts doesn't work
Per title--the embedded `CommunityModel` in posts has blocked as `nil`, so the `toggleBlock` function doesn't do anything.

Sjmarf:
Closes #889


EricBAndrews (approved):
Nice 🐛🔨

PR #895, opened 2024-02-03
Fix cutoff image tap area

Sjmarf:
Fixed a bug where an image in a large post that was cutoff due to the "limit image height in feed" setting would have a tap area that extended outside of the image. It appears that we previously fixed this using a `.overlay()`, but that solution doesn't work anymore for whatever reason (probably iOS 17)


Sjmarf:
Nevermind, it turns out my fix doesn't work properly when the image _doesn't_ need to be cutoff xD

After some further testing, it seems that the hard cutoff only occurs when you toggle the "limit image height in feed" setting on/off. If you reload the feed, it scales the image rather than cutting it off. So I don't think we particularly need to fix this

PR #897, opened 2024-02-03
Fix "Show n replies" button tap area

Sjmarf:
Previously you had to tap on the "show n replies" text to expand the replies, and tapping elsewhere on the button wouldn't work. Now, you can tap anywhere on the "show n replies" button to expand


EricBAndrews (approved):
👍

PR #899, opened 2024-02-04
Bubble picker scroll-on-tap

Sjmarf:
Minor improvement to the bubble-style picker that we use for tabs in communities/profiles/search. If you tap on a tab that is partially outside of the screen, it will now scroll so that the tab is fully visible. 

Currently, we only have enough tabs for this on a moderator's profile page.

https://github.com/mlemgroup/mlem/assets/78750526/c9c026ee-1f15-4289-97a3-3fe5cb8489c1


EricBAndrews (approved):
Nice! Huge quality of life improvement

PR #908, opened 2024-02-05
Updated search bar prompt

Sjmarf:
We now support searching instances, so I changed the search bar prompt from "Search Communities & Users" to "Communities, Users & Instances".

When we add post/comment searching, the prompt obviously wouldn't fit anymore. We'd have to go the Apple Music Route and write "Communities, Posts & More"


boscojwho (approved):
😇

PR #909, opened 2024-02-05
More Context Menu on links/images

Sjmarf:
Follow up on #882. I don't mind whether this goes in 1.2 or is left for 1.3.

- Adds "Save" and "Share" actions to images in the feed.
- Added "Open", "Copy" and "Share" actions to link thumbnails in compact mode.
- Added "Open" action to links on large posts.
- Links now show the contents of the website in the context menu. I was hoping to make the preview tap-to-open, like in Safari. Unfortunately, I can't work out how to do this. So I settled for an "Open" button instead.


boscojwho:
I don't see an "Open" action on links in large post (feed and post view)

Sjmarf:
Yep, I forgot to commit it initially but the committed it later

Mlem/Views/Shared/Cached Image.swift
```diff
@@ -13,11 +14,14 @@ import QuickLook
...
 
 struct CachedImage: View {
+    @Dependency(\.notifier) var notifier
     let url: URL?
     let shouldExpand: Bool
+    let hasContextMenu: Bool
     
     // state vars to track the current image size and whether that size needs to be recomputed when the image actually loads. Combined with the image size cache, this produces good scrolling behavior except in the case where we scroll past an image and it derenders before it ever gets a chance to load, in which case that image will cause a slight hiccup on the way back up. That's kind of an unsolvable problem, since we can't know the size before we load the image at all, but that's fine because it shouldn't really happen during normal use. If we really want to guarantee smooth feed scrolling we can squish any image with no cached size into a square, but that feels like squishing a lot of images for the sake of a fringe case.
     @State var size: CGSize
+    let fixedSize: CGSize?
```
boscojwho, line 24: Request some documentation for `let fixedSize`, seems to be used to set frame but also absence of it means previews aren't available?

Mlem/Views/Shared/Cached Image.swift
```diff
@@ -112,15 +126,30 @@ struct CachedImage: View {
...
-                    imageView
-                }
+                    .if(shouldExpand) { content in
+                        content
+                            .onTapGesture {
+                                imageDetailSheetState.url = url // show image detail
+                                onTapCallback?()
+                            }
+                    }
+                    .if(hasContextMenu && fixedSize == nil) { content in
```
boscojwho, line 136: Propose combining the two context menu view modifiers, and moving the preview logic into a view builder function for clarity/readability:
```swift
  .if(hasContextMenu) { content in
                        content
                            .contextMenu {
                                contextMenuActions(image: image)
                            } preview: {
                                contextMenuPreview(image: image)
                            }
                    }
``` 
and
```swift
@ViewBuilder
    private func contextMenuPreview(image: Image) -> some View {
        if fixedSize != nil {
            image
                .resizable()
                .onTapGesture {
                    imageDetailSheetState.url = url
                    onTapCallback?()
                }
                .transition(.identity)
        } else {
            EmptyView()
        }
    }
```

boscojwho (changes requested):

boscojwho:
> Yep, I forgot to commit it initially but then committed it later. Should work now

Looks good, requested two minor changes not related to actual functionality, see if they make sense =)

Tested using `headline`, `compact`, and `large` post modes on a variety of links (news, YouTube, Lemmy, Mastodon), all seemed to open properly in in-app Safari view.

Mlem/Views/Shared/Cached Image.swift
Sjmarf: When fixedSize is used, we need to explicitly set the preview so that the image appears at the correct aspect ratio in the context menu. This happens in the thumbnails in compact mode.

Showing the preview would also theoretically work in all other cases too, but unfortunately a small transition bug occurs if we state the preview explicitly for images in large posts. To avoid this, I don't state the preview, and SwiftUI uses the view itself as the preview. This has no visual difference to setting a preview explicitly, but fixes the aforementioned transition bug

boscojwho (approved):
🤌

PR #923, opened 2024-02-10
Mark Read On Scroll

### LINKED ISSUE #645: Mark Read on Scroll
When the "mark read on scroll" setting is enabled, posts would be marked as read automatically when they appear on your screen. Apollo had this

EricBAndrews:
# Checklist
      - closes #645 

# Pull Request Information

## About this Pull Request
Implements mark read on scroll. 

Posts are marked read when the post below them appears; for smaller display sizes, they are marked read when the post some threshold below them appears. This gives a nice approximate "post crossed the screen" heuristic without any complicated geometry reading or timers.

The API calls are batched to avoid spamming the server. This is handled by the new `MarkReadBatcher`, which is available as a `Dependency`. It dispatches posts in batches of 50, or less on conditional dispatch (e.g., whenever the feed is changed, the current batch is flushed so that those pending posts do not appear as unread when navigating back to the original feed).

I have also added a setting to enable/disable mark read on scroll; when the site version is too low to support batching, the feature and toggle are disabled and a message is displayed.


Sjmarf (changes requested):
Looks good! The batcher should be flushed when switching accounts, too.

EricBAndrews:
> Looks good! The batcher should be flushed when switching accounts, too.

Nice catch! Fixed.

Sjmarf (approved):
LGTM 👍

PR #926, opened 2024-02-13
Face ID: Use DispatchQueue.async not .sync.

boscojwho:
Replace biometric logic logic that dispatched callback work using `.sync` with `.async`.


boscojwho:
already done in another PR

PR #932, opened 2024-02-17
Tab navigation custom behaviours

boscojwho:
# Pull Request Information

## About this Pull Request
Allows users to customize tab bar actions behaviour:
- System: Pop to root, then scroll to top (and pop back to sidebar, if available).
- Dismiss: Go back to previous page until root page, then scroll to top (and pop back to sidebar, if available).
- Dismiss after Scroll: Always scroll-to-top before going back to previous page (and pop back to sidebar, if available).

Appreciate any input on behaviour naming and explanation text.

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -414,6 +414,7 @@
 		CD4DBC032A6F803C001A1E61 /* ReplyToPost.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD4DBC022A6F803C001A1E61 /* ReplyToPost.swift */; };
 		CD525F652A4B6D8F00BCA794 /* CommunityLinkView.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD525F642A4B6D8F00BCA794 /* CommunityLinkView.swift */; };
 		CD59E8A52A72C943005757F4 /* MarkAllAsReadRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD59E8A42A72C943005757F4 /* MarkAllAsReadRequest.swift */; };
+		CD5F76BC2B75BE700013A827 /* MarkReadBatcher.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD5F76BB2B75BE700013A827 /* MarkReadBatcher.swift */; };
 		CD6483302A38D31C00EE6CA3 /* UpvoteCounterView.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD64832F2A38D31C00EE6CA3 /* UpvoteCounterView.swift */; };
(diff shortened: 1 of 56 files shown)


Sjmarf:
I think the wording here should be clarified somewhat. 

Would an inline list picker give us more space (the one with the checkmarks)? Maybe it could be titled "Tap tab bar icon to..." and have the following options:

- "Do nothing"
- "Scroll to top"
- "Return to previous page"
- "Return to first page"

boscojwho:
> I think the wording here could be improved somewhat.
> 
> Would an inline list picker give us more space (the one with the checkmarks)? Maybe it could be titled "Tap tab bar icon to..." or "When I tap the tab bar icon..." and have the following options:
> 
> * "Do nothing"
> * "Scroll to top"
> * "Return to previous page"
> * "Return to first page"

Yea, that could work better, will tweak and update

Sjmarf:
Looks good! I think it might look a little better in Title Case, without the trailing ellipses?

EricBAndrews:
We might want to move all the tab bar stuff into its own section of Settings--with this PR we're going to have:
- Account switcher behavior in Profile
- Tab bar appearance in Appearance
- Tab tap behavior in General

Certainly not a blocker to this PR though

Sjmarf:
That could work. 

What about the other categories? There are other cases where we will have a similar split-up grouping in future. An example would be when we add swipe action customisation, which in our current layout would go inside of a new "Gestures" page. That splits them up from the interaction bar customisation, which feels slightly off to me. And what if we then add ellipsis menu customisation too? Where would that go? It would be strange to put that in Appearance.

One solution go the proposed tab-bar settings route for the other categories too, and simply put all of the navigation links that are currently under Appearance under the root view. Then, gestures and ellipsis menu customisation could go in the same menu as appearance settings.

boscojwho:
> Looks good! I think it might look a little better in Title Case, without the trailing ellipses?

Both work, I chose the ellipses without title case to indicate that there's more to the action than the literal name (i.e. nudge users to read the footer text lol)

EricBAndrews:
Re: menu names--these behaviors are super complex to put in a tiny little switcher item. Ellipsis + footer works well enough for me as a pro tem solution, but longer term we might want to look into using animations to describe the behavior the way we do with the quick switcher

boscojwho:
> Re: menu names--these behaviors are super complex to describe in a single list entry. Ellipsis + footer works well enough for me as a pro tem solution, but longer term we might want to look into using animations to describe the behavior the way we do with the quick switcher

Yea, eventually animations, figured I'd get this in without them for now because of time.

EricBAndrews (changes requested):
Nice! Works pretty much perfectly. I found a couple minor defects while testing:
- Scroll to Top does not behave properly in Community view
- Do Nothing still scrolls to top in Feeds view

EricBAndrews:
Closing this as stale

PR #945, opened 2024-02-18
Post Interactions

Sjmarf:
Implements functionality for the upvote, downvote and save buttons. Adds menu functions for those actions, too. Swipe actions don't work yet.


Mlem/Models/Content/Post/Post2Providing.swift
```diff
@@ -40,4 +42,74 @@ extension Post2Providing {
...
+        }
+        switch newVote {
+        case .upvote:
+            upvoteCount += 1
+        case .downvote:
+            downvoteCount += 1
+        default:
+            break
+        }
+        myVote = newVote
```
EricBAndrews, line 65: This seems like logic we're going to want to reuse--maybe put it in a `ScoringOperation` extension:
```
func computeVoteDelta(newVote) -> (Int, Int) {
  if newVote == self { return }
  var upvoteDelta = 0
  var downvoteDelta = 0
  switch self { ... }
  switch newVote { ... }

  return (upvoteDelta, downvoteDelta)
}

```

Mlem/Views/Shared/Components/Components/VoteButtonView.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  UpvoteButtonView.swift
```
EricBAndrews, line 2: *nit*
```suggestion
//  VoteButtonView.swift
```

EricBAndrews (changes requested):
Looks good! A couple minor suggestions. Glad to see those old button views getting some love lol

EricBAndrews (changes requested):
Looks good! Everything seems to work. Something we had very inconsistently implemented in v1 (present in `Comment Model Logic` but never made it to posts) and I think we should use universally in v2 is a little bit of logic to prevent submitting a vote or a save while another one is in-flight to prevent this sort of thing:

https://github.com/mlemgroup/mlem/assets/44140166/521eeeff-f5ac-41fe-9c1a-a05266d1bc35

Mlem/Models/Content/Post/Post2Providing.swift
Sjmarf: Good idea. I've implemented something similar to this 👍

Sjmarf:
That's a good point. 

Instead of preventing future requests, could we just not update the model when we get an un-matching response back? Something like:

```swift
let response = try await source.api.voteOnPost(id: id, score: newVote)
if response.postView.myVote == myVote { // Adding this `if` statement
    update(with: response.postView)
}
```

EricBAndrews:
That would work too--it's less state to track and makes the interface feel marginally more responsive, though it would display inaccurate information in fringe failure cases like this one (assuming we stick with the paradigm of reverting to pre-call state on failure):
- User upvotes
- Client state fakes to upvoted
- User downvotes
- Client state fakes to downvoted
- Upvote call fails
- Post is reverted to neutral
- Downvote call succeeds
- Post is not updated because response does not match; client is now inconsistent with server

Sjmarf:
Hmm, yeah. Maybe preventing other calls from starting would be better. Another option might be to cancel the previous task early if another task starts.

EricBAndrews:
Task cancellation seems good as well--as far as I can tell it avoids inconsistencies. I'd be content with either solution--I think blocking calls with in-flight counterparts might be a little nicer code-side, especially if we create a reusable type to handle tracking that state like below, but I'd be happy to merge either approach.

```
struct CallStatusTracker {
  private var isVoting: Bool = false
  private var isSaving: Bool = false

  func canVote() -> Bool { !isVoting }
  func startedVoting() { isVoting = true }
  func endedVoting() { isVoting = false }

  ...
}
```

Sjmarf:
I've implemented a cancelling system.

EricBAndrews (changes requested):
I'm still able to get some really weird behavior if I click it a lot:

https://github.com/mlemgroup/mlem/assets/44140166/690803c3-9b4a-4d67-9bde-ab811dfbff76

I'm not sure exactly what the sequence is that produces this--I just upvote/downvote a bunch then spam the save with upvote on save enabled.

Sjmarf:
I _think_ I may have improved it slightly. I ran into a strange quirk (see Slack), which this hopefully works around.

There's still some weird state flickering if you spam the buttons hard whilst on bad internet. I don't think I'll be able to get rid of this. I haven't been able to get this to happen unless I spam the buttons quite fast, which is something that the average person won't do. Of course, it's very possible that you may have different results.

Mlem/Models/Content/Post/Post2Providing.swift
```diff
@@ -40,4 +43,117 @@ extension Post2Providing {
...
+        }
+    }
+    
+    func toggleSave() {
+        if let oldTask = post2.tasks.save, !oldTask.isCancelled { oldTask.cancel() }
+        post2.tasks.save = Task(priority: .userInitiated) { await toggleSaveTask() }
+        if myVote != .upvote, UserDefaults.standard.bool(forKey: "upvoteOnSave") {
+            vote(.upvote)
+        }
+    }
```
EricBAndrews, line 98: Upvote still applies on unsave--this should fix it 

```suggestion
    func toggleSave() {
        let wasSaved = isSaved
        if let oldTask = post2.tasks.save, !oldTask.isCancelled { oldTask.cancel() }
        post2.tasks.save = Task(priority: .userInitiated) { await toggleSaveTask() }
        if myVote != .upvote, !wasSaved, UserDefaults.standard.bool(forKey: "upvoteOnSave") {
            vote(.upvote)
        }
    }
```

EricBAndrews (changes requested):

Mlem/Models/Content/Post/Post2Providing.swift
Sjmarf: Oops, I forgot to commit that. Thanks :D

EricBAndrews (approved):
Nice 👍

PR #955, opened 2024-02-26
Resolve posts when switching accounts

Sjmarf:
When switching accounts whilst a post is open, Mlem will attempt to resolve the post on the new account. Whilst the resolution is in-progress, the user is unable to interact with the post and the navigation title is dimmed. If the resolution fails, interactions remain disabled and the navigation title reads "Failed to resolve" in red. 

This is a basic implementation that applies to the Expanded post only and not the feed, though I have plans to bring this to the feed too soon.


EricBAndrews (approved):
LGTM 👍

PR #956, opened 2024-02-28
Eric/mod context menus

### LINKED ISSUE #935: Post Moderation Context Menus
Add mod actions context menu to posts

EricBAndrews:
# Checklist
      - closes #935 

# Pull Request Information

This PR adds community moderation context menus to posts, enabling moderators to:
- Pin/unpin post
- Lock/unlock post
- Remove post
- Ban/unban creator

To make these menus available in non-community feeds, `SiteInformationTracker` has been updated to fetch the full `UserModel` for the current user and maintain a set of moderated communities.

Moderation tools are handled via the new `ModToolSheet` and `ModToolTracker`, which function in the same manner as the composers.

`BanUserView` has also been refactored to extract reusable components.

Screenshot 2024-02-28 at 11 39 37 AM


Sjmarf:
Nice! I noticed that, occasionally, the "Community moderation" submenu will close by itself shortly after I open it. Possibly due to a view update?

I think the wording of some of the options could be tweaked a little. I think "Community Moderation" could be just "Moderation". And "Pin post" could be just "Pin", etc. 

At some point we should show a small lock icon on posts that are locked, though I don't mind if that isn't in this PR.

Mlem/Models/Trackers/SiteInformationTracker.swift
```diff
@@ -39,8 +43,14 @@ class SiteInformationTracker: ObservableObject {
...
                 if let userInfo = response.myUser {
-                    myUser = UserModel(from: userInfo.localUserView.person)
+                    // need to fetch full user info to get moderated communities, required to display mod functions from non-community feed
+                    myUser = try await personRepository.loadUser(for: userInfo.localUserView.localUser.personId)
                     myUser?.isAdmin = response.admins.contains { $0.person.id == myUser?.userId }
+                    if let communities = myUser?.moderatedCommunities {
+                        moderatedCommunities = Set(communities.map(\.communityId))
+                    } else {
+                        moderatedCommunities = .init(minimumCapacity: 1)
+                    }
```
Sjmarf, line 53: You can get the list of moderated communities from MyUserInfo without fetching the ApiPersonResponse separately

Mlem/Views/Shared/Posts/Feed Post.swift
```diff
@@ -91,10 +91,21 @@ struct FeedPost: View {
...
+            community: isMod ? postModel.community : nil,
+            modToolTracker: isMod ? modToolTracker : nil
+        )
+    }
 
     var body: some View {
-        // this allows post deletion to not require tracker updates
-        if postModel.post.deleted {
+        // this allows post deletion/removal to not require tracker updates
+        if postModel.post.deleted || postModel.post.removed {
```
Sjmarf, line 108: Moderators are unable to restore posts because of this check. Maybe we should only hide removed posts for non-moderators, and show a "removed" label for moderators?

Mlem/Views/Tabs/Feeds/Feed Types/CommunityFeedView.swift
```diff
@@ -155,7 +156,7 @@ struct CommunityFeedView: View {
                 switch selectedTab {
                 case .posts: posts()
                 case .about: about()
-                case .moderators: moderators()
+                case .moderators: ModeratorListView(community: communityModel) // moderators()
```
Sjmarf, line 159:
```suggestion
                case .moderators: ModeratorListView(community: communityModel)
```

Sjmarf (changes requested):

Mlem/Views/Shared/Moderation/ModToolSheet.swift
```diff
@@ -0,0 +1,38 @@
...
+import Foundation
+import SwiftUI
+
+struct ModToolSheet: View {
+    let tool: ModTool
+    
+    var body: some View {
+        switch tool {
+        case let .instanceBan(user, shouldBan):
+            BanUserView(user: user, community: nil, shouldBan: shouldBan, postTracker: nil) // TODO: add
```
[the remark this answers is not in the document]
EricBAndrews: It is used, but it's currently only accessible via the user profile page navigation menu--I plan to add admin actions to the context menu as well, at which point it will be much more accessible.

Mlem/Views/Shared/Moderation/ModToolSheet.swift
```diff
@@ -0,0 +1,38 @@
...
+    func moderators(for community: CommunityModel) -> some View {
+        VStack(spacing: 0) {
+            Divider()
+            
+            ModeratorListView(community: community, navigationEnabled: false)
+            
+            Spacer()
+        }
+        .navigationTitle("Moderators")
+    }
```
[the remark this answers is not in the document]
EricBAndrews: Yep, artifact from an earlier idea

Mlem/Views/Shared/Moderation/Tools/BanUserView.swift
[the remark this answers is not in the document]
EricBAndrews: Nice catch!

Mlem/Views/Shared/Posts/Feed Post.swift
EricBAndrews: Implemented 👍 

The API still doesn't return removed posts, though, so this is only accessible until the feed refreshes or via the modlog when that is implemented.

EricBAndrews:
Haven't been able to repro the submenu, but I'll keep an eye out for it.

I've tweaked the wording--"post" has been removed from the context menus. I kept "ban user" (reworded to "ban author") on the grounds that the author is not the entity the context menu is attached to.

I've also implemented the lock indicator--for 2.0 we might want to revisit how we handle tags, because it's getting *crowded* in there (in the extreme scenario)

Screenshot 2024-03-02 at 8 07 48 PM

Sjmarf:
> Re: menu naming, I plan to add an "Instance moderation" submenu containing admin actions; this lets us keep phrasing like "ban author" instead of needing "ban author from community/instance," and makes it explicit whether the user is acting as moderator or as admin--I'm not strongly attached to this UX, though, so if you've got ideas I'd love to hear them.

One option would be "Moderation" and "Administration". 

What would be going in the administration menu? Obviously banning users from the instance; would there be anything else? If it's just banning users, it doesn't make much sense to have a separate disclosure group. In that case we could just have a single "Moderation" group and put the community/instance picker in the ban sheet itself.

If we only need one menu, we could even put all of the menu functions at the top level without having a disclosure group. The advantage of this is that it requires less taps to perform an action. The disadvantage is that the menu might get too long if we add many more actions here.



> I've also implemented the lock indicator

Nice. Maybe orange or yellow would be a better color? Green doesn't look quite right to me.

> for 2.0 we might want to revisit how we handle tags, because it's getting crowded in there (in the extreme scenario)

Yeah. Perhaps (for this view mode) they could go vertically here? Aside from the NSFW tag, which would stay where it is now.

309512487-c72c2ca1-92bd-4c60-83cf-f4e23483d825

Mlem/Models/Content/Post/PostModel+MenuFunctions.swift
```diff
@@ -136,40 +157,107 @@ extension PostModel {
...
+        ) {
+            modToolTracker.removePost(self, shouldRemove: !self.post.removed)
+        })
+
+        if creator.userId != siteInformation.userId {
+            functions.append(MenuFunction.toggleableMenuFunction(
+                toggle: creatorBannedFromCommunity,
+                trueText: "Unban author",
+                trueImageName: Icons.communityUnban,
+                falseText: "Ban author",
```
Sjmarf, line 224: We should use Title Case here. Also, we should standardise our wording of "User" vs "Author" - here, we use "Ban Author" but we use "Block User" in other places.

```suggestion
                trueText: "Unban Author",
                trueImageName: Icons.communityUnban,
                falseText: "Ban Author",
```

Mlem/Models/Trackers/SiteInformationTracker.swift
```diff
@@ -40,7 +44,12 @@ class SiteInformationTracker: ObservableObject {
                 allLanguages = response.allLanguages
                 if let userInfo = response.myUser {
                     myUser = UserModel(from: userInfo.localUserView.person)
-                    myUser?.isAdmin = response.admins.contains { $0.person.id == myUser?.userId }
```
Sjmarf, line 43: I don't think this needs to be removed?

Sjmarf (changes requested):

Mlem/Models/Content/Post/PostModel+MenuFunctions.swift
EricBAndrews: Changed to "User"--I think "Author" is marginally more precise, but that's OOS for this PR

Mlem/Models/Trackers/SiteInformationTracker.swift
EricBAndrews: Whoops, nice catch

EricBAndrews:
It looks like all the admin functions are analogous enough to mod functions for the popup approach to work--as far as I can tell the options are:

Moderator | Admin | Proposed UX
--- | --- | ---
Feature community | Feature local | Popup
Lock | Lock | Idempotent, no popup
Remove post | Purge post | Toggle in sheet
Ban user from community | Ban user from instance, Purge user from instance | Selector in sheet
Appoint user as mod | Appoint user as admin | Popup under "promote"? Separate actions? This is the weird one for sure

For now I've kept it simple and just pulled the mod functions out of the submenu and added to a section beneath the normal operations--the menu is *long* now, but that should be fixed by #960.

Sjmarf:
> It looks like all the admin functions are analogous enough to mod functions for the popup approach to work

Sounds good 

> Appoint User: Popup under "promote"? Separate actions? This is the weird one for sure

I don't think we should put this in the post/comment context menu at all. How often is an admin or moderator going to be appointing other users? I think it's better to put these options elsewhere.

At the bottom of the moderator list in `CommunityView` and the admin list in `InstanceView`, we could put a "+ Add" button that shows a user-search popup. Additionally, we could put "Appoint as admin" and "Appoint as moderator" actions in the ellipsis menu on user profiles. "Appoint as moderator" could show a sheet that lets you choose from a list of communities you moderate, or alternatively it could just appoint to the community you navigated to the profile from.

> For now I've kept it simple and just pulled the mod functions out of the submenu and added to a section beneath the normal operations--the menu is long now, but that should be fixed by https://github.com/mlemgroup/mlem/pull/960.

I'll replicate this in #960 👍

Sjmarf (approved):
Nice 👍

PR #960, opened 2024-03-02
Remove duplicate context menu functions

Sjmarf:
> [!note]
> Built ontop of #956 - drafting until that is in.

- Hid menu functions for posts/comments when the action is available via the interaction bar, unless "Show Duplicate Context Menu Actions" is enabled in accessibility settings.
- Collapsed the "Block User" and "Block Community" functions into one. 
- Hid the "Report" button if you are a moderator, and the "Remove" button for your own posts.
- Refactored a lot of the MenuFunction logic.


EricBAndrews (approved):
LGTM 👍

PR #961, opened 2024-03-03
Select Text

### LINKED ISSUE #256: Share as image/select text

Sjmarf:
Closes #256

This adds the [SwiftUIIntrospect](https://github.com/siteline/swiftui-introspect) library as a dependency. I think this is worth having - I also plan to use it in future for detecting the cursor location within a TextField for inline image uploading.

<hr>


Mlem/Views/Shared/SelectTextView.swift
```diff
@@ -0,0 +1,57 @@
...
+                .background(Capsule().fill(Color.accentColor))
+                Button {
+                    dismiss()
+                } label: {
+                    Image(systemName: "xmark.circle.fill")
+                        .resizable()
+                        .aspectRatio(contentMode: .fit)
+                        .frame(height: 30)
+                        .symbolRenderingMode(.palette)
+                        .foregroundStyle(.secondary, .secondary.opacity(0.2))
```
EricBAndrews, line 44: This introduces an inconsistency with our image viewer sheet, which feels similar enough that we should keep them aligned. The close/cancel/dismiss action is also almost always in the top left--it's a bit weird to see it over on the right.

EricBAndrews (changes requested):
Looks good! Functionality all works.

Minor note on button placement/styling.

Mlem/Views/Shared/SelectTextView.swift
Sjmarf: I put it in the top right for consistency with what iOS uses (share sheet, maps, etc). As far as I can tell they use this design everywhere Apple isn't entirely consistent with "cancel" buttons, though - they appear to be in the top-right when possible, or in the top-left if a "done" action is placed in the top-right.

If we want to remain consistent, maybe the image close button would be better on the right too?

EricBAndrews: Moving the image viewer close button works for me--I agree that the apple default seems to prefer the right unless it's taken by a send.

Sjmarf: Done 👍

EricBAndrews (approved):
Nice 👍

PR #966, opened 2024-03-08
Appoint Mod from Profile

EricBAndrews:
# Checklist

# Pull Request Information

This PR adds the ability to appoint users as moderators from their profile page via the navigation bar menu. It also reworks the add mod UI element to be reusable across all contexts, replacing the simple search with a form that allows modification of the community or user depending on context:

https://github.com/mlemgroup/mlem/assets/44140166/d1ee89ff-84db-4e84-bd86-f5dfe155b39b

Admins can use this feature to appoint a user as a moderator to any community:

https://github.com/mlemgroup/mlem/assets/44140166/730894d4-e828-421a-888e-c61a27fffa2d

This PR also refactors `CommunityResultView` into `CommunityListRow`, analagous to `UserListRow`.


Sjmarf:
Nice! I like the form approach. It allows for potentially selecting multiple communities at some point in future, too. That would be useful if someone is moving to another instance, for example, and wants to appoint their new account as a moderator to all of the communities they moderate. 

I wonder if it's worth adding a context menu action in the community list in the profile to remove the user as a moderator from that account? 

IMG_2125D7534CA7-1

Mlem/Views/Tabs/Profile/UserView+Logic.swift
```diff
@@ -10,6 +10,22 @@ import SwiftUI
...
+    var menuFunctions: [MenuFunction] {
+        var ret = user.menuFunctions({ user = $0 }, modToolTracker: modToolTracker)
+        
+        // add moderator needs to be defined as a menu function out here instead of within the user model so that it can take the $user binding
+        // TODO: 2.0 move add moderator menu function into Person
+        if !isOwnProfile || siteInformation.myUser?.isAdmin ?? false {
+            ret.append(.standardMenuFunction(
+                text: "Appoint as Moderator",
+                imageName: Icons.moderation
+            ) {
```
Sjmarf, line 22: We should hide this menu function if the user does not moderate any communities and is not an admin

Sjmarf (changes requested):

Mlem/Views/Tabs/Profile/UserView+Logic.swift
EricBAndrews: Good call, implemented 👍 

I've also added a waiting overlay and cancel buttons to the search view

EricBAndrews:
That's a good idea--in the future I want to add a lot more paths to access mod functions like that, but I'm holding off until 2.0 because the new middleware should make the currently very awkward task of propagating mod status back to the view pretty trivial.

Sjmarf (approved):
Nice!

PR #967, opened 2024-03-08
Rearrange Menu Functions

Sjmarf:
Moved upvote/downvote/save/reply onto one row. Also moved the ellipsis menu of an expanded post to the toolbar. Pin and Lock are only shown in the expanded post.


EricBAndrews (approved):
LGTM 👍

PR #970, opened 2024-03-10
Inline Image Uploading

### LINKED ISSUE #667: Inline image uploading
When typing the body of a new post, an 'add attachment' button should be available on the keyboard toolbar. Clicking this would upload an image and add the link in GH markdown format to the body of the post, like this:

`Alt text`

Sjmarf:
Closes #667 

Allows uploading inline images in `ResponseEditorView` and `PostComposerView`. Ideally, the inline image upload button would have been on a keyboard toolbar. Unfortunately, this caused issues due to an iOS 17 bug. If that bug is fixed in future I'll move it to the keyboard toolbar.


Mlem/Views/Shared/Composer/ResponseEditorView.swift
```diff
@@ -23,100 +23,145 @@ struct ResponseEditorView: View {
...
+            .overlay {
+                // Loading Indicator
+                if isSubmitting {
+                    ProgressView()
+                        .frame(maxWidth: .infinity, maxHeight: .infinity)
+                        .background(Color.gray.opacity(0.3))
+                        .accessibilityElement(children: .ignore)
+                        .accessibilityLabel("Submitting Response")
+                        .edgesIgnoringSafeArea(.all)
+                        .allowsHitTesting(false)
```
EricBAndrews, line 86: We use this basic construct in a couple different places--maybe we want it in a reusable component?

EricBAndrews (approved):
LGTM, one comment but non-blocking.

Mlem/Views/Shared/Composer/ResponseEditorView.swift
Sjmarf: Done 👍

PR #971, opened 2024-03-10
Purge post/user

Sjmarf:
This PR implements the following functions:
- Purge post
- Purge user
- Ban user from instance using the post context menu, without having to navigate to the profile first

The user can choose whether to ban/unban from the community or instance using a picker at the top of the ban sheet. "Instance" is selected by default.

If a user is banned from a community but not from the instance, or vice versa, both "Ban" and "Unban" buttons will be shown in the context menu.

To purge a post, I've added a toggle in the post removal sheet:

And to purge a user, I've added this picker to the ban sheet:


Mlem/Views/Shared/Moderation/Tools/BanUserView+Logic.swift
```diff
@@ -0,0 +1,87 @@
...
+import SwiftUI
+
+extension BanUserView {
+    func confirm() {
+        if !banFromInstance, let communityContext {
+            communityBan(from: communityContext)
+        } else {
+            instanceBan()
+        }
+    }
```
EricBAndrews, line 17: *nit*

Technically this could lead to an instance ban being sent if the state gets mucked up such that `banFromInstance` is false but `communityContext` is nil--safer to do something like this:
```suggestion
        if banFromInstance {
            instanceBan()
        } else if let communityContext {
            communityBan(from: communityContext)
        } else {
            assertionFailure("banFromInstance false but communityContext nil!")
        }
    }
```

EricBAndrews (changes requested):
Looks good, one minor nit 👍

EricBAndrews (approved):
Nice 👍

PR #975, opened 2024-03-15
Modlog

### LINKED ISSUE #946: Add Access to Modlog
## Introduction
It is my understanding that accessing the modlog for an instance is done by going to https://instance/modlog. This is currently not a feature of mlem. 

## Requirements
Lemmy exposes an API endpoint to access the modlog.

## Discussion
Unofficial documentation can be found at https://lemmy.readme.io/reference/get_modlog

## What I Need
Similar to the web UI, it should show each entry to the modlog.

EricBAndrews:
# Checklist
      - closes #946 

# Pull Request Information

## About this Pull Request

This PR adds the ability to view the modlog for instances and communities. The API returns fifteen separate types, one for each modlog action; for sanity's sake, I consolidate those all into a single `ModlogEntry`. To address the fact that the API doesn't paginate the desired way, the modlog tracker is handled as a 15-way multi-tracker.

The modlog can be accessed from the "Moderation" tab of a Community and the "Administration" tab of an Instance.

Contextual information (moderator, affected user, post, etc.) is navigable using either the ellipsis menu or the context menu. Because the modlog is fetched from the instance it is requested from, not necessarily the user's home instance, links to the relevant items in the modlog are handled as `URL`s that get resolved using the `resolveObject` endpoint; if the object fails to resolve, the user is taken to the web page of the instance providing the object. This has necessitated the addition of an `openUrlMenuFunction`.

### Community modlog
https://github.com/mlemgroup/mlem/assets/44140166/47fdc62c-63ec-4372-8a79-307e8a65bb86

### Instance modlog
https://github.com/mlemgroup/mlem/assets/44140166/a12c2390-23c3-446b-b200-cab8cc4dad31

### Modlog filtering

https://github.com/mlemgroup/mlem/assets/44140166/4a25109e-8082-4afa-8583-a037eb54d464


Sjmarf:
Nice! I'll do a proper review this weekend. Some immediate thoughts:

- Should the filter button go in the toolbar? The downside of this is it doesn't show the full text, and icons could be ambiguous.
- Should we group the actions in the filter menu somehow? It's a little long. Maybe disclosure groups for "Post", "Comment", "Community" etc? Or alternatively disclosure groups for "Ban", "Remove", "Purge", etc.
- I think we should show still show moderator vs administrator somehow when viewing a modlog with moderator names enabled. Maybe a mod/admin icon underneath the action icon on the left, maybe in a neutral color?
- I'm not a huge fan of using the circular community icon in the context menus - it looks too detailed to me. I think using `house` would work better here.


- Non-moderators get an error when they click "View post". I think we should either disable or remove that action in cases where the user wouldn't be able to view the post.
- I think we should add a content warning here, similar to what lemmy-ui has. I think we should show a full-screen content warning that obscures the content, with an option to disable it in settings. The warning wouldn't be shown for communities you moderate, or for your own instance if you are an administrator.

EricBAndrews:
> - Should the filter button go in the toolbar? The downside of this is it doesn't show the full text, and icons could be ambiguous.
> - Should we group the actions in the filter menu somehow? It's a little long. Maybe disclosure groups for "Post", "Comment", "Community" etc? Or alternatively disclosure groups for "Ban", "Remove", "Purge", etc.

I tried it in the toolbar but it's pretty awkward and not particularly discoverable. Post-2.0 I plan to add searching and filtering tools that will require some real estate at the top of the view anyway, so the lonely little "Action Type" section will be a little less odd.

> - I think we should show still show moderator vs administrator somehow when viewing a modlog with moderator names enabled. Maybe a mod/admin icon underneath the action icon on the left, perhaps in a neutral color?

Agreed. There are a couple of different approaches here:
1. Dumb and easy: admin-exclusive actions are flagged as admin, everything else is flagged as mod. This wouldn't even require us to have modlog names enabled, and is in line with how the desktop site handles it
2. (1), but don't bother displaying a separate icon--simply use a different color set for admin actions (pink/mint?)
3. Sophisticated: non-admin-exclusive actions are flagged as admin if the user is a moderator on that instance, moderator otherwise. This introduces some further difficulties:
   - Unclear how to handle modlog with hidden names. Can either fallback on (1), which would be inconsistent, or not display the actor class at all.
   - Unclear how to handle the case where an admin who is also a moderator on a given community performs a non-admin action on an item in that community

> - I'm not a huge fan of using the circular community icon in the context menus - it looks too detailed to me. I think using house would work better here.

`house` is already used to refer to Local, but `building.2` should do the trick

> - Non-moderators get an error when they click "View post". I think we should either disable or remove that action in cases where the user wouldn't be able to view the post.

I had that thought as well--currently working on the implementation.

> - We don't need to show "View community" for community modlogs.

With the aforementioned post-2.0 searching/filtering, I plan for the modlog page to be context-agnostic and for the context to simply pre-populate a filter (either instance or community) on initial navigation. The user will then be able to add or remove filters at will. A nice way to handle this would be to display the filters as a stack of labels at the top of the view, which can serve as navigation links to the filtered entity. I'll try to rig this up in a static way for 1.3--should be pretty easy.

> - I think we should add a content warning to the page, similar to what lemmy-ui has. I think we should show a full-screen content warning that obscures the content, with an option to disable it in settings.

Good point, I'll look into it

Sjmarf:
> I tried it in the toolbar but it's pretty awkward and not particularly discoverable. Post-2.0 I plan to add searching and filtering tools that will require some real estate at the top of the view anyway, so the lonely little "Action Type" section will be a little less odd.

Sure, sounds good 👍 

> Agreed. There are a couple of different approaches here:
> 1. Dumb and easy: admin-exclusive actions are flagged as admin, everything else is flagged as mod. This wouldn't > even require us to have modlog names enabled, and is in line with how the desktop site handles it
> 2. (1), but don't bother displaying a separate icon--simply use a different color set for admin actions (pink/mint?)
> 3. Sophisticated: non-admin-exclusive actions are flagged as admin if the user is a moderator on that instance, moderator otherwise. This introduces some further difficulties:
Unclear how to handle modlog with hidden names. Can either fallback on (1), which would be inconsistent, or not display the actor class at all.
Unclear how to handle the case where an admin who is also a moderator on a given community performs a non-admin action on an item in that community

Hmm, that's a little tricky. I'm not sure what the best approach here would be. 3) with a fallback to 1) makes the most sense to me, though it's a little inconsistent as you mention...

> house is already used to refer to Local, but building.2 should do the trick

Nice, that works 👍 

> With the aforementioned post-2.0 searching/filtering, I plan for the modlog page to be context-agnostic and for the context to simply pre-populate a filter (either instance or community) on initial navigation. The user will then be able to add or remove filters at will. A nice way to handle this would be to display the filters as a stack of labels at the top of the view, which can serve as navigation links to the filtered entity. I'll try to rig this up in a static way for 1.3--should be pretty easy.

Cool 👍

Mlem/Models/Trackers/Feeds/Modlog/ModlogChildTracker.swift
```diff
@@ -0,0 +1,65 @@
...
+        case let .community(community):
+            self.instance = nil
+            self.communityId = community.communityId
+        }
+        
+        super.init(internetSpeed: internetSpeed, sortType: sortType)
+    }
+    
+    override func fetchPage(page: Int) async throws -> FetchResponse<ModlogEntry> {
+        let items = try await apiClient.getModlog(
```
Sjmarf, line 55: This means making thirteen separate requests for each modlog page that we load - one for each child tracker. Is there a better way to do this? Maybe page loading could be delegated to the parent tracker somehow, so only one request needs to be made per page?

Mlem/Views/Shared/Components/Modlog/ModlogView.swift
```diff
@@ -0,0 +1,375 @@
...
+    @StateObject var communityBansTracker: ModlogChildTracker
+    @StateObject var instanceBansTracker: ModlogChildTracker
+    @StateObject var moderatorAddsTracker: ModlogChildTracker
+    @StateObject var communityTransfersTracker: ModlogChildTracker
+    @StateObject var administratorAddsTracker: ModlogChildTracker
+    @StateObject var personPurgesTracker: ModlogChildTracker
+    @StateObject var communityPurgesTracker: ModlogChildTracker
+    @StateObject var postPurgesTracker: ModlogChildTracker
+    @StateObject var commentPurgesTracker: ModlogChildTracker
+    @StateObject var communityHidesTracker: ModlogChildTracker
```
Sjmarf, line 79: It might be better to put these in a dictionary mapping `ModLogAction` to `ModlogChildTracker`? That means we can initialise these trackers iteratively, and replace the switch statement on line 248 with a simple dictionary access

Sjmarf (changes requested):

Mlem/Models/Trackers/Feeds/Modlog/ModlogChildTracker.swift
EricBAndrews: The trouble arises from the fact that certain actions are far more frequent--we need to load maybe 20 pages of post removals for every page of community transfers, so loading full API pages from the parent tracker would require loading lots and lots of unused data. This is one of the core problems that child trackers were built to solve--they only load the next *n* items of a particular type when that type runs out, rather than loading everything from every type every time. 

The initial preheat, however, is definitely way more expensive than it needs to be. I'm looking into a way to use a single call to preheat the children and then switch over to ad hoc loading.

Mlem/Views/Shared/Components/Modlog/ModlogView.swift
Sjmarf: On second thought, this probably won't work without `@Observable`. Probably doable in 2.0 though

Mlem/Models/Trackers/Feeds/Modlog/ModlogChildTracker.swift
Sjmarf: If the child trackers request the nth page from the parent instead of directly requesting it, the parent could maintain a cache of every page loaded so far and only fetch the next page if it isn't yet cached.

EricBAndrews: Implemented ad-hoc with first page optimizations per Slack

EricBAndrews:
I have implemented the content warning and more efficient loading. Some outstanding items:

- Show whether action was done by moderator or admin: per Slack, leaving out of this PR. It turns out this is a fiendishly complex thing to do thanks to federation. We should revisit this for 2.0 once we have better multi-instance handling.
- Disclosure groups for modlog actions: I added dividers to help orient the user, but refrained from using full disclosure groups. It's a picker, not a context menu, so rendering a long list is not such a sin--in my opinion it is better to render the longer list and keep everything one tap away than collapse into disclosure groups. We could perhaps follow the example we set in the feed sort picker and collapse some of the super low priority items into disclosure groups, but it's really hard to say what sorts of actions a user might want to see more frequently.

Sjmarf (approved):
LGTM!

PR #977, opened 2024-03-16
Remove/Purge Community

### LINKED ISSUE #973: Community Banning
Allow admins to ban/purge communities

Sjmarf:
Closes #973. 

- You can remove a community from the `CommunityFeedView` context menu or from the `CommunityListRow` context menu.
- Added the divider between community and feed options in the `CommunityFeedView` ellipsis menu back in (we had this before 1.2, but it got removed).
- "Blur NSFW" will no longer be shown in the `CommunityFeedView` ellipsis menu if you've disabled NSFW content entirely.

Screenshot 2024-03-16 at 19 32 41

This doesn't state-fake at all (CommunityModel is still a struct, making it difficult to handle in all cases). Ideally, we would:
- State-fake community removal in both communities and in search results
- Remove the result from search results if you purge it
- Pop the `CommunityFeedView` from the navigation stack when you purge it

All of these would be much easier to do in 2.0, so I think we should leave them out for now.


EricBAndrews (approved):
Nice! 👍

PR #984, opened 2024-03-20
Ocean App Icon

Sjmarf:
Added a new "Ocean" app icon. Also tweaked the corner radius of the app icons in the selector to better match the corner radius of the icons on the iOS home screen


EricBAndrews (approved):
Nice!

PR #994, opened 2024-03-27
Rearrange Mod Actions

Sjmarf:
I moved a few files in this one, hence the large line count.

## Separate Purge

Added a separate purge button. "Purge" is shown for your own posts, and "Remove" is hidden. This is because, looking at the modlog, it seems that admins occasionally like to purge their own posts. Let me know if you disagree with this

Screenshot 2024-03-27 at 15 58 24

## Pin to Instance

For admins, the "Pin" button shows a popup asking for them to choose between community and instance

Screenshot 2024-03-27 at 20 08 34

## Grouping

Added a new "Moderation" section in Settings, in which the user can choose to instead group moderation actions using a separate menu or a disclosure group.

Screenshot 2024-03-27 at 16 00 43

Using separate menu:

Screenshot 2024-03-27 at 16 01 08

Using disclosure group:

Screenshot 2024-03-27 at 16 01 39

If "separate menu" is selected, the actions are grouped using a disclosure group in the context menu.

## Show All Actions in Feed

There is also a "Show all actions in feed" setting which shows otherwise hidden actions like "Pin" and "Lock" in the feed.
Screenshot 2024-03-27 at 16 03 52

Screenshot 2024-03-27 at 16 04 16

## Other, Minor Changes

- Removed the confirmation for reporting. There's a popup sheet, so I don't think a confirmation is necessary.
- "Unban" and "Restore" are now consistently non-destructive
- Fixed a bug in which the instance pin icon wasn't shown in compact mode


EricBAndrews (approved):
LGTM!

PR #996, opened 2024-03-29
Dev2 tweaks

Sjmarf:
- Simplified `cacheId` logic
- Fixed a bug where accounts created pre-2.0 wouldn't load correctly
- Made `VoteStatusManager` into a generic `StateManager`

I may make some further tweaks to `StateManager` in another PR

PR #997, opened 2024-03-29
State Manager Improvements

Sjmarf:
This PR makes a few changes to `StateManager`, formerly called `VoteStatusManager` (see #996).

##  1) The value associated with the manager is now stored as a property of the manager

The syntax for setting up a manager looks like so:

```swift
final class Post2: Post2Providing {
    // ...
    var votesManager: StateManager<VotesModel>
    var votes: VotesModel { votesManager.wrappedValue }
    
    init(
        votes: VotesModel,
        // ...
    ) {
        self.votesManager = .init(wrappedValue: votes)
        // ...
    }
}
```

I initially tried to do this with a property wrapper, but that lead to complications regarding `@Observable`.

The benefit of this is that we can move some of the logic that we'd otherwise have in the models into the `StateManager`. I've added a `performRequest` method to `StateManager` that holds the logic:

```swift
func performRequest(
    expectedResult: Value,
    operation: @escaping (_ semaphore: UInt) async throws -> Void
) {
    let semaphore = self.beginOperation(expectedResult: expectedResult)
    Task {
        do {
            try await operation(semaphore)
        } catch {
            print("DEBUG [\(semaphore)] failed!")
            self.rollback(semaphore: semaphore)
        }
    }
}
```

It can be used like so:

```swift
func vote(_ newVote: ScoringOperation) {
    votesManager.performRequest(
        expectedResult: self.votes.applyScoringOperation(operation: newVote),
    ) { semaphore in
        try await self.api.voteOnPost(id: self.id, score: newVote, semaphore: semaphore)
    }
}
```

## 2) Semaphore numbers are now unique across all state managers

Instead of incrementing its own `semaphore` count, `StateManager` will now increment a static count and set its  `lastSemaphore` property to the new value. This ensures that no two `StateManager` instances are using the same semaphore count to refer to different requests. I'll explain why this is needed further down. 

## 3) Fixing external updates

Previously, updates made from outside of the `StateManager`'s control (e.g, if a save request comes back whilst a vote operation is in progress) would cause state flickering.

I've fixed this by having the `Post2Providing.update` method accept the semaphore from `ApiClient.voteOnPost` as an argument, which it then passes to an `updateReceivedValue` method of `StateManager` along with the updated value. The `StateManager` will only update its `wrappedValue` if the semaphore matches the last semaphore it sent, or if it has no requests ongoing. Otherwise, it updates the clean state.

```swift
func update(with post: ApiPostView, semaphore: UInt? = nil) {
    votesManager.updateWithReceivedValue(
        .init(from: post.counts, myVote: ScoringOperation.guaranteedInit(from: post.myVote)),
        semaphore: semaphore
    )
    isReadManager.updateWithReceivedValue(post.read, semaphore: semaphore)
}
```

The above implementation is why unique semaphores are necessary - otherwise, the semaphore from one request may accidentally match the semaphore from another request. Of course, an alternative would be to pass separate `voteSemaphore` and `isReadSemaphore` values, but that would get too bulky.

## 4) Grouping updates

I've also added some neater syntax for updating multiple state managers at once. For example, voting on a post also needs to update the mark-read state. Here's the syntax for this on the model's side:

```swift
func vote(_ newVote: ScoringOperation) {
    groupStateRequest(
        votesManager.ticket(self.votes.applyScoringOperation(operation: newVote)),
        isReadManager.ticket(true)
    ) { semaphore in
        try await self.api.voteOnPost(id: self.id, score: newVote, semaphore: semaphore)
    }
}
```

Inside of the `groupStateRequest` code, `beginOperation` is called on each manager, with **the same** semaphore used for each manager. When this semaphore gets to the `update` method shown above, all of the managers that are part of the request will update their values correctly. If the request fails, all of the managers will be rolled back correctly.

## 5) Changes to the test frontend

Added a downvote button and dimmed the title if the post is read for testing purposes. Read status is properly state-faked when you vote.

Screenshot 2024-03-29 at 18 07 14

------------

I _think_ this makes the state manager foolproof now. Let me know if you see any potential holes in this design that I may have overlooked.

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -490,7 +490,7 @@
 		CDA1E8422B938E6E007953EF /* Instance+CacheExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDA1E8412B938E6E007953EF /* Instance+CacheExtensions.swift */; };
 		CDA1E8452B938F8C007953EF /* Person+CacheExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDA1E8442B938F8C007953EF /* Person+CacheExtensions.swift */; };
 		CDA1E8472B9390FD007953EF /* Post+CacheExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDA1E8462B9390FD007953EF /* Post+CacheExtensions.swift */; };
-		CDA1E8542B952C3D007953EF /* VoteStatusManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDA1E8532B952C3D007953EF /* VoteStatusManager.swift */; };
+		CDA1E8542B952C3D007953EF /* StateManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDA1E8532B952C3D007953EF /* StateManager.swift */; };
 		CDA1E8562B9531E7007953EF /* ApiPostAggregates+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDA1E8552B9531E7007953EF /* ApiPostAggregates+Extensions.swift */; };
 		CDB7AFC22B82FAE60013E0F0 /* GetPostsRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDB7AFC12B82FAE60013E0F0 /* GetPostsRequest.swift */; };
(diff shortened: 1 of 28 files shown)


EricBAndrews (approved):
Looks good to me! As far as I can tell the logic all checks out, and it performs well on my highly sophisticated stress testing benchmarks (clicking the buttons a lot).

PR #999, opened 2024-03-30
Action System

Sjmarf:
## `BasicAction`

`BasicAction` is the equivalent of the v1 `MenuFunction`. It stores a callback for the action as well as label, icon etc. In v1, `MenuFunction` only worked with context menus - in 2.0, we could have `BasicAction` values be accepted by the interaction bar, swipe action _and_ menu function views. This allows us to avoid duplicate code, and makes it easier to add customisation options in future.

## `ActionGroup`

`BasicAction` conforms to an `Action` protocol. `ActionGroup` also conforms to `Action`. It allows us to nest actions like so:

```swift
var menuActions: ActionGroup {
    ActionGroup(children: [
        ActionGroup(
            children: [upvoteAction, downvoteAction],
            displayMode: .section
        ),
        saveAction
    ])
}
 ```
 
 When rendered, the above definition shows the following menu:
Screenshot 2024-04-10 at 22 07 12

 
There are 4 different `displayMode` cases - `.section`, `.compactSection`, `.disclosure` and `.popup`. How these modes behave will depend on whether the action is shown in the interaction bar, in a context menu, or as a swipe action. 

For a context menu, they behave as follows:
- `.section` creates an inline group of actions, with a divider above and below. This replaces the `.divider` menu function we had in v1.
- `.compactSection` uses a `ControlGroup` to place it's children side-by-side.
- `.disclosure` opens a disclosure group containing the children. When using this mode, you can pass `label` and `menuIcon` parameters to the `ActionGroup` constructor to change the `DisclosureGroup` label.
- `.popup` opens a confirmation dialogue with each child as an option in the popup.

For an interaction bar button, `.section` and `.disclosure` will act in the same way - show a `Menu` when the interaction button is clicked, like the ellipsis menu on a post/comment. `.compactSection` will do the same, but display the icons side-by-side. `.popup` will open a popup.

For swipe actions, all display modes will act as `.popup`.

## `MenuButton`

I've re-implemented `MenuButton` from v1 in this PR. `ActionGroup` with `.popup` doesn't work yet, but the other types do.

## `ActionType` and `Actionable`

`ActionType` is an enum storing all of the action types. We'll use this to store interaction bar configurations, swipe action configurations, etc.

```swift
enum ActionType: String {
    case upvote, downvote, save
}

```

Some models are able to generate the `Action` for a given `ActionType` in an `action(ofType: _)` method. These models conform to an `Actionable` protocol.

I've split `InteractableContent` into `Interactable1Providing` and `Interactable2Providing`, which are conformed to by `Post1` and `Post2` respectively. `Interaction2Providing` conforms to `Interaction1Providing`, which conforms to `Actionable`. `Interaction1Providing` always returns disabled actions when `action(ofType: _)` is called, and `Interactable2Providing` returns actions that are enabled. This allows us to display the interaction bar easily even if we only have a `Post1`.

The Models can call these methods to generate the action and can modify the action before returning it if needed. 

## What about ModToolTracker etc?

This system provides no way of passing a `ModToolTracker` or similar to the callback of an action. Instead, we can define our `ModToolTracker` instance as a static property of itself, and access that everywhere. I haven't implemented it in this PR, but I've tested that it works:

```swift
@Observable
class ModToolTracker {
    var openTool: ModTool?
    
    static let shared: ModToolTracker = .init()
}
```

## Changes to the test interface

I've added a "Save" button, and a context menu to demo the new `MenuButton`. To test upvote-on-save, I've put the setting for it in the profile tab for now.


EricBAndrews (approved):
LGTM 👍

PR #1005, opened 2024-04-11
UIKit Tab Bar

### LINKED ISSUE #1004: [2.0] UiKit TabView solution
Using the system TabView allows us to take advantage of the native accessibility features of the tab bar, and should also be more performant. However, SwiftUI's `TabView` doesn't support additional gestures that we might want, such as the long-press or double-tap gestures that we use currently. Ideally, we would write a UIKit solution that adds this functionality.

Sjmarf:
Switched from using SwiftUI for the tab bar to using UIKit, with implementation of the long-press gesture and swipe-up gesture for accessing the account switcher. I've also added some tab re-selection logic, but it isn't hooked up to anything yet.

Closes #1004

## Rationale

Using UIKit over SwiftUI has several advantages:

- We can take advantage of the native UIKit behavior for accessibility, different rendering styles in different contexts, etc.
- It's more performant. The UIKit tab bar doesn't render every tab at once like our `ZStack`-based SwiftUI solution. It also eliminates the tiny delay that was previously present when switching tabs (which was especially noticeable in low power mode).
- We don't have to mess with extra view-side modifiers (`.fancyTabScrollCompatible`).
- The native tab bar hides the background of the tab bar when at the bottom of the page, which looks nicer imo.

## What to do with the inbox badge?

Using UIKit does have one disadvantage in that we cannot render two badges on a tab icon at once, like we currently do for mod mail. We'll unfortunately have to come up with a different solution in 2.0, if we want to use UIKit for the tab bar. UIKit allows us to change the text on the badge (which can be anything, not just integers) and the color of the badge. The only option I see is to put the two notification numbers in one badge. 

Potentially we could have an option to add a completely separate Mod Mail tab, in which case this wouldn't be an issue. The Mod Mail tab could replace the Settings tab, which we could hide as a toolbar item in the profile tab.

## Effects of using a UIKit tab bar

It should be noted that, because the UIKit tab view doesn't render all of the pages at once, some logic will not work as it did before. For example, `.onAppear` will be called whenever you navigate to the tab, and `.task(id: )` will only notice that a change in id has occurred when the tab is visited.


EricBAndrews (changes requested):
Looks good! This should be much less of a hassle to work with.

There's an interesting interaction between the sheet and the tab bar--opening the sheet grays the selected tab bar item until the sheet is fully dismissed:

https://github.com/mlemgroup/mlem/assets/44140166/28ce72e7-408d-4a24-b4c9-9d680b02debe

EricBAndrews (approved):
LGTM 🚀

Issue #61, opened 2023-06-25
Multi-account feed


ShadowJonathan:
Taking a look at the current defeds across Lemmy, I think that it's worth considering to look at "multi-account feeds", something that is able to stitch together feeds from multiple accounts, to allow someone to not have to "strategically position" their account across multiple servers, but just sign up for them, sub to communities, and be able to view and interact with them in some way or another.

Something to consider;
- We should make it more clear, from this feed view, with which account the user is currently interacting with with the post, when they'd open it.
- We have to have a conflict resolution algorithm when a user tries to add subs to the same community from different accounts, to resolve "which account" is going to interact with that community, from that feed.

EricBAndrews:
This should be pretty easy to set up with 2.0, at least on the rendering side, using the better instance handling + generic multi-trackers. We definitely need the account selector within the comment editor; handling other interactions shouldn't be too complicated--one idea would be to allow users to set a 'priority list' when creating a multi-instance feed and have interactions be performed from the highest-priority account with access to the post. Actions such as blocking a post or user should presumably be executed from all accounts.

Somewhat more complex is handling saved state, though we could perhaps handle that with a matching "saved from all accounts" feed.

Read state is also a little tricky, but that can be handled by just submitting it on all accounts--we won't have to worry about spamming the API, since it's *n* calls client-side but only 1 instance-side.

Sjmarf:
This feature would certainly be highly valuable to a lot of users. I don't think it'll be too confusing, since it would be an opt-in feature and not enabled by default.

What will the UX look like for setting up a multi-feed?

Maybe we could let a user add "multi-accounts" in the Accounts setup page. They could give a name to each multi-account, and choose which of their accounts are included. The user would also be able to add guest accounts to the multi-account in this system. The user could switch in and out of the multi-account easily via the quick switcher.

Encapsulating the "multi-feed" idea in the form of a distinct "account" as described above, rather than adding an "all accounts" feed or something similar allows us to utilise multi-accounts outside of just the All feed. Some examples of where this would be useful:

- The search tab could query all instances that compose the multi-account, and combine their results. This allows the user to quickly access a defederated community without having to switch accounts.
- The "View post" button in the modlog could find which child of the multi-account is authorised to access the post, and display the post using that child account's API client.

How would we display the subscription list in this setup? We could have two settings for the way they are displayed:

- Mix subscriptions in together, displaying the avatar(s) (of some other identifier) of the accounts that are subscribed to each community next to the community
[...]

PR #1007, opened 2024-04-12
`AppState` changes

Sjmarf:
This PR is focused around changes to `AppState`. It also implements basic Guest Mode functionality, and starts building the infrastructure for multi-accounts. 

None of this is concrete; if you have any alternative ideas feel free to offer them

### Guest Mode improvements

I've added an "Enter guest mode" button to the quick switcher (It uses  `lemmy.world`). When you switch accounts, the feed does not reload but all interactions on posts from the old account are disabled. The post models determine this by observing a new `permissions` property of their own `ApiClient`.

The `permissions` property is an enum with three cases:
- `.none` - Don't allow any requests to be made. Will throw an error if one is attempted.
- `.getOnly` - Allow requests to be made, but don't attach the token to any requests.
- `.all` - Allow any request to be made.

When you switch accounts, `AppState` locks the outgoing `ApiClient` by setting `permissions` to `.none`, and unlocks the incoming `ApiClient` by setting `permissions` to `.all`. This ensures that we can't accidentally make calls on inactive clients. 

### AppState

Currently, we store `AppState` as an `@State` property in `ContentView`. This unfortunately causes the `AppState` to be re-initialised whenever the account changes. I encountered an issue stemming from this behaviour when implementing the permissions system described above. I couldn't find any good workarounds for that problem, so I opted to try a different `AppState` storage solution.

Instead of storing it as an environment variable, I've stored the `AppState` instance as a static property of `AppState`. In addition to avoiding the re-initialisation issue, this allows us to directly access the `AppState` from outside of a view, or inside of a view's initialiser, if we want to. I think we should follow this pattern of using static singletons rather than environment variables for most of our state objects, since it allows us to be more flexible with where they're accessed. I mentioned this briefly in the Action PR, too.
 
The downside of doing it this way is that we have to store the `ApiClient` optionally within `AppState`, because the `AppState` will exist in memory during onboarding. My workaround to this is to return a mock `ApiClient` client with `.none` permissions in the case that no `ApiClient` is present. Any calls made to this client will throw an error due to its permissions.

Let me know what you think of this decision.

### Multi-Account Prep

I left some thoughts on how this could be implemented in #61. Whilst we may not necessarily be prioritising fully-functional multi-accounts in 2.0 (unless that's something you'd like to target?), I think we should consider the intricacies of it at this stage to avoid having to conduct a major rewrite further down the line.

In my comment on the multi-account issue, I pitched the idea of having full multi-*accounts* as opposed to just multi-feeds. If that's something that you agree with, `AppState` will need to handle multiple accounts.

In this PR, I've stored the account data (`api`, `user`, `instance`) in an `ActiveAccount` class, which is in turn stored inside of an array property of `AppState`. In future, we could add more than one item to this array to support multiple logged-in accounts at once. 

We could add convenience methods to `appState` for searching this array. For example, when the user navigates to a community page using a multi-account, we'd want to connect using the account that moderates the community, if one exists. I've added an `accountThatModerates` method to `AppState` as a proof-of-concept:

```swift
func accountThatModerates(actorId: URL) -> ActiveAccount? {
    return activeAccounts.first(where: { account in
        account.user?.moderatedCommunities.contains { $0.actorId == actorId } ?? false
    }) ?? .mock
}

```

Which could be used like so:

```swift
let feedApi = appState.accountThatModerates(communityId).api ?? appState.firstApi
```

### `APIClient` convenience properties

I've added `myInstance` and `myUser` as weak properties of `ApiClient`, which are set by the `ApiClient` when a get instance/user request is made. This will be useful when a community model needs to check if its `ApiClient` has moderator permissions, for example.

Previously, the ApiClient would fetch its version when it is initialised. Instead of doing this, I've added an async getter to the `version` property that fetches the version when it is accessed or, if the version has been fetched already, returns that previously fetched version. If a version-fetch task is already ongoing and `version` is accessed, it will wait for the existing task to finish and return the result. This ensures that we only make one version-fetch request, even if `version` is accessed multiple times in quick succession.

This system will be especially useful when dealing with the age-old "which sort mode should we use?" problem in the feed, which has historically been tricky to solve. Using the `version` property, we can easily do something like:

```swift
// This is pseudocode; methods with these names don't actually exist
.task {
    self.postSortMode = chooseDefaultSortMode(forVersion: await appState.api.version ?? .infinity)
    await self.loadPosts()
}
```


EricBAndrews (approved):
LGTM 🚀

PR #1009, opened 2024-04-13
Bubble Picker value rendering tweaks

Sjmarf:
Tweaked the way bubble picker values are rendered. In the inbox and mod mail, zeroes are shown when there are no items.

Also refactored `abbreviateNumber(int)` into `int.abbreviated`, and made `abbreviated` always round down when approximating. I think this is better than rounding to the nearest value?


EricBAndrews (approved):
LGTM pending passing CI 👍

PR #1010, opened 2024-04-14
Fix interaction bar being reset to default when upgrading to 1.3

Sjmarf:
Also rewrote the piece of code that was causing the lint issue... I upgraded to Swiftlint 0.53.0 and it still complains about that part of the code; I'll look into it some more later


EricBAndrews (approved):
LGTM 👍

PR #1015, opened 2024-04-15
Fix sent messages not being filtered out when "Show only unread" is turned on

Sjmarf:


EricBAndrews (approved):
Nice 👍 

Noticed this one myself this morning, you beat me to it lol

PR #1019, opened 2024-04-16
Block Instances

### LINKED ISSUE #255: Instance Blocking
Pending API support (https://github.com/LemmyNet/lemmy/issues/2397), allow instance blocking

Edit: API support has been added, slating for next cycle

### LINKED ISSUE #672: [0.19.0] Instance blocking
## Introduction
As discussed in the recent Lemmy announcement users will now have the ability to block instances.

> Users can now block instances, so that their communities are hidden from listings. This is done via POST /api/v3/site/block with parameters int instance_id, bool block.

> https://github.com/LemmyNet/lemmy/pull/3869

## Requirements
We should add support for this as soon as possible, the option would need to be gated behind a version check as it will not be available below `0.19.0`.

Sjmarf:
Added a "Block" action to the ellipsis menu in `InstanceView`. Due to an API limitation, the "block" action is only available when you visit the `InstanceView` from a community page or user page (on 0.19.4 and above), and not from search. I've [submitted an issue on the Lemmy repo](https://github.com/LemmyNet/lemmy/issues/4636) to hopefully resolve this limitation in future.

This PR doesn't add the block lists where you'll be able to unblock users/communities/instances - I'll do that in a follow-up PR

Closes #672

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -34,6 +34,7 @@
 		030E86482AC6FD1D000283A6 /* _assignIfNotEqual.swift in Sources */ = {isa = PBXBuildFile; fileRef = 030E86472AC6FD1D000283A6 /* _assignIfNotEqual.swift */; };
 		030E864C2AC7037F000283A6 /* SearchBarExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 030E864B2AC7037F000283A6 /* SearchBarExtensions.swift */; };
 		030FF6862BCB218000F6BFAC /* Int+Abbreviated.swift in Sources */ = {isa = PBXBuildFile; fileRef = 030FF6852BCB218000F6BFAC /* Int+Abbreviated.swift */; };
+		030FF6882BCEE58900F6BFAC /* BlockInstance.swift in Sources */ = {isa = PBXBuildFile; fileRef = 030FF6872BCEE58900F6BFAC /* BlockInstance.swift */; };
 		0317D46F2B558CB500EEE72C /* BadgeView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0317D46E2B558CB500EEE72C /* BadgeView.swift */; };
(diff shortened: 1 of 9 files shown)


EricBAndrews (changes requested):
I've found a couple issues while testing:

- Wrong instance blocked
  - Repro: attempt to block `slrpnk.net` from `lemmy.ml` and check block status using web ui
  - Expected behavior: `slrpnk.net` is blocked
  - Actual behavior: `bildung.social` is blocked
  - Notes: it's at least consistent--unblocking `slrpnk.net` from Mlem removes `bildung.social` from the block list
- `block_already_exists` error on unblocked instance
  - Repro: attempt to block `hexbear.net` from `lemmy.ml`
  - Expected behavior: instance blocked
  - Actual behavior: `instance_block_already_exists` error appears twice. Web UI indicates no block took place.
- Block status not reflected in `InstanceView`
  - Repro: block `hexbear.net` from web ui and navigate to its page in Mlem
  - Expected behavior: menu offers `Unblock` option
  - Actual behavior: menu offers `Block` option

Sjmarf:
Thanks, I'll look into it when I get a chance later.

> Wrong instance blocked

Hmm, interesting. `APISite` objects have both an `id` and `instance_id`, which were always the same during my testing. It's possible that they're different under certain circumstances, and I'm using the wrong one

> block_already_exists error on unblocked instance

This is probably a result of the first issue - accessing the wrong instance id 

> Block status not reflected in InstanceView

This unfortunately isn't possible as there is no `blocked` field of `APISite`. The hope is that the instance page won't be accessible anyways if you've blocked the instance - you can't visit a community or user page from that instance, so you can't get to instance view.

Sjmarf:
> Hmm, interesting. APISite objects have both an id and instance_id, which were always the same during my testing. It's possible that they're different under certain circumstances, and I'm using the wrong one

Turns out this was indeed the case. It should work correctly, now.

EricBAndrews (approved):
LGTM 👍

PR #1020, opened 2024-04-16
Better mark read on scroll

### LINKED ISSUE #1018: * Mark read on scroll marks it too early
**Describe the bug**
When scrolling down the feed posts are marked as read when they're barely halfway through the screen.

**Expected Behavior**
Is they were marked after they left the screen it would make it easier to read before it turning a darker color

**To Reproduce**
Scroll with Mark read on scroll turned on

**Mlem Information:**
1.3 (913)

EricBAndrews:
# Checklist
      - closes #1018 

# Pull Request Information

This PR refines the behavior of mark read on scroll:
- When posts would previously be marked read, they are now *staged* to be marked read
- When posts disappear, if they were staged, they are added to the batch and client-side state faked
This two-step behavior lets us ensure we only mark posts read that have properly been "scrolled past" while hiding the actual state change from the user.

This logic all lives in the `MarkReadBatcher`, which now maintains `staged: Set<Int>` to track staged posts. The view just needs to call `stage` appropriately and blindly call `add` on every post disappear, and the batcher handles the rest.

A nice side effect of this system is that `add` is no longer called using indexed values, which means that the state faking can happen in there without tripping the weird pass-by-reference corner cases that previously rendered that impossible.

### DIFF
--- a/Mlem/ContentView.swift
+++ b/Mlem/ContentView.swift
@@ -212,7 +212,7 @@ struct ContentView: View {
                 
                 // flush batcher(s) to avoid batches being lost on quit
                 Task {
-                    await markReadBatcher.flush(includeStaged: true)
+                    await markReadBatcher.flush()
                 }
                 
                 // activate biometric lock
--- a/Mlem/ContentView.swift
+++ b/Mlem/ContentView.swift
@@ -212,7 +212,7 @@ struct ContentView: View {
                 
                 // flush batcher(s) to avoid batches being lost on quit
                 Task {
-                    await markReadBatcher.flush()
+                    await markReadBatcher.flush(includeStaged: true)
                 }
                 
                 // activate biometric lock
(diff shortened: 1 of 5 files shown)


Sjmarf (changes requested):
This is much nicer!

I'm getting a `Thread 28: EXC_BAD_ACCESS (code=1, address=0x10)` on line 87 of `MarkReadBatcher` when scrolling down quickly in compact mode. I can consistently reproduce this.

I do think it marks posts as read just a *little* too far above the screen...one option might be to mark a post as read when the *previous* post disappears. If you get a super-tall post, this doesn't necessarily guarantee that you will have read the entire thing by the time the previous post disappears, but I don't think that's a big problem. Anyway, super-tall posts only really occur when the post has inline images, and the post description doesn't get truncated like it should (which is a bug).

EricBAndrews:
Nice catch, fixed 👍 

I've wrapped all the `staged` stuff in a semaphore block so it shouldn't ever be touched by two threads at once.

Sjmarf (approved):
LGTM 👍

PR #1024, opened 2024-04-18
Fixed app slowdown with many communities

### LINKED ISSUE #958: Mlem client slows down to a crawl after subscribing to hundreds of communities
I have subscribed to 872 communities after finding many on !trendingcommunities.

I was hoping the app would maintain its normal speed.

The app became too slow to use. Such as loading posts or tapping into comments.

There is a threshold before performance is impacted.

Iphone 13 Pro + IOS 17.3.1 + 1.2 + 847

EricBAndrews:
# Checklist
      - closes #958 

# Pull Request Information

This PR fixes app slowdown when subscribed to hundreds of communities.

The slowdown was caused by `FeedsView` triggering unnecessary recomputations of the community sections, which required iterating over the list of subscribed communities several times and performing several filtering operations, many redundant.

To resolve this, I have implemented the following changes:
- Added a quick lookup set for subscribed and favorited communities
- Made `allSections` and `visibleSections` stored rather than computed, and they are recomputed only when `subscribed` or `favorited` changes
- Refactored section generation logic to avoid unnecessary filter calls (`allSections` and `visibleSections` are now built side-by-side)
- `CommunityListSection` now includes the list of communities in the section. This is now computed when the sections are recomputed rather than on-the-fly.
- Refactored how alphabetic sections are generated. They are now accumulated using a single pass of the sorted `subscribed` list, reducing the number of iterations by a factor of 26
- Removed the `SidebarEntry.contains` function; its logic is now entirely unnecessary. `SidebarEntry` is now a simple struct that contains a label and an icon.
- Removed `allCommunities`

### DIFF
--- a/Mlem/Models/Content/Community/Community List/CommunityListModel.swift
+++ b/Mlem/Models/Content/Community/Community List/CommunityListModel.swift
@@ -31,9 +31,16 @@ class CommunityListModel: ObservableObject {
             .$favoritesForCurrentAccount
             .dropFirst()
             .sink { [weak self] value in
-                self?.updateFavorites(value)
+                if let self {
+                    Task {
+                        await self.updateFavorites(value)
+                    }
+                }
             }
             .store(in: &cancellables)
+        let (newAllSections, newVisibleSections) = recomputeSections()
+        self.allSections = newAllSections
+        self.visibleSections = newVisibleSections
     }
     
     func load() async {
@@ -47,7 +54,7 @@ class CommunityListModel: ObservableObject {
             let newFavorited = favoriteCommunitiesTracker.favoritesForCurrentAccount
             
             // combine the two lists
-            combine(newSubscribed, newFavorited)
+            await combine(newSubscribed, newFavorited)
         } catch {
             errorHandler.handle(
                 .init(underlyingError: error)
@@ -59,9 +66,9 @@ class CommunityListModel: ObservableObject {
         subscribedSet.contains(community.id)
     }
     
-    func updateSubscriptionStatus(for community: APICommunity, subscribed: Bool) {
+    func updateSubscriptionStatus(for community: APICommunity, subscribed: Bool) async {
         // immediately update our local state
-        updateLocalStatus(for: community, subscribed: subscribed)
+        await updateLocalStatus(for: community, subscribed: subscribed)
         
         // then attempt to update our remote state
         Task {
@@ -69,7 +76,7 @@ class CommunityListModel: ObservableObject {
         }
     }
     
-    private func updateLocalStatus(for community: APICommunity, subscribed: Bool) {
+    private func updateLocalStatus(for community: APICommunity, subscribed: Bool) async {
         var newSubscribed = self.subscribed
         
         if subscribed {
@@ -83,7 +90,7 @@ class CommunityListModel: ObservableObject {
             }
         }
         
-        combine(newSubscribed, favorited)
+        await combine(newSubscribed, favorited)
     }
     
     private func updateRemoteStatus(for community: APICommunity, subscribed: Bool) async {
@@ -99,7 +106,7 @@ class CommunityListModel: ObservableObject {
             if let indexToUpdate = self.subscribed.firstIndex(where: { $0.id == updatedCommunity.id }) {
                 var newSubscribed = self.subscribed
                 newSubscribed[indexToUpdate] = updatedCommunity
-                combine(newSubscribed, favorited)
+                await combine(newSubscribed, favorited)
             }
         } catch {
             let phrase = subscribed ? "subscribe to" : "unsubscribe from"
@@ -112,17 +119,15 @@ class CommunityListModel: ObservableObject {
             )
             
             // as the call failed, we need to revert the change to the local state
-            await MainActor.run {
-                updateLocalStatus(for: community, subscribed: !subscribed)
-            }
+            await updateLocalStatus(for: community, subscribed: !subscribed)
         }
     }
     
-    private func updateFavorites(_ favorites: [APICommunity]) {
-        combine(subscribed, favorites)
+    private func updateFavorites(_ favorites: [APICommunity]) async {
+        await combine(subscribed, favorites)
     }
     
-    private func combine(_ subscribed: [APICommunity], _ favorited: [APICommunity]) {
+    private func combine(_ subscribed: [APICommunity], _ favorited: [APICommunity]) async {
         // store the values for future use
         self.subscribed = subscribed
         subscribedSet = Set(subscribed.map(\.id))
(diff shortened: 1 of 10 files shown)


Mlem/Models/Content/Community/Community List/CommunityListModel.swift
```diff
@@ -205,28 +119,125 @@ class CommunityListModel: ObservableObject {
...
+                    sidebarIcon: nil
+                )
+                
+                // This looks sinister but I didn't know how to string replace in a non-string based regex
+                return CommunityListSection(
+                    viewId: character,
+                    sidebarEntry: sidebarEntry,
+                    inlineHeaderLabel: character,
+                    accessibilityLabel: "Communities starting with the letter '\(character)'",
+                    communities: allCommunities.filter { community in
```
Sjmarf, line 236: Because the list is already sorted alphabetically, there's probably a better alternative to using `filter` here. It doesn't change the time complexity though, so not a huge deal

Mlem/Models/Content/Community/Community List/CommunityListModel.swift
```diff
@@ -205,28 +119,125 @@ class CommunityListModel: ObservableObject {
...
+                communities: .init()
+            )
+        }
+        newAllSections.append(topSection)
+        
+        let favoritesSection = withDependencies(from: self) {
+            CommunityListSection(
+                viewId: "favorites",
+                sidebarEntry: FavoritesSidebarEntry(
+                    sidebarLabel: nil,
```
Sjmarf, line 176: I noticed that jumping to the favorites section doesn't work correctly, though it doesn't currently work on `dev` either so not a problem with this PR

Sjmarf (approved):
Looks good! Nothing blocking. Currently there's no way of refreshing the community list, could we add a `.refreshable()` at some point?

Mlem/Models/Content/Community/Community List/CommunityListModel.swift
EricBAndrews: Opened a separate issue for it

Mlem/Models/Content/Community/Community List/CommunityListModel.swift
EricBAndrews: Optimized it anyway--*26n* might be the same as *n* for algo nerds, but a factor of 26 is significant in my book

Mlem/Models/Content/Community/Community List/CommunityListModel.swift
```diff
@@ -205,28 +117,125 @@ class CommunityListModel: ObservableObject {
...
+            } else {
+                while currentSectionIndex < 25, !community.name.uppercased().starts(with: currentSectionTitle) {
+                    currentSectionIndex += 1
+                    currentSectionTitle = alphabet[currentSectionIndex]
+                    communities[currentSectionTitle] = .init()
+                }
+                assert(communities.keys.contains(currentSectionTitle), "No '\(currentSectionTitle)' key in communities!")
+                communities[currentSectionTitle]?.append(community)
+            }
+        }
```
Sjmarf, line 207: > // it's not guaranteed that non-alphabetics be sorted to one side or the other of the alphabetics, so each element does have to be individually checked, hence the quick-lookup alphabetSet

An alternative would be to use custom sorting logic, as shown below. This is slightly faster (~6ms vs ~21ms with a large number of communities) and results in less code. Lemmy only allows lowercase community names, which could have resulted in even faster solutions but unfortunately KBin allows uppercase names :(

```suggestion
        var communities: [String: [APICommunity]] = .init()
        
        var sortedCommunities = subscribed.sorted(by: { item1, item2 in
            // The "~" character has a greater value than all letters in the ASCII table, so prepending it ensures that the community is sorted to the end of the array
            let val1 = (item1.name.first?.isLetter ?? false) ? item1.name.lowercased() : "~" + item1.name.lowercased()
            let val2 = (item2.name.first?.isLetter ?? false) ? item2.name.lowercased() : "~" + item2.name.lowercased()
            return val1 < val2
        })
     
        for section in [String].alphabet {
            communities[section] = Array(sortedCommunities.prefix(
                while: { $0.name.first?.uppercased() == section }
            ))
            sortedCommunities.removeFirst(communities[section]?.count ?? 0)
        }
    
        communities["other"] = sortedCommunities
```

Mlem/Models/Content/Community/Community List/CommunityListModel.swift
```diff
@@ -15,38 +14,45 @@ class CommunityListModel: ObservableObject {
...
+    @Published private(set) var visibleSections: [CommunityListSection] = .init()
+    
+    private(set) var subscribed: [APICommunity] = .init()
+    private var subscribedSet: Set<Int> = .init()
+    private(set) var favorited: [APICommunity] = .init()
+    
     init() {
         favoriteCommunitiesTracker
             .$favoritesForCurrentAccount
             .dropFirst()
```
Sjmarf, line 30: What's `dropFirst` for?

Sjmarf (changes requested):

Mlem/Models/Content/Community/Community List/CommunityListModel.swift
EricBAndrews: Implemented 👍

Mlem/Models/Content/Community/Community List/CommunityListModel.swift
EricBAndrews: 🤔 good question. Once upon a time it might have served a purpose, but I can't find one any more.

EricBAndrews:
I've made it `refreshable`. This creates a visual bug in the Simulator, but it appears to be a Simulator problem--I can't reproduce on-device, so I think it's fine

Screenshot 2024-04-19 at 12 21 36 PM

Mlem/Models/Content/Community/Community List/CommunityListModel.swift
```diff
@@ -205,28 +116,89 @@ class CommunityListModel: ObservableObject {
...
         
-        // update our published value for the view to render
-        mainQueue.schedule { [weak self] in
-            self?.communities = sorted
+        return [String].labelAlphabet.map { character in
+            CommunityListSection(
+                viewId: character,
+                sidebarEntry: .init(sidebarLabel: character, sidebarIcon: nil),
+                inlineHeaderLabel: character,
+                accessibilityLabel: "Communities starting with a symbol or number",
```
Sjmarf, line 199:
```suggestion
                accessibilityLabel: "Communities starting with \(character == "#" ? "a symbol or number" : character)",
```

Mlem/Models/Content/Community/Community List/CommunityListModel.swift
```diff
@@ -205,28 +116,89 @@ class CommunityListModel: ObservableObject {
...
-        let combined = subscriptions + favorites.filter { !subscriptions.contains($0) }
-        let sorted = combined.sorted()
+        assert(sections.values.reduce(0) { x, communities in
+            x + communities.count
+        } == subscribed.count, "mapping operation produced mismatched counts")
         
-        // update our published value for the view to render
-        mainQueue.schedule { [weak self] in
-            self?.communities = sorted
+        return [String].labelAlphabet.map { character in
```
Sjmarf, line 194: Using `labelAlphabet` here means that communities that begin with an accented character (`à`, etc) won't be shown in the list. `Character.isLetter`, which we're using in the grouping logic, returns `true` for these characters. Lemmy won't let you create communities with those characters in the name, but KBin does. 

To fix this, we could either:
- Use `sections.keys.map` here instead.
- Modify the grouping logic to use `[String].labelAlphabet.contains()` instead of `isLetter`.
- Modify the grouping logic to categorize accented characters under their base version. This could get dicey for non-latin alphabets, though.

Sjmarf (changes requested):

Mlem/Models/Content/Community/Community List/CommunityListModel.swift
EricBAndrews: Nice catch 👍 

`sections.keys.sorted().map` seems like the best approach--we lose a shred of time needing to sort it, but it's only ~28 items so that shouldn't matter.

Sjmarf (approved):
LGTM 👍

PR #1029, opened 2024-04-19
Block List

Sjmarf:
- Added a blocklist under account settings.
- Added "blocked"/"unblocked" notifications when blocking from user/community pages.
- All "blocked" notifications now show a checkmark rather than a cross when successful.


Sjmarf:
Update: Blocking instances now also works everywhere (including in the profile and from search). I'm using the hacky workaround described [here](https://github.com/LemmyNet/lemmy/issues/4636).

EricBAndrews (approved):
LGTM 🚀 

Here's hoping Lemmy adds a fix for resolving instance ID lol

PR #948, opened 2024-02-20
Render pinned posts in compact mode if already read.

boscojwho:
# Checklist
- [x] I have read CONTRIBUTING.md
- [x] I have described what this PR contains
- [x] If this PR alters the UI, I have attached pictures/videos
- [x] This PR addresses one or more open issues that were assigned to me:
      - closes #930 

# Pull Request Information

## About this Pull Request
Render pinned posts (`FeedPost`) in compact mode if already read.

PR #1036, opened 2024-04-23
Render pinned posts in compact mode if already read

EricBAndrews:
rebased and fixed logic for #948 

SwiftFormat had a field day with this one, so there are also a bunch of minor formatting diffs


Sjmarf (approved):
Looks good. Should we make this toggle-able in settings?

EricBAndrews:
Let's hold off for now--we can add that toggle if users want it, but I don't predict it being highly requested

Issue #150, opened 2023-06-13
* Subscript displays as strikethrough


adjoseph16:
**Describe the bug**
Mlem renders markdown text that should be subscript as strikethrough. 

**Mlem Information:**
Version: 0.0.6 (22)

I’m using an Apple 13 Pro

Mobile site screenshot:
![IMG_6882](https://github.com/buresdv/Mlem/assets/78319881/b2e25dc6-99b1-4f24-be98-bbd4a0b47bdc)

Mlem screenshot:
![IMG_6880](https://github.com/buresdv/Mlem/assets/78319881/13bb221c-5704-4570-9742-51925317cdf5)

PR #1038, opened 2024-04-25
Markdown rendering

Sjmarf:
This PR replaces `swift-markdown-ui` with a custom parser and renderer to more closely follow Lemmy's API spec. Tables are missing at the moment, but I'll be adding them back in in a future PR. I've included a block of markdown in the "Profile" tab temporarily for testing purposes.

Added support for `^Superscript^`, `~Subscript~` and spoilers:
```
::: spoiler Title!
Content
:::
```

I've tweaked the way images are rendered, too. If an image is on its own line, it'll be rendered full-width. If it is not, it'll be rendered inline. This mirrors the Lemmy-ui behaviour. Rendering images inline allows us to support Lemmy's custom emoji system (they're just inline images). I've set it to resize all inline images to the height of the line which makes sense for emojis, though Lemmy-ui keeps images at full height.

This PR does not:
- Render full-width images - a placeholder is shown instead. We need to rewrite our `CachedImage` system before implementing this.
- Implement tables, as mentioned before.
- Allow nesting spoiler blocks, though Lemmy-ui doesn't support this either.
- Add custom markdown parsing for community or user links (`!community@lemmy.ml` etc). This needs to be done at the `cmark-lemmy` level in a future PR. We should be able to avoid the weird email-based stuff we're doing currently.

To find all of the links in a post/comment so that we can render tappable links, I've included a `[MarkdownBlockNode].links` property that returns all links contained within the markdown, excluding those inside of spoilers. When support for `!community@lemmy.ml` syntax is added, this property could return those links too.


EricBAndrews:
This should close #150, yeah?

EricBAndrews:
LGTM

Still working my way through the MarkdownUI repo--I'll put up comments on that as issues or PRs to that repo, though, so we can go ahead and merge this to hook it up 👍

Sjmarf:
> This should close #150, yeah?

Yep! Closes #150

EricBAndrews (approved):

Issue #1013, opened 2024-04-15
Removing Uploaded Photo Cancels/Clears Post


d42ohpaz:
**Describe the bug**
I was writing a post, and I had uploaded a photo. I decided to change the photo I had uploaded, so I clicked the x on the photo. The popup dismissed, and the login screen popup was displayed saying my session expired. I cancelled the login popup, and the new post popup was redisplayed, but now empty. I lost my entire post. 

**Expected Behavior**
I expected to remove the photo, but keep editing my post without data loss. 

**Actual Behavior**
The popup dismissed, and the login screen popup was displayed saying my session expired. I cancelled the login popup, and the new post popup was redisplayed, but now empty.

**To Reproduce**
Upload a photo. Remove the photo. 

**Mlem Information:**
Version 1.2, Build 847 from TestFlight. 

**Additional Context**
Not sure if it matters, but I was trying to post to a community I am not a member of on a different server than I am logged into.

Sjmarf:
Thanks for the bug report! We're sorry to hear that this happened.

This probably happens due to our current `.sheet` system, in which both the token renewal sheet and post composer sheet are attached at the same level. As two sheets can't display at once at the same level, one closes when the other opens. To solve this, we could have an `AppRoute` equivalent for sheets, and attach a `.sheet(item: Binding<SheetRoute>)` object to every sheet view. This would allow sheets to stack properly, and would resolve the FaceID issue too. This is something we should implement in 2.0.

Sjmarf:
In 2.0 the session expiry sheet will open over the post composer rather than closing the post composer first.

PR #1044, opened 2024-04-28
Navigation System

### LINKED ISSUE #1033: [2.0] Rebuild Navigation
I think we should keep the basic idea of using enum cases to represent different views. In 2.0, we could unify our sheet management system with the `NavigationStack` navigation system. This could help to resolve some sheet-related crashes that we currently experience due to multiple sheets being displayed simultaneously. Ideally, the view-side code could look something like this:

```swift
struct ViewOne: View {
    @Environment(\.navigation) var navigation
    
    var body: some View {
        VStack {
            Button("Push view") {
                navigation.push(.viewTwo)
            }
            Button("Open Sheet") {
                navigation.openSheet(.mySheet)
            }
        }
    }
}
```

Sjmarf:
Closes #1033

This PR re-implements `AppRoute` as `NavigationPage`, with two dummy cases. In addition to pages, sheets are also included as part of this enum. This allows us to easily implement "global" sheets that need to open over the top of any already open sheets, such as the token renewal sheet (related to [this issue](https://github.com/mlemgroup/mlem/issues/1013)) and the FaceID lock sheet.

## How it works

We can define the root view of each tab like so:

```swift
NavigationRootView(root: .page1)
```

Inside of `NavigationRootView`, a `NavigationStack` is displayed with a path attached. This path is stored in a `NavigationLayer` model which is made available as an environment variable to all child views of the `NavigationStack`.

Each view on the `NavigationStack` can access this environment model and push a new view to the stack like so:

```swift
struct PageOne: View {
    @Environment(NavigationLayer.self) var navigation
    
    var body: some View {
        VStack {
            Button("Go to page 2") {
                navigation.push(.page2)
            }
        }
    }
}
```

Opening a sheet or full screen cover works in a similar way:

```swift
 navigation.openSheet(.page2, hasNavigationStack: true)
 ```
 ```swift
  navigation.openFullScreenCover(.page2, hasNavigationStack: true)
  ```

If `hasNavigationStack` is set to `true`, the opened sheet will contain a `NavigationStack` so that the sheet view can use `navigation.push` to push new views within the sheet.

If `.openSheet` is called within the sheet, a second sheet is opened and displayed over the first. Sheets can be stacked infinitely.

## iPad Layout

For the "Feeds" tab, we can use `NavigationSplitRootView` instead of `NavigationRootView`. The `navigation` environment variable is made available in both the sidebar and detail of the `NavigationSplitView`, and buttons from the sidebar are able to modify the detail path properly. This is something that we had issues accomplishing in the past, so I'm glad that it seems to work now - maybe `@Observable` or the new environment variable system is the reason why it now works correctly.

## Showing the token refresh view

Whenever we perform a request on the view-side, we could catch a `notLoggedIn` error and push the token renewal sheet to the stack like so:

```swift
.task {
    do {
        self.posts = try await apiClient.loadPosts()
    } catch ApiClientError.notLoggedIn {
        navigation.openSheet(.tokenRenewal)
    }
}
```

Or, if we re-implement the `errorHandler` system in 2.0, something like this:

```swift
.task {
    do {
        self.posts = try await apiClient.loadPosts()
    } catch {
        errorHandler.handle(error, navigation: navigation)
    }
}
```


EricBAndrews (approved):
LGTM! I presume the dummy pages will be removed before merge lol

PR #1049, opened 2024-05-03
Navigation System Tweaks

Sjmarf:
- Fixed account login sheet not opening.
- The quick switcher and login sheets now use the navigation system (all sheets need to use the navigation system for it to work properly).
- The `.sheet` modifier used for the navigation system is now attached *above* the `TabView`, rather than inside each tab as I had done previously.
- The quick switcher will dismiss itself when the app moves to the background, unless the login sheet is open on-top.
- Added all 5 tabs, with empty views in each. Tab icon discussion raised in Slack.
- Removed unused code.


Sjmarf:
Drafting; I thought of a couple of ways I can improve this

Sjmarf:
Switched to the tab icon design we currently use in 1.0

EricBAndrews (changes requested):
Code looks good, but I can't run it--it compiles successfully, but on launch I get:

`Thread 1: Fatal error: 'try!' expression unexpectedly raised an error: SwiftUI.AnyNavigationPath.Error.comparisonTypeMismatch`

Sjmarf:
Fixed 👍

EricBAndrews (approved):
LGTM 👍

PR #1055, opened 2024-05-09
`Action` getter rework

Sjmarf:
Changed the way that action getters are defined to avoid defining `menuActions` twice.


EricBAndrews (approved):
Nice 👍

Issue #941, opened 2024-02-18
2.0: SettingsView


EricBAndrews:
Good as-is, just needs an update to use new middleware models

PR #1056, opened 2024-05-10
Account List in Settings

Sjmarf:
- Adds the Account List to the settings tab.
- Re-implemented `AvatarView`, `AvatarBannerView` and `AvatarStackView`.
- Added extensions to `NavigationLink` for shorter syntax when navigating to a `NavigationPage`.
- Fixed signing-out of accounts.

Related to #941


Mlem/App/Views/Shared/Avatar/AvatarBannerView.swift
```diff
@@ -0,0 +1,102 @@
...
+    init(_ model: (any ProfileProviding)?, type: AvatarType, showEmptyBanner: Bool = false) {
+        self.model = model
+        self.type = type
+        self.showEmptyBanner = showEmptyBanner
+    }
+    
+    static let bannerHeight: CGFloat = 170
+    static let avatarOverdraw: CGFloat = 40
+    static let avatarSize: CGFloat = 108
+    static let avatarPadding: CGFloat = AppConstants.postAndCommentSpacing
```
EricBAndrews, line 40: Feel free to disregard as OOS, but `postAndCommentSpacing` should be renamed `standardSpacing` and the `halfSpacing` and `doubleSpacing` constants from `dev` should be brought back in--not sure how they didn't make it over lol

Mlem/App/Views/Shared/Avatar/AvatarBannerView.swift
```diff
@@ -0,0 +1,102 @@
...
+    
+    @ViewBuilder
+    func avatarView() -> some View {
+        AvatarView(url: model?.avatar, type: type)
+            .frame(height: AvatarBannerView.avatarSize)
+    }
```
EricBAndrews, line 46: *nit*

Per new CONTRIBUTING; should also be moved below `body`

```suggestion
    @ViewBuilder
    var avatarView: some View {
        AvatarView(url: model?.avatar, type: type)
            .frame(height: AvatarBannerView.avatarSize)
    }
```

Mlem/App/Views/Shared/Avatar/AvatarStackView.swift
```diff
@@ -0,0 +1,74 @@
...
+                        .frame(maxWidth: 0)
+                        .padding(outlineWidth)
+                }
+            }
+            Spacer().aspectRatio(1 / 2, contentMode: .fit)
+        }
+    }
+    
+    @ViewBuilder
+    func plusIconView() -> some View {
```
EricBAndrews, line 48: *nit* per CONTRIBUTING

```suggestion
    var plusIconView: some View {
```

EricBAndrews (changes requested):
Code looks good, a couple nits related to CONTRIBUTING updates.

I also get a couple of compilation errors due to `read` and `saved` still being `isRead` and `isSaved` (FeedsView:113, Interactable2Providing+Extensions:21).

EricBAndrews (changes requested):
Looking good! I've noticed a couple very slight rough edges while testing:
- Signing out of an account that you are not currently logged into from the quick-switcher closes the switcher
- Signing out of the current account puts the app into guest mode--is this intentional? Discussion raised in Slack
- "Enter Guest Mode" is still available in the switcher even when you're already in Guest Mode--on this note, the switcher should probably have an active indication that guest mode is active, as it may appear broken for there to be no green active account indicator

Sjmarf:
> Signing out of an account that you are not currently logged into from the quick-switcher closes the switcher

Fixed 👍 

> "Enter Guest Mode" is still available in the switcher even when you're already in Guest Mode--on this note, the switcher should probably have an active indication that guest mode is active, as it may appear broken for there to be no green active account indicator

This button is temporary - in future we'll let the user add multiple different guest accounts to a section below the list of accounts.

EricBAndrews (approved):
Nice! 🚀

Issue #803, opened 2023-12-08
Comment bar color themes


Sjmarf:
Comments have a vertical colored bar on the leading edge. The color of the bar shows the depth of the comment. The colors used for this should be customisable in Appearance settings.

Apollo had this; someone requested this on mlemapp [here](https://lemmy.world/comment/5890731).

Sjmarf:
Curated color palettes sounds good to me - that's what Apollo had.

Sjmarf:
Possible child of #252

Sjmarf:
Duplicate of #139

Issue #252, opened 2023-07-10
Themes


Sjmarf:
Ability to customise the appearance of the app beyond just light / dark mode. Apollo has this, Memmy has this.

Post 1.0, obviously

BarterClub:
Any easy change would be a color picker for ui option.

PR #1058, opened 2024-05-10
Color Profiles

EricBAndrews:
# Checklist
      - groundwork for #803, #252 

# Pull Request Information

This PR adds support for app-wide color profiles. A new protocol, `PaletteProviding`, defines the set of color definitions that constitute a color profile. In this PR, two profiles are implemented: `StandardPalette` and `MonochromePalette`. Propagating palette information to the app is handled by the global `Palette` object, which is itself `PaletteProviding` and operates according to the standard pattern for making `any Protocol` concrete.

I've only defined colors that we're currently using, on the theory that it will be pretty easy to add colors as we need them. I've opted for semantic rather than literal colors, which will make the protocol a little larger but should give us better flexibility for themes. At some point we're presumably going to need literal colors for things like settings squircles.

This PR also updates CONTRIBUTING.md per Slack


Mlem/App/Globals/Definitions/Palette.swift
```diff
@@ -0,0 +1,67 @@
...
+    
+    var palette: any PaletteProviding {
+        switch self {
+        case .standard:
+            StandardPalette()
+        case .monochrome:
+            MonochromePalette()
+        }
+    }
+}
```
Sjmarf, line 37: Instead of defining separate `StandardPalette` and `MonochromePalette` structs, maybe it would be better to define a single `ColorPalette` struct like so, with instances of the struct defined statically to represent each palette?

```swift
struct ColorPalette: PaletteProviding {
    let accent: Color
    // etc
    
    static let standard: Self = .init(accent: .blue)
    static let monochrome: Self = .init(accent: .primary)
}
```

CONTRIBUTING.md
```diff
@@ -39,10 +39,55 @@ When your code is approved, it can be merged into the `dev` branch by a member o
...
+  init() { ... }
+
+  var body: some View { ... }
+
+  // if necessary
+  var content: some View { ... }
+
+  Helper views
+}
+```
```
Sjmarf, line 70: `static let main` should go in here too.

CONTRIBUTING.md
```diff
@@ -39,10 +39,55 @@ When your code is approved, it can be merged into the `dev` branch by a member o
 ## Conventions
 
 Please develop according to the following principles:
+
 - One View per file. A file containing a View struct must end in "View". We're yet to decide on an official naming scheme for files - feel free to offer your thoughts [here](https://github.com/mlemgroup/mlem/issues/55).
```
Sjmarf, line 43: This should be updated at some point since we appear to have settled on `TitleCase+Extension` syntax

Mlem/App/Constants/Colors/MonochromePalette.swift
```diff
@@ -0,0 +1,22 @@
...
+    let primary = Color.primary
+    let background = Color(UIColor.systemBackground)
+    let secondaryBackground = Color(UIColor.secondarySystemBackground)
+    let tertiaryBackground = Color(UIColor.tertiarySystemBackground)
+    let accent = Color.black
+    let uiAccent = UIColor.black
+    
+    let upvote = Color.black
+    let downvote = Color.black
+    let save = Color.black
```
Sjmarf, line 21: These should be `Color.primary` so that it works correctly in dark mode. This will, however, cause the interaction buttons to appear completely white in dark mode - w'll probably need a separate `selectedInteractionBarItem` color to fix this.

```suggestion
    let accent = Color.primary
    let uiAccent = UIColor.label
    
    let upvote = Color.primary
    let downvote = Color.primary
    let save = Color.primary
```

Mlem/App/Globals/Definitions/Palette.swift
```diff
@@ -0,0 +1,67 @@
...
+import SwiftUI
+
+protocol PaletteProviding {
+    // basics
+    var primary: Color { get }
+    var background: Color { get }
+    var secondaryBackground: Color { get }
+    var tertiaryBackground: Color { get }
+    var accent: Color { get }
+    var uiAccent: UIColor { get }
```
Sjmarf, line 18: `UIColor` has an initialiser that accepts a `Color`, so we can avoid having a separate `uiAccent` property here.

```swift
UIColor(Color.red)
```

Sjmarf (changes requested):

Mlem/App/Globals/Definitions/Palette.swift
EricBAndrews: Yeah, that seems better--it looks like `static` is implicitly `lazy`, too, so it should be equally performant

Sjmarf (approved):
Looks good! 🚀

PR #1063, opened 2024-05-14
ContentLoader

EricBAndrews:
# Checklist

## Pull Request Information

This PR updates `ExpandedPost` to be navigable using `AnyPost` (and therefore using any class in the `PostProviding` family)

It also contains a first iteration at how we can handle loading upgradable content in the form of `ContentLoader`. `ContentLoader` takes in an `Upgradable` (see aforementioned PR) and a `@ViewBuilder` that creates a view from the given `Upgradable`'s `MinimumRenderable` type. If the `Upgradable` doesn't conform to `MinimumRenderable`, a loading spinner is displayed while `upgrade()` is performed. If the `Upgradable` is at least `MinimumRenderable`, the view is displayed; if it is not fully `Upgraded`, then `upgrade()` is called.

This system lets us neatly handle displaying upgradable content regardless of its current upgrade status by simply wrapping it in a `ContentLoader`. I haven't tested this, but we should in theory be able to nest one `ContentLoader` in the view closure of another, letting us render as fast as we can upgrade models.

The notion of `MinimumRenderable` does introduce rigidity in tying "renderable" directly to the model, but considering our models I do not judge at this time that that will be an issue.

### DIFF
--- a/Mlem/App/Views/Shared/ContentLoader.swift
+++ b/Mlem/App/Views/Shared/ContentLoader.swift
@@ -11,27 +11,36 @@ import SwiftUI
 
 struct ContentLoader<Content: View, Model: Upgradable>: View {
     let model: any Upgradable
+    var content: (Model.MinimumRenderable) -> Content
     
-    var content: (Model.Upgraded) -> Content
-    
-    init(model: Model, content: @escaping (Model.Upgraded) -> Content) {
+    init(model: Model, content: @escaping (Model.MinimumRenderable) -> Content) {
         self.model = model
         self.content = content
     }
     
     var body: some View {
-        if let upgradedModel = model.upgraded as? Model.Upgraded {
-            content(upgradedModel)
+        if let modelValue = model.wrappedValue as? Model.MinimumRenderable {
+            content(modelValue)
+                .task {
+                    if !model.isUpgraded {
+                        await upgradeModel()
+                    }
+                }
         } else {
             ProgressView()
                 .task {
-                    print("upgrading...")
-                    do {
-                        try await model.upgrade()
-                    } catch {
-                        print(error)
-                    }
+                    print("model not renderable!")
+                    await upgradeModel()
                 }
+        }
+    }
+    
+    func upgradeModel() async {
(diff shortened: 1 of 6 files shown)


Mlem/App/Views/Shared/ContentLoader.swift
```diff
@@ -0,0 +1,37 @@
...
+                .task {
+                    print("upgrading...")
+                    do {
+                        try await model.upgrade()
+                    } catch {
+                        print(error)
+                    }
+                }
+        }
+    }
```
Sjmarf, line 36: This logic works great when upgrading a `Stub` all the way to `Upgraded`. But what about if we're upgrading a mid-tier model? In those cases, we'd want to display what information we *do* have whilst the upgrade is in progress. For example, we'd be able to display all of the post information using just a `Post1`, except for the status of interactions. Should `ContentLoader` have an option to display the `content` with any non-stub model, or is the intent that we wouldn't be using `ContentLoader` in cases where we need to do this?

Mlem/App/Views/Shared/ContentLoader.swift
```diff
@@ -0,0 +1,46 @@
...
+    
+    var body: some View {
+        if let modelValue = model.wrappedValue as? Model.MinimumRenderable {
+            content(modelValue)
+                .task {
+                    if !model.isUpgraded {
+                        await upgradeModel()
+                    }
+                }
```
Sjmarf, line 28: `.task` is executed whenever navigating back to the tab from another. Theoretically this could cause `upgradeModel` to be called more than once, if `.task` runs a second time whilst `upgradeModel` is in progress. A view-side boolean check would resolve this

Sjmarf (changes requested):

Mlem/App/Views/Shared/ContentLoader.swift
EricBAndrews: Nice catch! Fixed 👍

Sjmarf (approved):
Nice! 🚀

PR #1070, opened 2024-05-20
Move UserStub to Mlem

Sjmarf:
See also the supporting MlemMiddleware PR [here](https://github.com/mlemgroup/MlemMiddleware/pull/9).

- `UserStub` has been moved from MlemMiddleware to Mlem and renamed to `Account`. 
- MlemMiddleware contains a new `Person4` class based on `ApiMyUserInfo`. This replaces `User`, which was removed.
- On Slack I mentioned potentially having a class in `Mlem` that conforms to both `Person4Providing` and `UserStubProviding`. I actually don't think we'll need this, so I've not included it. `Account` is effectively decoupled from the middleware system.
- Renamed `ProfileProviding` to `Profile2Providing` and added a lesser `Profile1Providing` that only requires `name` and `avatar`. `Account` conforms to `Profile1Providing`.
- Mlem now checks for changes of avatar URL and site version when you enter an account, and updates the `Account` accordingly.

In a future PR I'll be adding a `GuestAccount` class that is similar to `Account`.


EricBAndrews (changes requested):
Code looks good, build fails on `api.contentUpdateCallback` (see MlemMiddleware PR)

EricBAndrews (approved):
LGTM 👍

PR #1071, opened 2024-05-25
Saved Guest Accounts

Sjmarf:
You can now save guest accounts to a list below the main accounts. To add a new guest account, tap "Add Account" then "Add Guest". 

I'm temporarily using an alert as the text input for the guest domain. Before 2.0, we will instead show a sheet that allows the user to search for instances. The alert occasionally bugs out and goes full-caps, still :(

It's possible for a guest account to be active without it being a _saved_ guest account. Currently, this only happens if you have no accounts saved, in which case Mlem defaults to `lemmy.world`, and the account is marked as "temporary" in the account list. Switching to a different account will remove the temporary account from the list. In future, there could be other ways of accessing temporary guest accounts. For example, `InstanceView` could have a "Visit as guest" action that logs in as a temporary guest. In the account list, there could be a "Keep" action on the temporary guest item that adds it to the saved guest list.

### Implementation

I've renamed `ActiveAccount` to `UserSession`, so it isn't too similar in name to `Account`. I didn't like calling it `ActiveAccount` because the data-types of variables with names such as `account` wouldn't be immediately obvious. So maybe `UserSession` is better?

`Account` has been renamed to `UserAccount`, and a new `GuestAccount` class has been added. Both conform to an `Account` protocol. Similarly, `UserSession` and `GuestSession` (which replace `ActiveAccount`) both conform to a `Session` protocol.

I've also renamed `AccountListButton` to `AccountListRow` (to match `CommunityListRow` etc used in v1), and moved its rendering code to `AccountListRowBody`.

### DIFF
--- a/Mlem/App/Globals/Definitions/AppState.swift
+++ b/Mlem/App/Globals/Definitions/AppState.swift
@@ -12,34 +12,34 @@ import SwiftUI
 
 @Observable
 class AppState {
-    private(set) var guestAccount: ActiveGuestAccount = .init(url: URL(string: "https://lemmy.world")!)
-    private(set) var activeAccounts: [ActiveUserAccount] = []
+    private(set) var guestSession: GuestSession = .init(url: URL(string: "https://lemmy.world")!)
+    private(set) var activeSessions: [UserSession] = []
 
     func changeAccount(to account: any Account) {
         ToastModel.main.add(.account(account))
         
-        activeAccounts.forEach { $0.deactivate() }
-        guestAccount.deactivate()
+        activeSessions.forEach { $0.deactivate() }
+        guestSession.deactivate()
         
         // Save because we updated `lastUsed` in the above `deactivate()` calls
         AccountsTracker.main.saveAccounts()
         
         if let account = account as? UserAccount {
-            let activeAccount = ActiveUserAccount(account: account)
-            activeAccounts = [activeAccount]
+            let activeAccount = UserSession(account: account)
+            activeSessions = [activeAccount]
         } else if let account = account as? GuestAccount {
-            activeAccounts = []
-            guestAccount = .init(account: account)
+            activeSessions = []
+            guestSession = .init(account: account)
         } else {
             assertionFailure()
         }
     }
     
     func deactivate(account: UserAccount) {
-        if let index = AppState.main.activeAccounts.firstIndex(where: { $0.account === account }) {
-            activeAccounts[index].deactivate()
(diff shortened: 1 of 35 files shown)


Mlem/App/Globals/Definitions/AppState.swift
```diff
@@ -8,120 +8,78 @@
...
+    func deactivate(account: any Account) {
+        if let account = account as? UserAccount {
+            if let index = AppState.main.activeSessions.firstIndex(where: { $0.account === account }) {
+                activeSessions[index].deactivate()
+                activeSessions.remove(at: index)
+            }
+        } else if let account = account as? GuestAccount {
+            guard account == guestSession.account else {
+                assertionFailure("Tried to deactivate a non-active GuestAccount")
+                return
```
EricBAndrews, line 59: This prevents saved guest accounts from being removed if they're not active--is this intentional?

EricBAndrews (changes requested):
Looks good!

Should we maybe change "sign out" to "remove" for guest accounts? It feels weird to sign out of something you didn't sign into

Mlem/App/Globals/Definitions/AppState.swift
```diff
@@ -8,120 +8,78 @@
...
+    var firstSession: any Session { activeSessions.first ?? guestSession }
+    var firstAccount: any Account { firstSession.account }
+    var firstApi: ApiClient { firstSession.api }
     
-    func accountThatModerates(actorId: URL) -> ActiveAccount? {
-        activeAccounts.first(where: { account in
-            account.person?.moderatedCommunities.contains { $0.actorId == actorId } ?? false
+    func accountThatModerates(actorId: URL) -> UserSession? {
+        activeSessions.first(where: { session in
+            session.person?.moderatedCommunities.contains { $0.actorId == actorId } ?? false
```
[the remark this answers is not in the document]
Sjmarf: Sure, I'll throw it in a future MlemMiddleware PR 👍

Mlem/App/Globals/Definitions/AppState.swift
Sjmarf: Nice catch 👍

Sjmarf:
> Should we maybe change "sign out" to "remove" for guest accounts? It feels weird to sign out of something you didn't sign into

Changed to "remove" 👍

EricBAndrews (approved):
LGTM 👍

PR #1073, opened 2024-05-27
Eric/feed post

EricBAndrews:
# Checklist
      - progress towards #61 (very indirectly)

# Pull Request Information

This PR re-implements the base views for feed posts, as well as two required components:
- `EllipsisMenu`: same as in 1.0, but takes an `ActionGroup` now
- `FullyQualifiedLabelView` and `FullyQualifiedNameView`: render a user or community label and name, respectively

`ThumbnailImageView` has been stubbed out pending proper image handling, and for large posts both the image and the website complex are currently using stub views.

Post and community context have been refactored to use `@Environment` values.

Added a new tile post layout; `FeedsView` now uses a `LazyVGrid` instead of a `LazyVStack`, and just uses one column for non-tile layouts.


Mlem/App/Views/Root/Tabs/Feeds/FeedsView.swift
```diff
@@ -116,28 +116,13 @@ struct MinimalPostFeedView: View {
...
-                            ForEach(post.menuActions.children, id: \.id) { action in
-                                MenuButton(action: action)
-                            }
-                        }
+                        FeedPostView(post: .init(post: post))
+                            .contentShape(.rect)
                     }
-                    .buttonStyle(.plain)
+                    .buttonStyle(EmptyButtonStyle())
                     Divider()
```
Sjmarf, line 126: Kinda unrelated, but wrapping the `NavigationLink` and `Divider` in a `VStack(spacing: 0)` results in a significant performance boost

Mlem/App/Models/Settings/Options/PostSize.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Mlem
+//
+//  Created by Eric Andrews on 2024-05-19.
+//
+
+import Foundation
+import SwiftUI
+
+enum PostSize: String, CaseIterable {
+    case compact, tile, headline, large
```
Sjmarf, line 12: Considering how this'll work with the rest of the app, considering that tiled posts can only be shown in feeds - I reckon we should remove `tile` from this enum and have a `experiments.tiledPosts` boolean `AppStorage` value to control whether tiled posts are on or not.

Mlem/App/Views/Shared/Labels/FullyQualifiedLabelView.swift
```diff
@@ -0,0 +1,75 @@
...
+//
+
+import Foundation
+import MlemMiddleware
+import SwiftUI
+
+enum FullyQualifiedLabelStyle {
+    case small(showAvatar: Bool)
+    case medium(showAvatar: Bool)
+    case large(showAvatar: Bool)
```
Sjmarf, line 15: If `showAvatar` is used an argument for all three enum cases, might it be better to make the cases argument-less and have `showAvatar` as a standalone parameter of `FullyQualifiedLabelView`?

```swift
FullyQualifiedLabelView(entity: post.community, labelStyle: .small, showAvatar: true)

File: Mlem/Info.plist
Comment (Sjmarf) [inline Mlem/Info.plist:1]: Are we meant to remove this?

File: Mlem/App/Views/Shared/Labels/FullyQualifiedLabelView.swift [hunk 4]
```diff
@@ -0,0 +1,75 @@
...
+        case .large: .bottom
+        }
+    }
+}
+
+/// View for rendering fully qualified labels (i.e., user or community names)
+struct FullyQualifiedLabelView: View {
+    // these aren't used now but they will be for flairs
+    @Environment(\.postContext) var postContext: (any Post1Providing)?
+    @Environment(\.communityContext) var communityContext: (any Community3Providing)?
```
Sjmarf, line 46: The value of this needs to be set somewhere - it isn't yet. The post view expect `communityContext` to be the context of the community you're viewing, whereas `FullyQualifiedLabelView` expects it to be the community of the post it's attached to. So, `communityContext` needs to be set on the label views from the post views.

Sjmarf (changes requested):
Looks good!

Mlem/App/Constants/Colors/StandardPalette.swift
```diff
@@ -10,15 +10,22 @@ import SwiftUI
...
+        positive: .green,
+        negative: .red,
+        warning: .red,
         upvote: .blue,
         downvote: .red,
         save: .green,
-        selectedInteractionBarItem: .white
+        selectedInteractionBarItem: .white,
+        administration: .cyan,
+        moderation: .red,
```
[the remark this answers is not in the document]
EricBAndrews: You right

Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/TilePostView.swift
```diff
@@ -0,0 +1,176 @@
...
+        dimension + // picture
+            minTitleHeight + // title section
+            (AppConstants.standardSpacing * 3) + // vertical spacing--not actually sure why it has to be 3 instead of 2, but it does
+            2 // spacing between title and community
+    }
+    
+    var body: some View {
+        content
+            .frame(width: dimension, height: frameHeight)
+            .background(palette.background)
```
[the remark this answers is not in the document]
EricBAndrews: Nice catch!

Mlem/App/Utility/Extensions/Content Models/Post1Providing+Extensions.swift
```diff
@@ -17,4 +18,42 @@ extension Post1Providing {
...
+    func taggedTitle(communityContext: (any Community3Providing)?) -> Text {
+        let hasTags: Bool = removed
+            || pinnedInstance
+            || (communityContext != nil && pinnedCommunity)
+            || locked
+        
+        return postTag(active: removed, icon: Icons.removeFill, color: Palette.main.negative) +
+            postTag(active: pinnedInstance, icon: Icons.pinFill, color: Palette.main.administration) +
+            postTag(active: communityContext != nil && pinnedCommunity, icon: Icons.pinFill, color: Palette.main.moderation) +
+            postTag(active: locked, icon: Icons.lockFill, color: Palette.main.orange) +
```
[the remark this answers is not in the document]
EricBAndrews: Yeah, that makes sense. I originally had it as a passthrough literal because "lock" is such a narrow scope to tie a color to compared to more ubiquitous concepts like upvotes, downvotes, moderation, etc; I've renamed this to `secondaryAccent`, since we use it as basically a miscellaneous "other" color (lock, collapse comment swipe background, OP flair--I can't make a coherent group out of those)

Mlem/App/Environment Values/CommunityContext.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Mlem
+//
+//  Created by Eric Andrews on 2024-05-22.
+//
+
+import MlemMiddleware
+import SwiftUI
+
+struct CommunityContext: EnvironmentKey {
+    static let defaultValue: (any Community3Providing)? = nil
```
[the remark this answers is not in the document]
EricBAndrews: Sure--I had originally done it with Community3 to enforce that the model is upgraded because community context is largely useless without the moderators list, so this way we'll just have to ensure that it's handled view-side ourselves.

Mlem/App/Views/Shared/Labels/FullyQualifiedLabelView.swift
EricBAndrews: The design is for `communityContext` to be set:
- By the expanded post view
- By the community feed view
In the feed, there is no community context, and so it is `nil`.

> The post views expect communityContext to be the context of the community you're viewing, whereas FullyQualifiedLabelView expects it to be the community of the post it's attached to.

Outside of a specific community or expanded post, `communityContext` isn't useful to have defined for the label--afaik it's only used to determine whether the user should be flaired as moderator, which can only be determined using a full `Community3Providing`. Since we don't want to upgrade every single community we fetch, it therefore doesn't make much sense to provide this to the label unless we're in a detailed context (expanded post or community feed); in those contexts, the current community and the post's community are identical.

Once we drop 0.18 support, we can get rid of this altogether and just rely on the post context.

Sjmarf: Ah, that makes sense 👍

Mlem/Info.plist
EricBAndrews: No, I have no idea how that happened lol

Mlem/App/Models/Settings/Options/PostSize.swift
EricBAndrews: Sure, that works. It makes a couple switch statements a bit ungainly but maps more accurately to the behavior.

Mlem/App/Models/Settings/Options/InstanceLocation.swift
```diff
@@ -0,0 +1,14 @@
+//
+//  InstanceLocation.swift
+//  Mlem
+//
+//  Created by Jake Shirley on 7/3/23.
+//
+
+import Foundation
+
+enum InstanceLocation: String, CaseIterable {
```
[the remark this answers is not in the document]
EricBAndrews: Sure, that makes sense 👍

Sjmarf (changes requested):
Nice! Found a bug: toggling "Tile Posts" off causes the posts to be rendered in a two-column layout until you refresh

EricBAndrews:
Nice catch!

Sjmarf (approved):
LGTM 🔥

PR #1076, opened 2024-05-30
[1.3] Fix marking as read

Sjmarf:
In 0.19.4, Lemmy removed support for the `postId` parameter of `MarkPostReadRequest`. From 0.19.4 onwards, only the `postIds` parameter can be used. `postIds` was introduced in 0.19.0, but both `postId` and `postIds` were supported until 0.19.4.

Without this fix, an error notification is shown when opening a post or opening the image viewer on a 0.19.4 instance. This PR uses the `postIds` parameter instead of `postId` on instances running 0.19.0 or higher, which fixes the issue.

We should release a 1.3.1 patch with this fix. At the moment, 0.19.4 is in the release candidate stage; `lemmy.ml` is the only major instance running 0.19.4.

### DIFF
--- a/Mlem/API/APIClient/APIClient+Post.swift
+++ b/Mlem/API/APIClient/APIClient+Post.swift
@@ -36,8 +36,13 @@ extension APIClient {
 
     // swiftlint:enable function_parameter_count
     
-    func markPostAsRead(for postId: Int, read: Bool) async throws -> SuccessResponse {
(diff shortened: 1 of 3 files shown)


EricBAndrews (approved):
LGTM!

Linter is mad at this one but I think we can safely ignore it

Sjmarf:
Feel free to merge - I can't because of the linter :D

We should also increment the version number if we're doing a release

PR #1080, opened 2024-06-02
fix lint

EricBAndrews:
Latest version of SwiftLint doesn't like optional data <-> string conversion: https://realm.github.io/SwiftLint/non_optional_string_data_conversion.html

This PR replaces those with the non-optional construction.


Sjmarf (approved):
Nice 👍

PR #1083, opened 2024-06-05
New Filtering

EricBAndrews:
This PR reintroduces the ellipsis menu to the feed view; the hide/show read toggle is there, but others have not been implemented.

It also adds a mockup feed header and builds out the feed view structure a little more; infinite load is enabled again, though it doesn't always work when read posts are hidden (issue open on MlemMiddleware).

Additionally, to support scrolling back to top when the hide/show read toggle is flipped, `FancyScrollView` now has an optional `scrollToTopTrigger`. This is a `Binding<Bool>`; when it changes, the `FancyScrollView` scrolls to the top.


Mlem/App/Views/Shared/FancyScrollView.swift
```diff
@@ -18,17 +18,21 @@ struct FancyScrollView<Content: View>: View {
     
     @ViewBuilder var content: () -> Content
     @Binding var isAtTop: Bool
+    @Binding var scrollToTopTrigger: Bool
```
Sjmarf, line 21: Could we use the existing `isAtTop` binding for this?

Mlem/App/Views/Root/Tabs/Feeds/FeedsView.swift
```diff
@@ -25,27 +24,28 @@ struct FeedsView: View {
...
+    @State var scrollToTopTrigger: Bool = false
     
     init() {
         // need to grab some stuff from app storage to initialize with
-        @AppStorage("internetSpeed") var internetSpeed: InternetSpeed = .fast
-        @AppStorage("upvoteOnSave") var upvoteOnSave = false
-        @AppStorage("showReadPosts") var showReadPosts = true
-        @AppStorage("defaultPostSorting") var defaultPostSorting: ApiSortType = .hot
+        @AppStorage("behavior.internetSpeed") var internetSpeed: InternetSpeed = .fast
+        @AppStorage("behavior.upvoteOnSave") var upvoteOnSave = false
```
Sjmarf, line 40: The toggle for this in the settings tab also needs renaming

Sjmarf (changes requested):
Looks good! I noticed that the feed now erroneously refreshes whenever you switch tabs or navigate to a different page. Maybe something to do with the `.task` changes?

Mlem/App/Views/Shared/FancyScrollView.swift
EricBAndrews: I don't think so--I tried an approach where it would scroll to top when `isAtTop` is set to `true`; this worked fine internally to handle the reselect action, but it wouldn't trigger from the binding. It might be possible, but keeping them separate keeps things stable. I've added a TODO comment to flag the idea for further investigation.

Mlem/App/Views/Root/Tabs/Feeds/FeedsView.swift
```diff
@@ -59,35 +59,65 @@ struct MinimalPostFeedView: View {
...
             }
-            .task {
-                if postTracker.items.isEmpty, postTracker.loadingState == .idle {
-                    print("Loading initial PostTracker page...")
-                    do {
-                        try await postTracker.loadMoreItems()
-                    } catch {
-                        handleError(error)
+            .task(id: showRead) {
+                scrollToTopTrigger.toggle()
```
Sjmarf, line 88: Navigating back from an expanded post or switching tab causes the page to scroll to top. Maybe using `.onChange` here instead would fix this?

Mlem.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
Sjmarf, line 1: My XCode doesn't like this. Are you using the latest XCode beta or something?

Screenshot 2024-06-12 at 18 17 40

I don't think we need to keep `package.resolved` in this repo anyway? From my understand, it's used by XCode internally to keep track of what *has* been downloaded, not what should be downloaded. Maybe we should `.gitignore` it?

EricBAndrews: I've added it to the .gitignore. If it still complains, I can rebuild on the non-beta Xcode and re-commit.

Sjmarf: Hmm, maybe I was wrong 

https://developer.apple.com/documentation/xcode/adding-package-dependencies-to-your-app#Coordinate-package-versions-across-your-team

Sjmarf: I'll download the new beta anyway. Though I don't know whether it might be a bad idea to require prospective contributors to be on the beta lol

Sjmarf (approved):
LGTM, aside from the `Package.resolved` conversation

PR #1086, opened 2024-06-07
Show transition on account switch

### LINKED ISSUE #1065: [2.0] Option to either "keep place" or "reset feed" when switching accounts
This could be a toggle in the "Accounts" page in Settings, possibly labelled something like "Keep Place when switching accounts". The default behavior should be to show the "Welcome <username>" overlay and reset the feed, as in v1.

Sjmarf:
Closes #1065

The toggle for this is under the account list in settings. It's set to reload the feed by default. In the account switcher, you can swipe or long press to access the non-default behaviour.

I also fixed the quick switcher always showing a`.large` sheet, not the `.medium` sheet it's supposed to - this seems to be caused by the `ProgressView`, so I've disabled it for now.

Also made some tweaks the toast system to (hopefully) stop toasts from occasionally getting stuck and not going away.

### DIFF
--- a/Mlem/App/Globals/Definitions/AppState+transition.swift
+++ b/Mlem/App/Globals/Definitions/AppState+transition.swift
@@ -20,6 +20,8 @@ extension AppState {
             window.addSubview(transitionView)
             UIView.animate(withDuration: 0.15) {
                 transitionView.alpha = 1
+            } completion: { _ in
+                self.appRefreshToggle = false
             }
             
             transitionView.translatesAutoresizingMaskIntoConstraints = false
@@ -31,6 +33,7 @@ extension AppState {
                     transitionView.alpha = 0
                 } completion: { _ in
                     transitionView.removeFromSuperview()
+                    self.appRefreshToggle = true
                 }
             }
         }
--- a/Mlem/App/Globals/Definitions/AppState+transition.swift
+++ b/Mlem/App/Globals/Definitions/AppState+transition.swift
@@ -20,20 +20,17 @@ extension AppState {
             window.addSubview(transitionView)
             UIView.animate(withDuration: 0.15) {
                 transitionView.alpha = 1
-            } completion: { _ in
-                self.appRefreshToggle = false
             }
             
             transitionView.translatesAutoresizingMaskIntoConstraints = false
             transitionView.heightAnchor.constraint(equalTo: window.heightAnchor).isActive = true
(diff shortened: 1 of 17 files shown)


Mlem/App/Globals/Definitions/AppState+transition.swift
```diff
@@ -0,0 +1,38 @@
+//
+//  AppState+transition.swift
```
EricBAndrews, line 2: *nit*

Capitalization

```suggestion
//  AppState+Transition.swift
```

Mlem/App/Globals/Definitions/AppState.swift
```diff
@@ -12,26 +12,56 @@ import Observation
...
         self.guestSession = .init(account: AccountsTracker.main.defaultGuestAccount)
-        changeAccount(to: AccountsTracker.main.defaultAccount, deactivateOldGuest: false)
+        setAccount(to: AccountsTracker.main.defaultAccount)
     }
     
-    func changeAccount(to account: any Account) {
-        changeAccount(to: account, deactivateOldGuest: true)
+    /// If `keepPlace` is `nil`, use the value from `UserDefaults`.
+    func changeAccount(to account: any Account, keepPlace: Bool? = nil) {
+        let keepPlace = keepPlace ?? UserDefaults.standard.bool(forKey: "accounts.keepPlace")
```
EricBAndrews, line 43: Why `UserDefaults` instead of `AppStorage`?

Mlem/App/Views/Root/TransitionView.swift
```diff
@@ -0,0 +1,34 @@
...
+import SwiftUI
+
+struct TransitionView: View {
+    let accountName: String?
+    @State var accountNameOpacity: CGFloat = .zero
+    
+    var body: some View {
+        VStack(spacing: 24) {
+            if let accountName {
+                Text("Welcome")
```
EricBAndrews, line 18: Guest accounts displaying "Welcome lemmy.world" is slightly strange--it would make more sense for them to display something like "Welcome to lemmy.world"

EricBAndrews (changes requested):
Looks good!

Mlem/App/Globals/Definitions/AppState.swift
Sjmarf: Sure, I can use `@AppStorage` instead 👍

EricBAndrews (approved):
LGTM 🚀

PR #1087, opened 2024-06-12
Content resolution changes

Sjmarf:
See also [the supporting MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/19).

If `ContentLoader` fails to upgrade some content, it will try again using the instance that the content is hosted on. When this happens, a banner is shown at the top of the screen:

Tapping on the `?` button in the banner will open a sheet showing some more information:

When the active account changes, `ContentLoader` will attempt to resolve the model on the new account. If it fails, it will continue to show the model fetched from the outgoing account.

For testing purposes, I've added an "Open lemmy.world user" button to the inbox tab. Tapping this whilst using a Beehaw guest account will trigger the new behavior.


Mlem/App/Views/Pages/ExternalApiInfoView.swift
```diff
@@ -0,0 +1,170 @@
...
+    }
+    
+    @ViewBuilder
+    var content: some View {
+        VStack(spacing: 16) {
+            box(spacing: 16) {
+                if internalFederationStatus?.isAllowed ?? false, externalFederationStatus?.isAllowed ?? false {
+                    Text(
+                        // swiftlint:disable:next line_length
+                        "Your instance and **\(entityLocalApi.host ?? "")** appear to be federating with one another, but we weren't able to access this content from your instance. This could be because the content is new, and hasn't federated to your instance yet. Or, it could be because your instance has purged this content."
```
EricBAndrews, line 59:
```suggestion
                        "Your instance and **\(entityLocalApi.host ?? "")** federate, but the content could not be loaded. It may not have federated yet, or your instance may have purged it."
```

We don't *technically* speaking know for 100% certain that they federate, but we're sure enough that it's not worth the extra words imo

Mlem/App/Views/Pages/ExternalApiInfoView.swift
```diff
@@ -0,0 +1,170 @@
...
+                } else {
+                    avatars
+                        .padding(.horizontal, 16)
+                    Text(text)
+                        .multilineTextAlignment(.center)
+                        .padding(.horizontal, 16)
+                }
+            }
+            box(alignment: .leading) {
+                Text("To work around this, we're accessing this content from **\(fallbackApi.host ?? "")** instead of your instance.")
```
EricBAndrews, line 71:
```suggestion
                Text("This content will be loaded from **\(fallbackApi.host ?? "")** instead.")
```

Mlem/App/Views/Pages/ExternalApiInfoView.swift
```diff
@@ -0,0 +1,170 @@
...
+            }
+            box(alignment: .leading, spacing: 6) {
+                Text("What is Federation?")
+                    .font(.title2)
+                    .fontWeight(.bold)
+                    .padding(.horizontal, AppConstants.standardSpacing)
+                
+                Text(
+                    // swiftlint:disable:next line_length
+                    "Lemmy instances talk to each other so that content can be shared across sites. This is called \"federation\". Instance administrators can choose which other instances they would like their instance to federate with. Some instances will federate with all other instances, except those on a \"block-list\" curated by the administrators. Other instances might only federate to instances on an \"allow-list\"."
```
EricBAndrews, line 82:
```suggestion
                    "Lemmy instances talk to each other so that content can be shared across sites. This is called \"federation\". Instance administrators can choose which other instances they would like their instance to federate with. Some instances federate with all but a curated \"block-list\" of other instances; other instances might only federate with instances on an \"allow-list\"."
```

Mlem/App/Views/Pages/ExternalApiInfoView.swift
```diff
@@ -0,0 +1,170 @@
...
+        spacing: CGFloat = 16,
+        @ViewBuilder content: () -> some View
+    ) -> some View {
+        VStack(alignment: alignment, spacing: spacing) {
+            content()
+        }
+        .padding(.vertical, 16)
+        .frame(maxWidth: .infinity)
+        .background(
+            Color(uiColor: .secondarySystemGroupedBackground),
```
EricBAndrews, line 102:
```suggestion
            palette.secondaryGroupedBackground,
```

EricBAndrews (changes requested):
Looks good! This should go a long way towards smoothing out the weird UX effects of federation.

I've left a couple of wording edits, open to workshopping as always; I also have some comments on the middleware PR that may have downstream effects here.

EricBAndrews (approved):
LGTM 🔥

PR #1090, opened 2024-06-13
Image loading

Sjmarf:
Images are now shown in the feed (in all 4 feed modes). Tap on an image to open it in the image viewer. Image height in Large mode is limited to 1.2 times image width. It has anti-feed-jumpiness.

I think that the `.medium` community/person links look better than the `.large` ones in large posts, too. What do you think? I've also added the truncated body text back to `LargePost`, and the expanded post view now displays a `LargePost`.


Sjmarf:
Closes #1089

Mlem/App/Views/Shared/ImageView.swift
```diff
@@ -0,0 +1,87 @@
+//
+//  TappableImageView.swift
```
EricBAndrews, line 2:
```suggestion
//  ImageView.swift
```

Mlem/App/Views/Shared/TappableImageView.swift
```diff
@@ -0,0 +1,29 @@
...
+    var body: some View {
+        ImageView(url: url, onLoadingStateChange: { loading = $0 })
+            .onTapGesture {
+                if loading == .done, let url {
+                    // Sheets don't cover the whole screen on iPad, so use a fullScreenCover instead
+                    if UIDevice.isIPad {
+                        navigation.showFullScreenCover(.imageViewer(url))
+                    } else {
+                        navigation.openSheet(.imageViewer(url))
+                    }
```
EricBAndrews, line 25: I doubt this will be the only occurrence of this pattern; we may wish to define a `multiplatformAction` function analogous to the `MultiplatformView`. Possibly OOS for this PR, content to merge as-is.

EricBAndrews (approved):
LGTM!

A couple tiny nits, nothing worth the full "request changes" cycle.

Mlem/App/Views/Shared/TappableImageView.swift
Sjmarf: Would that be particularly beneficial? It would be more verbose to do it that way I think?

```swift
multiPlatformAction(pad: {
    navigation.showFullScreenCover(.imageViewer(url))
}, phone: {
    navigation.openSheet(.imageViewer(url))
})
```

vs

```swift
if UIDevice.isPad {
    navigation.showFullScreenCover(.imageViewer(url))
} else {
    navigation.openSheet(.imageViewer(url))
}
```

PR #1091, opened 2024-06-13
MultiplatformView

EricBAndrews:
This PR introduces the `MultiplatformView`, a construct designed to streamline creating different views for different platforms. Its operation is very simple: it takes in two `View` closures, one to render for iPhone and one to render for iPad, and renders the one appropriate for the current platform.

```
var body: some View {
  MultiplatformView {
    contentForPhone
  } pad: {
    contentForPad
  }
}
```

Ideally we would do this with compile-time checks to avoid the runtime conditional, but I can't seem to find a good way to check for iPadOS at compile time.


Sjmarf (approved):
LGTM

PR #1092, opened 2024-06-14
Thread Safe Content Loader

EricBAndrews:
This PR fixes an issue where the `ContentLoader` could be triggered twice. The `upgradeState` variable has been replaced by an `AsyncSemaphore`.


Sjmarf (approved):
Nice 👍

PR #1104, opened 2024-06-20
Add Paste Link from Clipboard Button

### LINKED ISSUE #1103: [2.0 or before] A way to open link from clipboard in Mlem
https://lemmy.ml/post/17008269

Sjmarf:
Closes #1103

This is the best place I could think to put it, lmk if you think somewhere else would be better:


EricBAndrews (approved):
LGTM! 🔥

PR #1109, opened 2024-06-24
[1.3] v2 `SavedAccount` compatibility

Sjmarf:
Added compatibility for decoding Mlem v2's saved accounts. This ensures that the saved accounts aren't lost if the user downgrades from v2 to v1 (by leaving the beta).


EricBAndrews (changes requested):
Nice! Good corner case to think of--should save our users some frustration.

Core functionality works, but it breaks unit tests. It looks like the `/api/v3` suffix on the `instanceLink` and the placeholder token (`redacted` vs. `token_1`) are the source of the fails.

EricBAndrews (approved):
LGTM 👍

PR #1113, opened 2024-06-26
Comments

Sjmarf:
Added a basic comment view. `ExpandedPostView` uses a simple array to store the comments for now. In future we should use an actual tracker for this to handle multiple pages. See also [the accompanying MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/24).

### DIFF
--- a/Mlem/App/Views/Shared/CommentView.swift
+++ b/Mlem/App/Views/Shared/CommentView.swift
@@ -23,8 +23,9 @@ struct CommentView: View {
                 HStack {
                     FullyQualifiedLinkView(entity: comment.creator_, labelStyle: .small, showAvatar: true)
                     Spacer()
-                    EllipsisMenu(actions: comment.menuActions, size: 24)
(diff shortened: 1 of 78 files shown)


Mlem/App/Views/Shared/CommentView.swift
```diff
@@ -0,0 +1,67 @@
...
+//
+
+import LemmyMarkdownUI
+import MlemMiddleware
+import SwiftUI
+
+struct CommentView: View {
+    @Environment(Palette.self) private var palette
+    
+    private let threadingColors = [Color.red, Color.orange, Color.yellow, Color.green, Color.blue, Color.purple]
```
EricBAndrews: Fine to leave OOS for this PR, but these should be defined in `Palette`

EricBAndrews (approved):
Looks good!

Happy to leave OOS for this PR, but we shouldn't be using `monospaceDigits` for dates:

Screenshot 2024-07-01 at 4 29 46 PM

PR #1118, opened 2024-06-30
Fix `ContentLoader`

Sjmarf:
Fixes the state update issue we had with `ContentLoader`. I've done this by calling the `content` method passed to `init` inside of the initializer itself rather than in `body`.


EricBAndrews (approved):
LGTM 🚀 

Very nice solution!

PR #1119, opened 2024-07-01
Link Resolution

### LINKED ISSUE #1047: [2.0] Lemmy link handling
When a link is clicked, we need to detect whether the link is a Lemmy link. If so, we should attempt to display the page in-app rather than opening the browser. We already do this in 1.0, but I think the system could be improved for 2.0.

I personally don't like showing a toast to indicate loading status like we do in 1.0. Instead, I think it would be nicer if a page gets pushed to the navigation stack immediately, showing a loading symbol there. We can do this by creating a stub model (eg `CommunityStub`) and passing it to the relevant `NavigationPage` enum case, something like this:

```swift
let communityStub = CommunityStub(api: appState.firstApi, actorId: communityLink)
navigation.push(.community(communityStub))
```

Pushing a page immediately is a smooth experience if the link *was* correctly identified as a Lemmy link, but it's less smooth if our prediction was incorrect. In the case of false positives, we'd have to pop the page we just pushed before pushing the safari view instead. Alternatively, we could show an error message in the page we just pushed, with an "Open in browser" button.

To reduce the chance of false positives occurring, we could improve the way that our link system checks for Lemmy domains. Possibly we could compare the domain against a list of the top n instance domains (we already store a domain list for use in the search homepage), or even the entire list of Lemmy domains if it's reasonable to store that. We could also continue to check for `/u/` and `/c/` in the domain name, though we wouldn't know for certain if a link containing one of these points to a Lemmy domain. 

I intend to parse `!community@lemmy.world` and `@user@lemmy.world` links properly in the markdown module, so we can remove the weird email-hacking that we used to detect these in 1.0.

Sjmarf:
Adds logic for resolving Lemmy links. Closes #1047


Mlem/App/Views/Root/Tabs/Settings/LinkSettingsView.swift
```diff
@@ -0,0 +1,44 @@
...
+                .pickerStyle(.inline)
+                .labelsHidden()
+            }
+            
+            Section {
+                Toggle(
+                    "Open in Reader",
+                    isOn: Binding(
+                        get: {
+                            !openLinksInBrowser && openLinksInReaderMode
```
EricBAndrews: I don't think we need to flip the toggle when in-app browser is disabled--just disabling the toggle should be clear enough along with the explanatory text.

EricBAndrews (approved):
LGTM!

One minor design note, non-blocking

Mlem/App/Views/Root/Tabs/Settings/LinkSettingsView.swift
Sjmarf: Sure 👍

PR #1120, opened 2024-07-01
[2.0] Fix Haptics Adjusting System Volume

### LINKED ISSUE #629: Launching mlem switches volume controls to Alarm+Notification volume
**Describe the bug**
On my iPhone, (iOS 16.6.1, iPhone 11 Pro Max) I have my sound settings configured as so:
1. Settings -> Sounds & Haptics -> Ringtone And Alert Volume:
  a. Volume level: max
  b. Change With Buttons: disabled

This makes it so that my alarm and ringtone volume is "locked" at maximum volume. Changing the volume with the buttons or through the Control Center (swipe down from top right corner of the phone) should *only* affect the volume of the speaker (music, game sounds, videos, etc).

I've noticed however, that after launching mlem, changing the volume, at least through Control Center, changes the Alarm & Ringtone volume instead.

**Expected Behavior**
My alarm and ringtone volume should not be changed when I adjust the volume

**Actual Behavior**
After launching mlem, if I swipe into Control Center and say, drop the volume down to zero, my alarms and ringtones are now muted!

**To Reproduce**
This doesn't appear to be 100% reproducible, but what I've seen:
1. Force quit mlem if it's running
1. Configure iOS so that the ringtone volume is 100% and that the buttons do *not* change the ringtone volume
2. Completely mute the volume of music/video/games on your phone by holding down the volume down button
3. Swipe down from the top to open Control Center, confirm that the volume bar is at 0%
4. Now launch mlem
5. Swipe down from the top to open Control Center again and the volume bar will be at 100%
6. Adjust that volume bar so it's at 0%
7. Now go to Settings -> Sounds & Haptics and you'll see that the ringtone volume is at 0%

Note: I reproduced this several times in a row this morning but now that I'm typing this up it's not reproducing the behavior. My guess is that this might be related to content that's rendered by the app? or perhaps this is somehow a bug in iOS itself? I will say that I've never had this problem until I installed mlem, so it at least seems to be triggered by something the app is doing.

**Mlem Information:**
Version: 1.0.2, Build: 687

**Additional Context**
I'll update this issue if I can find a reliable way to reproduce it 100% of the time

### LINKED ISSUE #660: Forces my volume to be on when app first opens*
**Describe the bug**
When I first open the app, my volume turns on one tick and I am unable to turn it off completely. It only happens the first time I open the app -- once I put it in the background, the volume automatically goes back to zero and I can bring the app back up and the volume will not change.

https://github.com/mlemgroup/mlem/assets/137654063/d1947464-244c-467d-a819-af2a2ddef797

**Expected Behavior**
For this not to happen.

**To Reproduce**
I open the app from fully closed.

**Mlem Information:**
V1.1, Build 697, test flight 

**Additional Context**
???????? Is all I've got really. Super weird, but maybe it's a me problem. This problem occurred before the IOS 17 update (I first noticed it August 16th according to an old screen recording) and has persisted after it.

EricBAndrews:
# Checklist
      - closes #660 
      - closes #629 

# Pull Request Information

This PR fixes the haptic engine changing system volume on launch.

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -1541,7 +1541,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = "local build";
 				DEVELOPMENT_ASSET_PATHS = "\"Mlem/Preview Content\"";
-				DEVELOPMENT_TEAM = 76ULHQGAPN;
+				DEVELOPMENT_TEAM = 8B9GNJW88W;
 				ENABLE_PREVIEWS = YES;
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = Mlem/Info.plist;
@@ -1561,7 +1561,7 @@
 					"@executable_path/Frameworks",
 				);
 				MARKETING_VERSION = 1.3;
-				PRODUCT_BUNDLE_IDENTIFIER = com.hanners.Mlem;
+				PRODUCT_BUNDLE_IDENTIFIER = eric.local;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SUPPORTED_PLATFORMS = "iphoneos iphonesimulator";
 				SUPPORTS_MACCATALYST = YES;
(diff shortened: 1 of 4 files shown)


Sjmarf (approved):
LGTM! It's great to eliminate this long-standing issue 🎉

PR #1127, opened 2024-07-03
[2.0] Haptic Fix

Sjmarf:
Fixes the issue where music would pause when opening the app. This doesn't reintroduce the ringer volume issue. No, I don't know why this fixes it xD


EricBAndrews (approved):
Nice! 🚀

Issue #938, opened 2024-02-18
2.0: FeedsView


EricBAndrews:
Good as-is, just needs update to new middleware models

PR #1131, opened 2024-07-05
More Feed Types

EricBAndrews:
# Checklist
      - closes #issue
      - progress towards #938 

# Pull Request Information

This PR adds the `All` and `Local` feed types, which includes:
- Re-introduced `FeedHeaderView` and `FeedIconView`
- Added `FeedDescription`, a data model to drive `FeedHeaderView`
- Created `PostGridView`, a helper view that handles rendering posts from a `StandardPostFeedLoader`
- Created `.loadFeed`, a convenience view modifier for performing initial `FeedLoader` loading

Additional changes:
- Added feed colors to `PaletteProviding`
- Added initializer to `ColorPalette`. In certain instances (e.g., `MonochromePalette`), colors such as `Palette.moderatedFeed` should be different from `Palette.moderation`; the initializer gives us the power to override those if needed but alias to the expected color if not.
- Fixed a bug where tab reselect was not reliably scrolling the feed to top. `SubscriptionListView` also had a `tabReselectConsumer`, which due to navigation quirks did not appropriately register appearance/disappearance; this resulted in both the `FeedsView` and the `SubscriptionListView` trying to consume the same reselect action. I have fixed this by handling tab reselection in the `SubscriptionListView` with a hand-rolled reselect consumer that queries the navigation path directly to determine whether to consume the reselect action.
- Fixed a bug where `FancyScrollView` was not updating `isAtTop` when `isAtTop` was not passed in as a `Binding`. `isAtTop` has been changed to a state variable inside `FancyScrollView`, and is communicated to parent views using a `PreferenceKey`. `Profile` has been updated to use this system.
- Fixed a bug where tapping navigation links in `SubscriptionListView` would crash the app due to concurrent access errors. The navigation path work is now `MainActor`.


Mlem/App/Views/Root/Tabs/Feeds/Feed Header/FeedHeaderView.swift
```diff
@@ -0,0 +1,92 @@
...
+//  FeedHeaderView.swift
+//  Mlem
+//
+//  Created by Eric Andrews on 2024-01-28.
+//
+
+import Foundation
+import SwiftUI
+
+struct FeedHeaderView: View {
```
Sjmarf, line 11: The padding on the feed header is a little off in "Tile" mode

Mlem/App/Views/Root/Tabs/Feeds/Feed Header/FeedDescription.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+
+import Foundation
+import MlemMiddleware
+import SwiftUI
+
+struct FeedDescription {
+    var label: String
+    var subtitle: String
+    var color: Color?
```
Sjmarf: These color values don't update after changing theme until app restart 🤔

Mlem/App/Views/Root/Tabs/Feeds/Feed Header/FeedHeaderView.swift
```diff
@@ -0,0 +1,92 @@
...
+    @AppStorage("beta.tilePosts") var tilePosts: Bool = false
+    
+    @Environment(AppState.self) var appState
+    @Environment(Palette.self) var palette
+    
+    let feedDescription: FeedDescription
+    let actions: [any Action]
```
Sjmarf: *Should* we be using actions here? They make sense in cases where where we need to display them in many different contexts, but otherwise I think they're more restrictive and verbose to instantiate (we're using 55 lines for these 3 buttons in `FeedsView`, which is quite a lot). Maybe a simple `ViewBuilder` would be better here? That would allow us to use a `Picker` in this menu, for example, which might look better and be more accessibility-friendly.

Mlem/App/Views/Root/Tabs/Feeds/Feed Header/FeedHeaderView.swift
```diff
@@ -0,0 +1,92 @@
...
+                        .font(.footnote)
+                        .foregroundStyle(palette.secondary)
+                }
+                .frame(height: 44)
+                .frame(maxWidth: .infinity, alignment: .leading)
+            }
+            .padding(.top, AppConstants.halfSpacing)
+            .padding(.bottom, tilePosts ? 0 : AppConstants.standardSpacing)
+            
+            if !tilePosts {
```
Sjmarf: We'll be using this view in the Inbox too, and this check makes the view specific to the main feed. Could this be a `showBottomDivider` parameter passed to the `FeedHeaderView` maybe?

Mlem/App/Views/Root/Tabs/Feeds/FeedsView.swift
```diff
@@ -99,20 +76,10 @@ struct MinimalPostFeedView: View {
...
-                } catch {
-                    handleError(error)
-                }
-            }
-            .task {
-                // NOTE: this is here due to an error in StandardPostFeedLoader where changing feed type doesn't properly reload, resulting in
-                // the first render not triggering a load. I'm currently working on a fix, but it's OOS for filtering
-                // -Eric
-                if postFeedLoader.items.isEmpty, postFeedLoader.loadingState == .idle {
+            .onChange(of: appState.firstApi, initial: false) { newValue, _ in
```
Sjmarf, line 114: Huh, I didn't know about `initial`. Nice

Mlem/App/Views/Root/Tabs/Feeds/FeedsView.swift
```diff
@@ -130,81 +97,38 @@ struct MinimalPostFeedView: View {
...
+        case let .aggregateFeed(_, type):
+            switch type {
+            case .all: FeedHeaderView(feedDescription: .all, actions: headerMenuActions)
+            case .local: FeedHeaderView(feedDescription: .local, actions: headerMenuActions)
+            case .subscribed: FeedHeaderView(feedDescription: .subscribed, actions: headerMenuActions)
+            case .moderatorView: FeedHeaderView(feedDescription: .moderated)
             }
-            .frame(height: 44)
-            .frame(maxWidth: .infinity, alignment: .leading)
+        case .community: FeedHeaderView(feedDescription: .subscribed) // TODO:
```
Sjmarf: This is kinda OOS, but I'm not sure I like `.community` being a part of this enum. The community view is similar to the aggregate feed in that it's a feed... but so is the profile and so is the search feed. Maybe there's a good reason for `.community` being included that I'm forgetting? I remember there being a lot of `"this shouldn't happen"` comments in relation to the `.community` case in v1, which seems like something that would be cleaner to avoid.

Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListView.swift
```diff
@@ -30,6 +31,14 @@ struct SubscriptionListView: View {
         .navigationBarTitleDisplayMode(.inline)
     }
     
+    var isDisplayed: Bool {
```
Sjmarf: The identifier could be a little more descriptive here - maybe `isDetailDisplayed`?

Mlem/App/Views/Shared/FancyScrollView.swift
```diff
@@ -64,11 +67,13 @@ struct FancyScrollView<Content: View>: View {
             }
             .coordinateSpace(name: "scrollView")
             .onPreferenceChange(ScrollOffsetKey.self) { offset in
+                print(offset)
```
Sjmarf:
```suggestion
```

Sjmarf (changes requested):
Nice! That's a clean fix for the sidebar tab re-selection bug - it was bugging me today lol

Mlem/App/Views/Root/Tabs/Feeds/FeedsView.swift
EricBAndrews: There's definitely a cleaner way to handle it, but it'll require some reworking of how the `PostFeedLoader` works--I'll take a look at it once this PR is in.

Mlem/App/Views/Root/Tabs/Feeds/Feed Header/FeedDescription.swift
EricBAndrews: Makes sense--they're initialized once and then the struct is never re-initialized when the palette changes.

EricBAndrews:
I've made some changes:

- `FeedHeaderView` is now completely dumb; it performs no menu work at all, and just renders the pretty header.
- `FeedHeaderView` `showDropdown` and `showDropdownBadge` are now consolidated into a single `DropdownStyle` enum. This also removes the need for the awkward assertion that a badge is not requested on a feed with no dropdown.
- `FeedsView` now has a `FeedSelection` enum that it uses to drive the feed selection. Header menu work has been moved into `FeedsView`
- `FeedDescription.color` is now a function that takes in a `Palette` and returns a `Color?`. The `FeedIconView` passes it the palette, and so re-evaluates the color when the palette changes.

Sjmarf (approved):
LGTM 🔥 

(MlemMiddleware version needs updating, ofc)

PR #1141, opened 2024-07-13
Expanded Post comment resolution

### LINKED ISSUE #1140: [2.0] Expanded post comment resolution on switching accounts

Sjmarf:
## `FeedsView`

If "Reload on switch" is off, the user will be prompted to refresh the feed when they switch accounts.

## `ExpandedPostView`

`ExpandedPostView` will now immediately attempt to resolve if the post it is instantiated with isn't a member of the active `ApiClient`. Previously, it wouldn't detect that the initial post is from an external account and would just display the external post without attempting to resolve it on the active account.

`ExpandedPostView` will also now resolve comments when the account is switched. It does this by fetching the post's comments from the new API, then modifying the `CommentWrapper` tree to replace old comments with new ones. 

Sometimes, a comment will be present on one API but not on another. If the comment was present on the old API but not on the new one, the old comment will remain in the tree but will not be able to be interacted with. I think this is a better alternative than removing those comments, which would cause the feed to jump. If a comment is present on the new API but not the old one, the new comments will be inserted into the feed - this _will_ cause the feed to jump, but I don't see a better alternative.


Mlem/App/Views/Shared/ExpandedPostView.swift
```diff
@@ -41,22 +46,41 @@ struct ExpandedPostView: View {
...
+                }
             } else {
                 ProgressView()
                     .tint(palette.secondary)
             }
         }
     }
     
-    @ViewBuilder
+    // @ViewBuilder
```
EricBAndrews: Dead code?

EricBAndrews (changes requested):
Looking good--something's a little funky with the re-resolution, though. If I open a post in a guest account and switch to an authenticated account, the post doesn't re-resolve (or at least, isn't interactable) until I switch tabs and switch back again.

Sjmarf:
Hmm, I'm struggling to reproduce the issue 🤔 Does it happen every time you do that, on every post?

EricBAndrews:
I tested this a bit more thoroughly--it's limited to when the guest account and the main account are on the same instance, which makes me think the issue is that `ExpandedPostView:50` isn't detecting the switch from an unauthenticated to an authenticated account within the same instance. Everything works fine when going from guest on one instance to user on another.

Sjmarf:
Thanks, that was the issue! Fixed 👍

EricBAndrews (approved):
LGTM 👍

PR #1142, opened 2024-07-14
Search

### LINKED ISSUE #939: 2.0: SearchView
Good as-is, just needs update to new middleware models

### LINKED ISSUE #1115: [2.0] Reusable single-entity search sheet
We had something like this in v1 for some of the moderation sheets

Sjmarf:
Adds the search tab (closes #939). Only one page of results at the moment, and only the user links work. Also added single-entity search sheets (closes #1115). Buttons for testing these sheets are at the bottom of the settings tab. A large portion of the line count is copy-pasting the SwiftUIX `SearchBar` from v1.

The "Add Guest" button now opens the new instance search sheet rather than an alert.


Mlem.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -33,8 +33,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/mlemgroup/MlemMiddleware.git",
       "state" : {
-        "revision" : "0165be3756d51d5489727658e2933b08f650978c",
-        "version" : "0.15.0"
+        "branch" : "sjmarf/block-list",
+        "revision" : "43166557b50d186e7950ee2d9fd494c6d4875eb4"
```

Mlem/App/Enums/PersonFlair.swift
```diff
@@ -0,0 +1,73 @@
...
+        case .admin:
+            return .teal
+        case .moderator:
+            return Palette.main.moderation
+        case .op:
+            return .orange
+        case .bot:
+            return .indigo
+        case .bannedFromInstance, .bannedFromCommunity:
+            return .red
```
EricBAndrews:
```suggestion
            return Palette.main.negative
```

Mlem/App/Enums/PersonFlair.swift
```diff
@@ -0,0 +1,73 @@
...
+    var color: Color {
+        switch self {
+        case .admin:
+            return .teal
+        case .moderator:
+            return Palette.main.moderation
+        case .op:
+            return .orange
+        case .bot:
+            return .indigo
```
EricBAndrews: *nit*

Palette; useful place for accent + semantic coloring discussed in Inbox

Mlem/App/Views/Shared/Search/Results/CommunityListRowBody.swift
```diff
@@ -0,0 +1,142 @@
...
+                    .padding(9)
+            } else {
+                AvatarView(url: community.avatar?.withIconSize(128), type: .community)
+                    .frame(height: 46)
+            }
+            
+            VStack(alignment: .leading, spacing: 2) {
+                Text(title)
+                    .lineLimit(1)
+                    .foregroundStyle(community.nsfw ? .red : .primary)
```
EricBAndrews:
```suggestion
                    .foregroundStyle(community.nsfw ? palette.warning: palette.primary)
```

Mlem/App/Views/Shared/Search/Results/CommunityListRowBody.swift
```diff
@@ -0,0 +1,142 @@
...
+                caption
+                    .font(.footnote)
+                    .foregroundStyle(.secondary)
```
EricBAndrews:
```suggestion
                    .foregroundStyle(palette.secondary)
```

Mlem/App/Views/Shared/Search/Results/InstanceListRowBody.swift
```diff
@@ -0,0 +1,68 @@
+//
+//  CommunityListRowBody.swift
```
EricBAndrews:
```suggestion
//  InstanceListRowBody.swift
```

Mlem/App/Views/Shared/Search/SearchSheetView.swift
```diff
@@ -0,0 +1,88 @@
...
+                }
+                results = try await Item.search(
+                    api: appState.firstApi,
+                    query: query,
+                    page: 1,
+                    limit: 20
+                )
+            } catch {
+                handleError(error)
+            }
```
EricBAndrews: This is a ton of work to put on the `MainActor`--everything except the actual assignments to `results` should happen on a background thread (especially `Task.sleep`). You can feel the hang if you enter a query and immediately dismiss the sheet--it's reasonably benign here because there aren't other animations to interrupt, but still a pattern to avoid.

```suggestion
        .task(id: query, priority: .userInitiated) {
            do {
                if !query.isEmpty {
                    try await Task.sleep(for: .seconds(0.2))
                }
                let response = try await Item.search(
                    api: appState.firstApi,
                    query: query,
                    page: 1,
                    limit: 20
                )
                Task { @MainActor in
                    results = response
                }
            } catch {
                handleError(error)
            }
        }
```

Mlem/App/Views/Shared/Search/SearchView.swift
```diff
@@ -0,0 +1,155 @@
...
+                            }
+                        case .instances:
+                            if instances.isEmpty {
+                                instances = try await MlemStats.main.searchInstances(query: query)
+                            }
+                        }
+                    } catch {
+                        handleError(error)
+                    }
+                }
```
EricBAndrews: Ditto comments above re: main actor work. Similar issue noted below--probably worth defining a little

```
@MainActor
func setCommunities(_ newCommunities: [Community2]) {
  communities = newCommunities
}
```

etc. for people, instances, so that this can just be

`setCommunities(try await appState.firstApi.searchCommunities(query: query, page: 1, limit: 20))`

Mlem/App/Views/Shared/Search/SearchView.swift
```diff
@@ -0,0 +1,155 @@
...
+            }
+            .task(id: query, priority: .userInitiated) { @MainActor in
+                guard !hasAppeared || searchBarFocused else { return }
+                hasAppeared = true
+                await refresh()
+            }
+            .onChange(of: appState.firstApi.actorId) {
+                Task { @MainActor in
+                    await refresh()
+                }
```
EricBAndrews: `@MainActor` work--see comment in `refresh`

Mlem/App/Views/Shared/Search/SearchView.swift
```diff
@@ -0,0 +1,155 @@
...
+                communities = try await appState.firstApi.searchCommunities(query: query, page: 1, limit: 20)
+            case .users:
+                people = try await appState.firstApi.searchPeople(query: query, page: 1, limit: 20)
+            case .instances:
+                instances = try await MlemStats.main.searchInstances(query: query)
+            }
+        } catch {
+            handleError(error)
+        }
+    }
```
EricBAndrews: This whole thing is being run on the main actor from L80. I suggest removing that `@MainActor` annotation and adopting a similar construction to the search sheet suggestion (assuming `@MainActor` setters from previous comment):

```suggestion
    func refresh() async {
        do {
            if !query.isEmpty {
                try await Task.sleep(for: .seconds(0.2))
            }
            setCommunities(.init())
            setPeople(.init())
            setInstances(.init())
            switch selectedTab {
            case .communities:
                setCommunities(try await appState.firstApi.searchCommunities(query: query, page: 1, limit: 20))
            case .users:
                setPeople(try await appState.firstApi.searchPeople(query: query, page: 1, limit: 20))
            case .instances:
                setInstances(try await MlemStats.main.searchInstances(query: query))
            }
        } catch {
            handleError(error)
        }
    }
```

EricBAndrews (changes requested):
Works very nicely! The search sheets are really clean.

I've left some comments about cleaning up unnecessary `@MainActor` work, plus some `palette` nits.

I've also noticed that switching to the `Users` tab on initial load can be quite sluggish--I think this is due to the avatars loading in, but I'm not immediately sure the best way to address it. Maybe we can put some batch prefetching into the search logic? Totally fine to raise an issue and call it OOS for this PR as well.

Mlem/App/Enums/PersonFlair.swift
Sjmarf: Oops, `PersonFlair` isn't meant to be in this PR. I'll remove it for now, it can be added in later. I agree that's a good place for the accent colors 👍

Sjmarf:
> I've also noticed that switching to the Users tab on initial load can be quite sluggish--I think this is due to the avatars loading in, but I'm not immediately sure the best way to address it. Maybe we can put some batch prefetching into the search logic? Totally fine to raise an issue and call it OOS for this PR as well.

In v1, we fetched results for _every_ tab whenever the search query changed, so that we could create the "Top Results" tab. This also meant that results appeared instantly when you switch to another tab. There isn't a "Top Results" tab in v2, so instead only the current tab's results are fetched. When you switch to a different tab, the results are then fetched and displayed. That's why it takes a short amount of time to load in.

Mlem/App/Views/Shared/Search/SearchView.swift
Sjmarf: I've marked the properties themselves as `@MainActor`, which I think is the same as doing this? Xcode doesn't seem to be complaining about it.

Mlem/App/Views/Shared/Search/SearchView.swift
```diff
@@ -33,9 +33,9 @@ struct SearchView: View {
     @State var selectedTab: Tab = .communities
     @State var resultsScrollToTopTrigger: Bool = false
     
-    @State var communities: [Community2] = []
-    @State var people: [Person2] = []
-    @State var instances: [InstanceSummary] = []
+    @MainActor @State var communities: [Community2] = []
+    @MainActor @State var people: [Person2] = []
+    @MainActor @State var instances: [InstanceSummary] = []
```
EricBAndrews, line 38: 👀 that's a neat trick

EricBAndrews: I'm now getting 

`Main actor-isolated property '<communities/people/instances>' can not be referenced from a non-isolated context`

on anything that touches these (L119, 123, 127, 140-142, 145, 147, 149) when I try to build on Xcode 15

EricBAndrews (changes requested):

Mlem/App/Views/Shared/Search/SearchView.swift
Sjmarf: Hmm, maybe it's an Xcode 16 thing then. I'll add the `set...` methods instead 👍

Sjmarf: Done

EricBAndrews (approved):
LGTM! 🔥

Issue #1117, opened 2024-06-29
[2.0] Full-featured feeds


EricBAndrews:
Restore full feed functionality:
- Multiple feed types
- Sorting from nav bar
- Rich feed header

EricBAndrews:
Closed in #1202, #1209

Issue #937, opened 2024-02-18
2.0: Show posts and comments in profile view


EricBAndrews:
ProfileView (old UserView) needs a full rewrite

Sjmarf:
We should make the `UserView` feed more like a feed in 2.0 - it should load infinitely and support sorting (we can do this via the search API).

PR #1157, opened 2024-07-24
Saved Feed

EricBAndrews:
# Checklist
      - Progress towards #1117 
      - Progress towards #937 
      - Progress towards #938 

# Pull Request Information

This PR depends on [this very tiny MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/37).

## App Info

The app version has been bumped to 2.0 and the team switched to mine; this should let us get our dev testflight back up and running.

## Saved Feed

Basically identical to 1.0. It makes no sense to show the saved feed for an unauthenticated account, so it behaves as follows:
- The option to show the saved feed is only displayed if the current user is authenticated
  - If the user switches from an authenticated to an unauthenticated session and the saved feed is currently rendered, it is switched to all
  - If the user switches from one authenticated session to another, the saved feed behaves like a normal feed

## Tiled Comments

This PR also introduces the tiled comments view, which allows showing posts and comments together in a tiled view. The "tiled posts" option has therefore been removed, and "tile" is just a post size choice.

This PR also performs some cleanup to make the tile sizing math more transparent.


Mlem/App/Views/Root/Tabs/Feeds/Feed Comments/TileCommentView.swift
```diff
@@ -0,0 +1,111 @@
...
+import SwiftUI
+
+struct TileCommentView: View {
+    @Environment(Palette.self) var palette
+    
+    let comment: any Comment2Providing
+    
+    @ScaledMetric(relativeTo: .footnote) var titleHeight: CGFloat = 36 // (2 * .footnote height), including built-in spacing
+    @ScaledMetric(relativeTo: .caption) var communityHeight: CGFloat = 16 // .caption height, including built-in spacing
+    var width: CGFloat { (UIScreen.main.bounds.width - (AppConstants.standardSpacing * 3)) / 2 }
```
Sjmarf: Using `UIScreen` in the calculation here causes issues when the tiled posts are displayed in a non-full-width context, such as on iPad when the sidebar is shown. It should be possible to avoid using `UIScreen` using `GeometryReader` or `.aspectRatio()` or both.

Sjmarf:
Good job! I really like the tiled comment design.

The saved feed doesn't appear to be loading multiple pages of saved content. Is that intentional at this stage?

Also, the Markdown doesn't seem to be rendering long paragraphs properly. Each line shown here is the start of a long paragraph:





This might be fixable by using a fixed frame or something? If not you could switch to the more malleable (but less fancy) `MarkdownText`, like is used in `LargePost`.

Mlem.xcodeproj/project.pbxproj
```diff
@@ -1949,8 +1969,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware.git";
 			requirement = {
-				kind = upToNextMinorVersion;
-				minimumVersion = 0.15.0;
+				branch = "eric/saved-feed-tweaks";
+				kind = branch;
```

Mlem/App/Views/Root/Tabs/Feeds/FeedsView.swift
```diff
@@ -80,11 +89,24 @@ struct FeedsView: View {
...
+                api: AppState.main.firstApi,
+                userId: firstUser.id,
+                sortType: .new,
+                savedOnly: true,
+                smallAvatarSize: AppConstants.smallAvatarSize,
+                largeAvatarSize: AppConstants.largeAvatarSize
+            ))
+            _feedOptions = .init(wrappedValue: FeedSelection.allCases)
+        } else {
+            _feedOptions = .init(wrappedValue: FeedSelection.guestCases)
```
Sjmarf, line 109: This value isn't updated when the active account changes. This either needs to be updated via an `.onChange` when the active account changes, or be a computed property rather than an `@State` property.

Mlem/App/Views/Root/Tabs/Feeds/FeedsView.swift
```diff
@@ -132,6 +170,7 @@ struct FeedsView: View {
                         do {
                             showRefreshPopup = false
                             try await postFeedLoader.refresh(clearBeforeRefresh: true)
+                            try await savedFeedLoader?.refresh(clearBeforeRefresh: true)
```
Sjmarf: We shouldn't refresh the `postFeedLoader` here when the saved feed is selected

Sjmarf (changes requested):

Mlem/App/Views/Root/Tabs/Feeds/FeedsView.swift
EricBAndrews: Nice catch, that functionality got eaten by the merge 😶

Mlem/App/Views/Root/Tabs/Feeds/Feed Comments/TileCommentView.swift
EricBAndrews: Good point. I've updated it to read width from an environment key, `parentFrameWidth`, which is currently set at the `PostGridView`/`PersonContentGridView` level--that should give us the granularity we need to handle those cases.

EricBAndrews:
I've made a couple (very minor) middleware changes, see [here](https://github.com/mlemgroup/MlemMiddleware/pull/38)

Sjmarf:
Tiled posts don't seem to be laid out properly, at least on iPad:

Simulator Screenshot - iPad Pro 13-inch (M4) - 2024-07-28 at 21 41 26

EricBAndrews:
Whoops, didn't test what happens when you switch feeds 😅 

Should be fixed now

Mlem/App/Environment Values/ParentFrameWidth.swift
```diff
@@ -0,0 +1,20 @@
...
+struct ParentFrameWidth: EnvironmentKey {
+    static let defaultValue: CGFloat = .zero
+}
+
+extension EnvironmentValues {
+    var parentFrameWidth: CGFloat {
+        get { self[ParentFrameWidth.self] }
+        set { self[ParentFrameWidth.self] = newValue }
+    }
+}
```
Sjmarf, line 20: OOS but these days [there's a cleaner way](https://developer.apple.com/documentation/swiftui/entry()) of doing these definitions. We should probably move all of our `EnvironmentValues` definitions into one file and use `@Entry`

Sjmarf (changes requested):
In dark mode there's no contrast around the box at the top of a comment tile:



Also the tiled post titles seem to be limited to one line - I believe we allowed two lines in the past? It looks like there's enough space for a second line

The feed switches from Saved -> All when you switch from authenticated account -> guest account. We should also switch Subscribed -> All when that happens

Mlem/App/Environment Values/ParentFrameWidth.swift
EricBAndrews: Wow, that's way better. IDK why all the guides I've found do it the hard way lol

I'll look into rolling it into my config PR, it feels adjacent in an information infrastructure sort of way

EricBAndrews:
Nice catch on the dark mode color!

> Also the tiled post titles seem to be limited to one line - I believe we allowed two lines in the past? It looks like there's enough space for a second line

That is the intended behavior, and what I see when I run it locally:

Screenshot 2024-07-28 at 8 44 58 PM

I've tested it across the whole range of font sizes, and it works as intended. What Xcode/Simulator pairing are you using to produce the error?

> The feed switches from Saved -> All when you switch from authenticated account -> guest account. We should also switch Subscribed -> All when that happens

👍

Sjmarf:
I'm using my physical iPhone 12 with default font size. Here's some screenshots:

image

image

Sjmarf:
> The feed switches from Saved -> All when you switch from authenticated account -> guest account. We should also switch Subscribed -> All when that happens

On second thoughts... maybe we shouldn't switch this automatically? When the user switches account whilst looking at All/Local, we _don't_ refresh and we show a "Feed is outdated" banner. This allows the user to tap on posts fetched on the old account from the new account. Forcibly switching from Subscribed to All and from Saved to All feels inconsistent with the aforementioned behavior, now that I think about it. Maybe instead we should show the "Feed is outdated" banner, but have the button show "Switch to All" rather than "Reload", or something like that? What do you think?

EricBAndrews:
> I'm using my physical iPhone 12 with default font size. Here's some screenshots:

I cannot reproduce this issue 🤔 . I've tried:
- On-device (iPhone Xs, iOS 17)
- iPhone 15/iOS 17 simulator
- iPhone 12/iOS 17 simulator
- iPhone 15/iOS 18 simulator, built against Xcode 16
- iPhone 12/iOS 18 simulator, built against Xcode 16

All cases show the correct line breaking behavior.

Re: feed switching, there are good arguments either way, but I think is is correct to switch from saved/subscribed to all when switching to a guest account for a couple reasons:
- No added value: retaining place while switching to another authenticated account is a useful feature because it lets you save/reply/etc from that account. This functionality does not provide any utility when switching to a guest account.
- Saved/Subscribed is a view that (obviously) cannot be navigated to within a guest account, so it is also inconsistent to make it accessible in such an account
- Increased technical cost: Rendering saved/subscribed in a guest account requires adding logic to handle a "dead" `PersonContentFeedLoader` and update the menu properly (keep current selection when switching accounts, then re-evaluate options when switching feeds to clear out the old one). None of this is infeasible or even very hard, but it does add complexity for a feature that, again, doesn't give the user any more power than they would have without it.

Sjmarf:
Interesting, I can't reproduce it on a simulator either. I'll try to work out what the issue is.

Those are good points about feed loading, I agree that switching to All is the best option 👍

Sjmarf:
It turns out that my physical device uses a slightly greater line spacing than the simulator. The sizes of the characters themselves is the same.



The height of two lines is 37.667 pixels on my physical device. It's 36 pixels on the simulator, and the view's height calculations rely on it being 36 pixels. Exactly 36 pixels are allocated to the `Text` view, so there isn't enough space to show a second line. 

I don't know how useful this is because point size != pixel size, but the point size of the `.footnote` font on the simulator is `15.513671875`, and `17.513671875000004` on my real device, which is a difference of 2 points.
 
 I was able to fix the issue by adding `.typesettingLanguage` as shown below. After doing this, two lines is exactly 36 pixels high on my physical device.
 
 ```swift
 // TilePostView line 71
 VStack(spacing: 7) {
    titleSection
        .typesettingLanguage(.init(languageCode: .english))
    
    communityAndInfo
}
.padding(AppConstants.standardSpacing)
 ```
 
(Repeating this for the `TileCommentView` title fixes the issue there, too).

I'm not entirely sure _why_ there's a difference in line height. I've ensured that the following settings are the same on both devices:

- Font Size
- Bold Text
- Display Zoom

It's possible that my physical device, which is British, uses a different line spacing for some reason due to region? I've tried setting the simulator to use `English (UK)` as the language, setting the region to the UK and emulating the simulator location but wasn't able to make it use the taller line spacing. Setting `.typesettingLanguage(.init(languageCode: .english, region: .unitedKingdom))` also fails to procure the taller line spacing.

Some other solutions would be to hardcode the line spacing with `.lineSpacing`, or to build the `TilePostView` using `.frame` and `.aspectRatio` rather than explicitly calculating sizes of view components. Here's an alternative solution that preserves the larger line spacing by making the size of the gap between the title and community a pixel smaller:

```swift
 // TilePostView line 71
VStack(spacing: 0) {
    titleSection
        .frame(maxHeight: .infinity, alignment: .top)
        .layoutPriority(1)
    
    communityAndInfo
}
```

Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/TilePostView.swift
```diff
@@ -56,25 +68,25 @@ struct TilePostView: View {
             
             Divider()
             
-            titleSection
-                .padding(AppConstants.standardSpacing)
+            VStack(spacing: 7) {
+                titleSection
```
Sjmarf, line 72:
```suggestion
                titleSection
                    .typesettingLanguage(.init(languageCode: .english))
```

Mlem/App/Views/Root/Tabs/Feeds/Feed Comments/TileCommentView.swift
```diff
@@ -0,0 +1,115 @@
...
+        content
+            .frame(width: width, height: frameHeight)
+            .background(palette.secondaryGroupedBackground)
+            .clipShape(.rect(cornerRadius: AppConstants.tilePostCornerRadius))
+            .contentShape(.contextMenuPreview, .rect(cornerRadius: AppConstants.tilePostCornerRadius))
+    }
+    
+    var content: some View {
+        VStack(alignment: .leading, spacing: AppConstants.standardSpacing) {
+            titleSection
```
Sjmarf, line 48:
```suggestion
            titleSection
                .typesettingLanguage(.init(languageCode: .english))
```

Sjmarf (approved):
LGTM!

Issue #146, opened 2023-06-15
Localization support


buresdv:
## Introduction
As it is now, the app doesn't support localization due to every string being hard-coded.

## Requirements
Extract all strings into a localizable file

EricBAndrews:
Closing this as we got localization set up in #1159--it should be straightforward to implement a given language now.

PR #1159, opened 2024-07-26
Basic localization

Sjmarf:
Thankfully Apple's system for this is actually really easy to use - introducing it doesn't really impede development speed at all. I think it's worth doing now so that we don't have to do as much rewriting later. 700 lines of the diff is the auto-generated translation file.

Here's the [article](https://developer.apple.com/documentation/xcode/localizing-and-varying-text-with-a-string-catalog) and [video](https://developer.apple.com/videos/play/wwdc2023/10155/) on how localization works. 

I've changed the `Action` and `ToastType` initialisers to use the same system that SwiftUI Views do - if a string literal is provided, it is assumed to be a `LocalizedStringResource` unless a string is specifically supplied.

Apple recommends using the english translation of a string as the string key. This is advantageous because it makes the code less verbose, and avoids accidentally having duplicate strings in the translation file (e.g. "Inbox" in the tab name and "Inbox" in the feed header).

Related to #146


EricBAndrews (approved):
LGTM! 🙌

Issue #1171, opened 2024-07-30
[2.0] Scroll gestures sometimes ignored on iOS 18


Sjmarf:
Occasionally when you try to scroll up or down in the feed on iOS 18 the view ignores the gesture. It seems that changing the quick-swipe `.highPriorityGesture` to `.gesture` fixes it. There's a comment on the `.highPriorityGesture` line about why it needs to be there:

```swift
 // needs to be high priority or else dragging on links leads to navigating to the link at conclusion of drag
```

If that's causing it, we need to find a solution that fixes both issues. Since this is new behaviour in iOS 18, it's also possible that it'll be fixed by Apple before final release.

EricBAndrews:
Some _very_ preliminary testing suggests the drag/link behavior is not present when downgrading from `.highPriorityGesture` to `.gesture` in the 2.0 codebase. I will implement and further test this change in my upcoming website preview PR.

Sjmarf:
This still happens when using Xcode 16, but it's fixed on Xcode 15. When Xcode 16 comes out properly, I'll see if the bug is still there and look into fixing it if so.

PR #1176, opened 2024-07-30
Swipe Actions Toggle

Sjmarf:
Adds a toggle for swipe actions. This is a good idea to have anyway, but I'm doing it now as a temporary workaround to #1171 because it's driving me crazy lol.


EricBAndrews (approved):
👍

PR #1179, opened 2024-07-31
New App Icons

Sjmarf:
- Added dark and tinted variants of each icon
- Removed clay/s icons
- Added pride icon

Because each app icon asset now contains three images, using `UIImage(named: )` to display the images in the picker no longer works. I spent a while trying to fix this, but couldn't work it out. I'll try again before the full v2.0 release - worst-case scenario, I can just add duplicate (lower-resolution) images of each icon to the asset catalogue and use those. I've opened an issue [here](https://github.com/mlemgroup/mlem/issues/1178) to keep track of this.


Mlem/App/Utility/Extensions/Bundle+Extensions.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  Bundle+IconFileName.swift
```
EricBAndrews: *nit*
```suggestion
//  Bundle+Extensions.swift
```

Mlem/App/Views/Shared/CollapsibleSection.swift
```diff
@@ -0,0 +1,70 @@
...
+                }
+                .font(.footnote)
+                .contentShape(.rect)
+                .onTapGesture { withAnimation(.default) { collapsed.toggle() }}
+                .padding(.vertical, 6)
+                .padding(.horizontal, 16)
+            }
+            
+            if !collapsed {
+                Color(uiColor: .systemGroupedBackground)
```
EricBAndrews: *nit*
```suggestion
                palette.groupedBackground
```

Mlem/App/Views/Shared/CollapsibleSection.swift
```diff
@@ -0,0 +1,70 @@
...
+                VStack {
+                    content()
+                }
+                
+                if let footer {
+                    Text(footer)
+                        .textCase(.uppercase)
+                        .font(.footnote)
+                        .padding(.horizontal, 16)
+                        .foregroundStyle(.secondary)
```
EricBAndrews: *nit*
```suggestion
                        .foregroundStyle(palette.secondary)
```

Mlem/App/Views/Shared/CollapsibleSection.swift
```diff
@@ -0,0 +1,70 @@
...
+                }
+            }
+        }
+        .background(Color(uiColor: .secondarySystemGroupedBackground))
```
EricBAndrews: *nit*
```suggestion
        .background(palette.groupedBackground))
```

EricBAndrews (changes requested):
It looks like some files got missed in the commit--I'm getting "build input files not found" on the whole `.../Settings/Icon` tree, and they're not showing up in Finder:

Screenshot 2024-08-02 at 11 01 58 PM

Sjmarf:
`Icon` was gitignored 🤔 I've fixed it 👍

EricBAndrews (changes requested):
Code looks good, and the new dark app icons are lovely!

I've noticed a couple problems:
- App icon is always the dark variant, regardless of phone light/dark mode--I'm also seeing "The app icon set <icon set> has 2 unassigned children" warnings on all icon sets, which is likely the cause
- Changing the icon works, but doesn't update the selector:

https://github.com/user-attachments/assets/ab989c27-966c-4754-afc7-f7474a222984

Sjmarf:
It seems Xcode 15 isn't able to build the new icon file format correctly. It only builds the first image in each asset (alphabetically). I've renamed the light-theme images to `aaa.png`, which causes them to take priority over the dark variants. The icons will only change with the theme if the app is built on Xcode 16.

EricBAndrews (approved):
LGTM 👍 

The Xcode 15 icons behavior isn't ideal since that's the current production build system, but 16 releases in a month so it's probably fine.

PR #1180, opened 2024-08-01
Community View

Sjmarf:
Initial CommunityView stuff. The "posts" and "about" tabs are functional.

I ran into some state update issues stemming from `ContentLoader`... again. Last time, switching from calling `content` in the view body to calling it in the initialiser fixed the issues we were having. Now moving it back fixes the issue I was having with _this_ PR. I can no longer reproduce the original issue, so I'm hoping somewhere in the slew of `ContentLoader` changes it got resolved.


Mlem.xcodeproj/project.pbxproj
```diff
@@ -2150,8 +2170,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware.git";
 			requirement = {
-				kind = upToNextMajorVersion;
-				minimumVersion = 0.19.0;
+				branch = "sjmarf/feed-loader-configuration";
+				kind = branch;
```

Mlem/App/Views/Pages/Community/CommunityView.swift
```diff
@@ -0,0 +1,197 @@
...
+            if let community = proxy.entity, let postFeedLoader {
+                content(community: community, postFeedLoader: postFeedLoader)
+                    .refreshable {
+                        showRefreshPopup = false
+                        do {
+                            try await postFeedLoader.refresh(clearBeforeRefresh: false)
+                        } catch {
+                            handleError(error)
+                        }
+                    }
```
EricBAndrews: Maybe bikeshedding, but we duplicate this same logic in `FeedsView` (with a little extra to pick the loader)--we maybe want to unify it into a `.refreshFeed` extension similar to `.loadFeed`. That would require some minor reworking of `FeedsView` to fit the canonical pattern.

EricBAndrews (changes requested):
Looks good!

Tile posts currently don't display quite right--the background doesn't change to `.groupedBackground` and they need `AppConstants.standardSpacing` padding on top of the feed.

Mlem/App/Views/Pages/Community/CommunityView.swift
Sjmarf: Good call 👍 Requires [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/41)

EricBAndrews (changes requested):
Nice job on the view modifier!

The refresh has a bit of a stutter, which is fine for now--it's also present in the main feed on `dev2`, but for some reason is more apparent in the community feed:

https://github.com/user-attachments/assets/ab6599a6-7f10-47fb-8cee-47a30ef39da5

This may be a middleware problem--again, content to leave OOS for this PR and fix down the road.

Tile posts also don't display quite right--the background doesn't get updated and they're pressed up against the header.

Sjmarf:
> The refresh has a bit of a stutter, which is fine for now--it's also present in the main feed on dev2, but for some reason is more apparent in the community feed:

Yeah, I did notice this but couldn't find a way to fix it. It's possible the issue is in middleware somewhere.

> Tile posts also don't display quite right--the background doesn't get updated and they're pressed up against the header.

Done 👍 Sorry, I forgot about this one :/

Sjmarf:
I think I've got a fix for the refresh thing, but it does in-fact require an MlemMiddleware PR. It seems to be the same issue that causes the stutter when comments load in. The fix doesn't require any changes in Mlem, so this is good to merge

EricBAndrews (approved):
LGTM! 🚀

PR #1182, opened 2024-08-02
LemmyMarkdownUI update

Sjmarf:
Updates LemmyMarkdownUI to include some bug-fixes and tweaks. 

The main change here is to how Mlem decides whether to render an image inline or full-size. Previously, an image would have to be on it's own line to be counted as full-size. The problem with that method is that people frequently _don't_ put images on their own line, which causes Mlem to incorrectly display those images as inline emoji. 

When you insert a custom emoji, it adds a tooltip to the inserted image in this format:

```
![alt_text](https://example.com/image.png "tooltip")
```

The new solution is to display all images that have a tooltip as inline emojis, and display all other images as full-size. I think this is a good enough fix because people are unlikely to use tooltips outside of custom emojis.

Full diff: https://github.com/mlemgroup/LemmyMarkdownUI/compare/0.1.0...0.3.0

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -2095,7 +2095,7 @@
 			repositoryURL = "https://github.com/mlemgroup/LemmyMarkdownUI";
 			requirement = {
 				kind = upToNextMinorVersion;
-				minimumVersion = 0.2.0;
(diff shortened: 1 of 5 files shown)


EricBAndrews (approved):
Nice! 🚀

PR #1184, opened 2024-08-04
Working post replies

Sjmarf:
Replying to posts is now functional. This necessitated moving the `ExpandedPost+Logic` stuff into a dedicated `ExpandedPostTracker` class. None of that logic has actually changed. 

You can choose which account to reply from using the drop-down at the top. If the targeted post fails to resolve on the new account, a suitable message is shown and the send button is disabled.


Mlem/App/Utility/Extensions/ApiClient+Extensions.swift
```diff
@@ -12,6 +12,8 @@ extension ApiClient {
         if token == nil {
             return AppState.main.guestSession.api === self
         }
-        return !AppState.main.activeSessions.allSatisfy { $0.api !== self }
+        return AppState.main.activeSessions.contains(where: { $0.api === self })
     }
+    
+    var isAuthenticatedAndActive: Bool { isActive && token != nil }
```
EricBAndrews: *nit*

Super useful construct, but the name is a mouthful--maybe `canInteract` is a little more ergonomic?

Mlem/App/Views/Shared/Images/SimpleAvatarView.swift
```diff
@@ -0,0 +1,65 @@
+//
+//  SimpleAvatarView.swift
+//  Mlem
+//
+//  Created by Sjmarf on 15/07/2024.
+//
```
EricBAndrews, line 6: Is this intentionally re-introduced?

Mlem/App/Views/Pages/ResponseComposerView.swift
```diff
@@ -14,19 +14,31 @@ struct ResponseComposerView: View {
...
     
     init?(
-        context: ResponseContext
+        context: ResponseContext,
+        expandedPostTracker: ExpandedPostTracker? = nil
     ) {
-        self.context = context
+        self.originalContext = context
+        self.resolvedContext = context
+        self.expandedPostTracker = expandedPostTracker
```
EricBAndrews: This line fails the build on Xcode 15

```suggestion
        self.originalContext = context
        self._resolvedContext = .init(wrappedValue: context)
        self.expandedPostTracker = expandedPostTracker
```

Mlem/App/Views/Shared/MarkdownTextEditor.swift
```diff
@@ -76,7 +76,7 @@ struct MarkdownTextEditor<Content: View>: UIViewRepresentable {
     }
  
     func updateUIView(_ textView: UITextView, context: Context) {
-        textView.text = text
+        text = textView.text
```
EricBAndrews: Not sure if it's this line specifically that's responsible, but I get the following concurrency error when opening the reply sheet, flagged on L11:

`Modifying state during view update, this will cause undefined behavior.`

Mlem/App/Views/Pages/ResponseComposerView.swift
```diff
@@ -94,15 +121,79 @@ struct ResponseComposerView: View {
...
+                resolutionState = .notFound
+            }
+        } catch {
+            Task { @MainActor in
+                resolutionState = .error(.init(error: error))
+            }
+        }
+    }
+    
+    func send() async {
```
EricBAndrews, line 175: In my testing, I found the reply was always sent from the account that was active when the sheet was first opened.

EricBAndrews (changes requested):

Mlem/App/Views/Shared/Images/SimpleAvatarView.swift
Sjmarf: Yes - `.clipShape` doesn't work in for images a context menu; this view crops the image itself rather than using `.clipShape`.

Mlem/App/Utility/Extensions/ApiClient+Extensions.swift
Sjmarf: Sure 👍

Mlem/App/Views/Shared/MarkdownTextEditor.swift
Sjmarf: Fixed 👍

Mlem/App/Views/Pages/ResponseComposerView.swift
Sjmarf: Hmm, I can't reproduce this behavior on iOS 17.5 or iOS 18. To be clear, changing the account from the drop-down at the top doesn't change the global "active" account - it just chooses which account that reply is posted from. Which iOS version are you using?

EricBAndrews: iPhone Xs iOS 17.0 simulator, built against Xcode 15.4. Also tested with accounts on real instances and differing instances:


https://github.com/user-attachments/assets/f184bed7-d173-49e5-875c-0b599648028e

Sjmarf: Thanks! I was able to reproduce this on a local instance and I think I've fixed it.

EricBAndrews (approved):
LGTM! 🔥 

The user switching UI is really clean!

PR #1191, opened 2024-08-06
Fix missing app icon images

### LINKED ISSUE #1178: [2.0] App Icon images not shown in app icon settings

Sjmarf:
Fixes #1178. This is done by adding duplicate images (with a smaller resolution), as per beta 5 release notes:

> If UIImage(named:) was called with the name of an app icon, it would previously return an undefined image of the icon set, without any masking or shadowing applied. On iOS 18, it now returns nil. To display an icon within the app itself, add the desired image to the asset catalog as an image set.


EricBAndrews (approved):
Nice! 🚀 

There are a couple build issues but they should be fixed by the dev2 merge.

PR #1192, opened 2024-08-07
Fix Nightly Build

EricBAndrews:
For ✨ Xcode Reasons✨, the lines modified here cause the build to _sort of_ fail. A valid archive is still produced, but because of the "failure," it is not automatically pushed out to the internal beta.

Moving the offending code into its own function appears to appease the compiler gremlins.

Screenshot 2024-08-06 at 10 19 41 PM


Sjmarf (approved):
LGTM 👍

PR #1195, opened 2024-08-07
Fix localhost account decoding

### LINKED ISSUE #1194: [2.0] adding a localhost account causes loss of all saved accounts

EricBAndrews:
Adds logic to ensure localhost accounts are always encoded and decoded to the same values.

Closes #1194

### DIFF
--- a/Mlem/Localizable.xcstrings
+++ b/Mlem/Localizable.xcstrings
@@ -612,12 +612,6 @@
     "Temporary" : {
 
     },
-    "Test" : {
-
-    },
-    "Test Long" : {
-
-    },
     "Theme" : {
 
     },
--- a/Mlem/Localizable.xcstrings
+++ b/Mlem/Localizable.xcstrings
@@ -459,6 +459,9 @@
     "Quote" : {
 
     },
+    "Really super long text" : {
+
+    },
     "Received %@" : {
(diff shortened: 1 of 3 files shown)


Mlem/Localizable.xcstrings
```diff
@@ -458,6 +458,9 @@
     },
     "Quote" : {
 
+    },
+    "Really super long text" : {
+
```
Sjmarf: All of these are from a `#Preview`. From Xcode 16 onwards the compiler excludes `#Preview` strings; until then we could either not include them in the commit or use `Text(verbatim: )` to avoid them being parsed as localized strings
```suggestion
```

Mlem/Localizable.xcstrings
```diff
@@ -611,6 +614,12 @@
     },
     "Temporary" : {
 
+    },
+    "Test" : {
+
+    },
+    "Test Long" : {
+
```
```suggestion
```

Mlem/Localizable.xcstrings
```diff
@@ -670,6 +679,9 @@
           }
         }
       }
+    },
+    "Unfavorited Community" : {
+
```
```suggestion
```

Sjmarf (approved):
Nice! This was bugging me earlier :D

PR #1197, opened 2024-08-07
Subscription List Efficiency

Sjmarf:
This fixes a lag spike that occurred when navigating between pages in a `NavigationStack` if your account has a lot of subscriptions. 

`NavigationLayerView` receives a view update when navigating back/forward because of the navigation path changing. There isn't a way to avoid this as far as I can tell (apart from maybe with UIKit). The root view of `NavigationLayerView`, which is `SubscriptionListView`, then gets the view update too. `SubscriptionListView` contains a `List` of items. Unlike `LazyVStack`, `List` instantiates _all_ of it's children when it's initialised. This involves calling `menuActions` for *every* community in the list, which turned out to be the cause of the performance issue. 

I've fixed this by instead passing a closure `{ community.menuActions() }` to `.contextMenu`, which is called only when the context menu is actually opened. I've done the same with `EllipsisMenu` for consistency (and whatever slight performance benefit that may have).

I also rearranged `SubscriptionListView` a little.


Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListItemView.swift
```diff
@@ -0,0 +1,76 @@
...
+//
+
+import MlemMiddleware
+import SwiftUI
+
+struct SubscriptionListItemView: View {
+    @Environment(NavigationLayer.self) private var navigation
+
+    @AppStorage("subscriptions.sort") private var sort: SubscriptionListSort = .alphabetical
+    @AppStorage("subscriptions.instanceLocation")
```
EricBAndrews: *nit*

Update to use `@Setting` after merging dev2

Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListItemView.swift
```diff
@@ -0,0 +1,76 @@
...
+    let section: SubscriptionListSection
+    let sectionIndicesShown: Bool
+    
+    var body: some View {
+        NavigationLink(.community(community)) {
+            HStack(spacing: 15, content: label)
+        }
+        .contextMenu { community.menuActions(feedback: [.toast], navigation: navigation) }
+        .swipeActions(edge: .trailing) {
+            Button("Unsubscribe", systemImage: "xmark") {
```
EricBAndrews: *nit* Should be defined in `Icons`

EricBAndrews (approved):
Nice! 👍

PR #1204, opened 2024-08-11
Image resolution

### LINKED ISSUE #1203: Fetch images at lower resolution
Certain images (such as c/pics) have super high-res images. We could fetch these images at a lower resolution until the user opens the image viewer, then seamlessly fetch and replace the full-res version. I think there's more lag spikes on c/pics than other communities, so this should hopefully help performance too.

Sjmarf:
Images are now fetched at a lower resolution. This mitigates lag caused by the feed having super high-res images (such as in c/pics).

Images are fetched at 1024px, for all post sizes. When you open the image viewer, the full-resolution version is fetched. 

For thumbnails in compact/headline mode, the 1024px version is fetched and then resized down to 256px. I can't tell whether resizing has a performance benefit here. Super-large images were definitely causing lag in c/pics, so in theory it makes a difference?

The reason why I'm not fetching the images at 256px directly in compact mode is that a higher-resolution version would have to be fetched when you open the post, which introduces a loading time.

Closes #1203


Mlem.xcodeproj/project.pbxproj
```diff
@@ -2308,8 +2316,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware.git";
 			requirement = {
-				kind = upToNextMinorVersion;
-				minimumVersion = 0.23.0;
+				branch = sjmarf/prefetching;
+				kind = branch;
```

EricBAndrews (changes requested):
Code looks good! I've noticed a couple of things:

1. The image viewer is _significantly_ slower to appear on large images (~.5s to the beginning of the animation, ~1s to fully open compared to previous ~.25s/.75s). You can see the difference side-by-side:

https://github.com/user-attachments/assets/eef13c54-7006-41ab-af1b-6a9e53709828 

https://github.com/user-attachments/assets/7b60f46c-bbcb-4f7e-ae09-6600350b794e

It looks like the issue is that SwiftUI wants to render the `DynamicImageView` before the sheet starts opening, which takes a moment on big images. I was able to replicate the original performance by using this code:

```
    @State var appeared: Bool = false

    ...

        if appeared {
            ZoomableContainer {
                DynamicImageView(url: url)
                    .padding(Constants.main.standardSpacing)
            }
            .toolbar {
                ToolbarItem(placement: .topBarTrailing) {
                    CloseButtonView()
                }
            }
        } else {
            ProgressView()
                .onAppear {
                    DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
                        appeared = true
                    }
                }
        }
```

For smaller images, the image appears as the sheet is coming up, and the progress view is never shown to the user.

2. Thumbnail images sometimes don't pick up loading state change until they disappear and reappear (tested on both Compact and Headline post size, larger sizes do not seem to be affected):

https://github.com/user-attachments/assets/48dd9eee-86f8-423d-ad11-8f8effaef5d9

Sjmarf:
> The image viewer is significantly slower to appear on large images

This isn't new behavior I don't think - I can also reproduce this on super-large images on dev2:

https://github.com/user-attachments/assets/684af395-5918-4936-a3a1-6134ec8e3a34

I don't _think_ it's any worse in this PR, and I can't find a (good) way to fix it at the moment. Are you happy for me to leave this as OOS? I'll open an issue for it.

> Thumbnail images sometimes don't pick up loading state change until they disappear and reappear

Fixed (I think) 👍

Mlem/App/Utility/Extensions/Content Models/DeletableProviding+Extensions.swift
```diff
@@ -12,7 +12,7 @@ extension DeletableProviding {
         if feedback.contains(.toast), !deleted {
             let task = toggleDeleted()
             Task {
-                let result = try await task.result.get()
+                let result = await task.result.get()
```
EricBAndrews: This change causes build failure on Xcode 15.4--I'm not entirely sure why it built for me last review, but I can't get it to build now even on a totally clean clone

```suggestion
                let result = try await task.result.get()
```

EricBAndrews (approved):
I'm content to leave image viewer appearance time OOS for this PR.

One small build failure--trivial enough not to be worth a re-review cycle.

PR #1209, opened 2024-08-12
Misc Feed QoL

### LINKED ISSUE #1148: [2.0] "Show hidden posts" toggle in feed
Related: #1147

EricBAndrews:
# Checklist
      - closes #1148
      - progress towards #1117, #938

# Pull Request Information

This PR contains a collection of miscellaneous minor feed improvements:

- Added hide/show read and post size options to toolbar menu
- Lightly restructured `PostGridView` responsibilities
  - To handle populating standard post grid toolbar items, `PostGridView` is responsible for displaying the toolbar; it takes in a list of actions to display in addition to the standard options.
  - `PostGridView` now handles all loading, including initial loading (I wanted to do the same to `PersonContentGridView`, but there are some middleware bugs to sort out first)
- Added individual All/Local/Subscribed/Saved navigation to sidebar view
- Removed some warnings
- Read posts are no longer hidden at the UI level


Mlem/App/Views/Root/Tabs/Feeds/PostGridView+Logic.swift
```diff
@@ -0,0 +1,55 @@
...
+                icon: size.icon(filled: postSize == size),
+                enabled: true
+            ) {
+                postSize = size
+            })
+        }
+        
+        return ret
+    }
+}
```
Sjmarf: I'm not sure we should be using actions when we only need them in one context. This is a lot of extra code that could be avoided by using a simple SwiftUI picker in the toolbar:

```swift
Picker("Post Size", systemImage: Icons.postSize, selection: $postSize) {
    ForEach(PostSize.allCases, id: \.self) { item in
        Label(item.label, systemImage: item.systemImage)
    }
}
```

(The same applies for the Hide/Show read button)

Another advantage of using a Picker over multiple buttons is that it's got better accessibility support. With multiple buttons, there's no way for VO to know whether a button is "on" or not. I also think it's a bit clearer than using a symbol variant to indicate selection. What do you think?

Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListView.swift
```diff
@@ -12,14 +12,15 @@ struct SubscriptionListView: View {
     @Environment(AppState.self) private var appState
     @Environment(NavigationLayer.self) private var navigation
     @Environment(TabReselectTracker.self) var tabReselectTracker
+    @Environment(Palette.self) var palette
     
     @Setting(\.subscriptionSort) private var sort
     
     @State var noDetail: Bool = false
+    @State var feedOptions: [FeedSelection] = FeedSelection.guestCases
```
Sjmarf: Does this need to be an `@State` value? I reckon it's reasonable enough to use a computed property instead, unless there's a reason we need the `@State`

Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListView.swift
```diff
@@ -28,6 +29,13 @@ struct SubscriptionListView: View {
         })
         .navigationTitle("Feeds")
         .navigationBarTitleDisplayMode(.inline)
+        .onChange(of: appState.firstApi, initial: true) {
+            if appState.firstApi.canInteract, appState.firstAccount is UserAccount {
```
Sjmarf: `appState.firstAccount is UserAccount` is sufficient here
```suggestion
            if appState.firstAccount is UserAccount {
```

Mlem/App/Views/Shared/PostGridView.swift
```diff
@@ -9,27 +9,52 @@ import Foundation
...
     @Setting(\.showReadInFeed) var showRead
     
     @Environment(AppState.self) var appState
+    @Environment(Palette.self) var palette
     
     @State var columns: [GridItem] = [GridItem(.flexible())]
     @State var frameWidth: CGFloat = .zero
     
     let postFeedLoader: CorePostFeedLoader
+    let actions: [any Action]?
```
Sjmarf: Maybe this should be non-optional with a default value of `[]`?

Sjmarf (changes requested):
Looks good! A few thoughts

Mlem/App/Views/Shared/PostGridView.swift
```diff
@@ -38,10 +63,23 @@ struct PostGridView: View {
...
+                        }
+                    }
+                    
+                    Divider()
+                    
+                    ForEach(standardActions, id: \.id) { action in
+                        MenuButton(action: action)
                     }
                 }
             }
```
Sjmarf: The community actions aren't shown if you switch tabs to "About" in `CommunityView`. This could be fixed by duplicating the toolbar actions in `CommunityView` to cover those cases.

Or alternatively, using `ToolbarItemGroup(placement: .secondaryAction)` lets you put groups of actions from multiple different views in the same ellipsis menu. There could be a `ToolbarItemGroup` in `PostGridView` for the feed actions, and a `ToolbarItemGroup` in `CommunityView` for the community actions. That also avoids having to pass `[any Action]` to `PostGridView`. 

iirc `Divider` doesn't work when using that system, though the same effect can be achieved using `Section`. The downside of using that system is that the order in which the groups are rendered can't be easily controlled (afaik). I set this up one time in v1 and it was in the right order for our use case

Sjmarf (changes requested):

Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListView.swift
EricBAndrews: That's a good point--I had it as `@State` because that's how it was in `FeedsView`, but it looks like a computed var works perfectly well in both cases

Mlem/App/Views/Shared/PostGridView.swift
EricBAndrews: Nice catch! The `ToolbarItemGroup` seems like the most canonical way to handle it--it sets them up correctly in this case, and if we run into issues down the line we can revisit and develop a more sophisticated solution.

Mlem/App/Views/Root/Tabs/Feeds/PostGridView+Logic.swift
EricBAndrews: Sure, that works for me

Mlem/App/Views/Pages/Community/CommunityView.swift
```diff
@@ -95,12 +95,18 @@ struct CommunityView: View {
             .environment(\.communityContext, community)
         }
         .background(postSize.tiled ? palette.groupedBackground : palette.background)
-        .loadFeed(postFeedLoader)
+        .toolbar {
+            ToolbarItemGroup(placement: .secondaryAction) {
+                ForEach(community.menuActions(navigation: navigation), id: \.id) { action in
+                    MenuButton(action: action)
+                }
```
Sjmarf: *nit* avoid calling `menuActions` until needed

```suggestion
                MenuButtons { community.menuActions(navigation: navigation) }
```

Sjmarf (approved):
LGTM 👍

PR #1221, opened 2024-08-17
fix scroll to top

EricBAndrews:
Fixes tab reselect scroll-to-top not working. I've gone back to using `@State` to track `isAtTop`; I analyzed view updates thoroughly with [`Self._printChanges()` ](https://developer.apple.com/documentation/swift-playgrounds/console-print-debugging#Understand-when-and-why-your-views-change), and both the `@State` and `Model` implementations cause an `@self changed` view update. Adding an overlay that prints the current value of `isAtTop` does invoke a new update which cites `isAtTop` as the driving change; from this I believe it is safe to conclude that the use of an `@State` variable offers no performance reduction as compared to `Model`.


Sjmarf (approved):
LGTM!

PR #1222, opened 2024-08-17
Upvote on save

EricBAndrews:
Implements upvote on save

### DIFF
--- a/Mlem/App/Utility/Extensions/Content Models/Interactable1Providing+Extensions.swift
+++ b/Mlem/App/Utility/Extensions/Content Models/Interactable1Providing+Extensions.swift
@@ -25,7 +25,7 @@ extension Interactable1Providing {
         if let self = self as? any Comment2Providing { return .comment(self) }
         return nil
     }
-
+    
     func toggleUpvoted(feedback: Set<FeedbackType>) {
         if let self2 {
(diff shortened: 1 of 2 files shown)


Mlem/App/Utility/Extensions/Content Models/Interactable1Providing+Extensions.swift
```diff
@@ -56,6 +56,12 @@ extension Interactable1Providing {
                 HapticManager.main.play(haptic: .success, priority: .low)
             }
             self2.toggleSaved()
+            
+            @Setting(\.upvoteOnSave) var upvoteOnSave
+            if upvoteOnSave, self2.votes.myVote != .upvote {
+                self2.updateVote(.upvote)
+            }
+
```
Sjmarf: Need to check whether we're saving or un-saving otherwise the post is upvoted on unsave

```suggestion            
            @Setting(\.upvoteOnSave) var upvoteOnSave
            if upvoteOnSave, !self2.saved {
                self2.updateVote(.upvote)
            }
            
            self2.toggleSaved()     
```

Sjmarf (changes requested):

EricBAndrews:
nice catch!

Sjmarf (approved):

Issue #1220, opened 2024-08-17
[2.0] Feed page loading doesn't always trigger when bottom of page reached


Sjmarf:

EricBAndrews:
Closed in #1223

PR #1223, opened 2024-08-17
Fix load feed cancellation error

EricBAndrews:
Fixed an issue where `loadFeed` was being cancelled, leaving the feed loader in a permanent `.loading` state. The work is now performed on an independent thread that is not cancelled when the parent view is derendered.


Sjmarf (approved):
Nice! Closes #1220

PR #1232, opened 2024-08-19
Widget Customization

Sjmarf:
Added widget customization. I think putting readout settings in the same page as the widgets themselves is better, what do you think?

I included an "Apply to all interaction bars" button at the bottom of the screen, because most of the time people will probably want identical setups in all contexts.

Sorry for the line count; most of it is pretty repetitive


Mlem/App/Configuration/User Settings/SettingPropertyWrapper.swift
```diff
@@ -11,11 +11,11 @@ import SwiftUI
 
 @propertyWrapper
 struct Setting<T>: DynamicProperty {
-    @ObservedObject private var defaults: Settings
+    @StateObject private var defaults: Settings
```
EricBAndrews: This throws the following purple warning:

`Accessing StateObject's object without being installed on a View. This will create a new instance each time.`

Since the only non-`AppStorage` values we need in `Settings` are interaction bar configurations, and those are never referenced outside of a `View` context, it is probably cleaner to handle them with a dedicated `InteractionBarConfigurationTracker` provided using `.environment` and let `Settings` remain a pure `@AppStorage` wrapper.

EricBAndrews (changes requested):
Impressive work! This is even better than the v1 customizer 😁 

I've left a couple of implementation notes below. On the design side, the only suggestion I have is to disable info stack combined score when individual scores are selected and vice versa. It's redundant to have both activated and potentially confusing, since we don't (and shouldn't) use a different icon for combined and separate:
Screenshot 2024-08-21 at 1 08 14 PM

Moreover, it very easily fixes this overcrowding behavior:
Screenshot 2024-08-21 at 1 09 42 PM

Mlem/App/Configuration/User Settings/SettingPropertyWrapper.swift
Sjmarf: Nice catch 👍

Sjmarf:
Good idea, done 👍

Mlem/App/Views/Root/Tabs/Settings/InteractionBarEditor/InteractionBarEditorView+Logic.swift
```diff
@@ -0,0 +1,106 @@
...
+            switch hoveredDropLocation {
+            case let .bar(index: dropIndex):
+                items.insert(trayPickedUpItem, at: dropIndex)
+            default: break
+            }
+        }
+    }
+    
+    func trayItemOutlineColor(_ item: Configuration.Item) -> Color {
+        if let barPickedUpIndex, items[barPickedUpIndex] == item, hoveredDropLocation == .tray {
```
EricBAndrews: This throws an IndexOutOfRange if you try to remove the rightmost widget

Mlem/App/Configuration/User Settings/Settings.swift
```diff
@@ -44,9 +43,11 @@ class Settings: ObservableObject {
...
     
     @AppStorage("subscriptions.instanceLocation") var subscriptionInstanceLocation: InstanceLocation = UIDevice.isPad ? .bottom : .trailing
+    
     @AppStorage("subscriptions.sort") var subscriptionSort: SubscriptionListSort = .alphabetical
     
     @AppStorage("person.showAvatar") var showPersonAvatar: Bool = true
     
     @AppStorage("community.showAvatar") var showCommunityAvatar: Bool = true
 }
+
```
EricBAndrews: *nit*

SwiftLint complains about an extra newline here

EricBAndrews (changes requested):

Mlem/App/Views/Root/Tabs/Settings/InteractionBarEditor/InteractionBarEditorView+Logic.swift
Sjmarf: Fixed 👍

EricBAndrews (approved):
LGTM! 🚀

PR #1234, opened 2024-08-20
Reply to inbox replies

Sjmarf:


EricBAndrews (changes requested):
Looks good 👍 

Like other interactions, replying should mark the inbox item as read--this should also fix an issue I noticed where the "outdated inbox popup" appears after replying due to the inbox item being marked read server-side.

Sjmarf:
Good catch - I've implemented this in [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/49).

EricBAndrews (approved):
LGTM! 💯

PR #1239, opened 2024-08-21
Comment Sorting

### LINKED ISSUE #1215: [2.0] Comment Sorting
See also #804

Sjmarf:
Added comment sorting. You can set a default sort in settings. The "Controversial" sort mode is version locked, but I don't think it's worth implementing a fallback system this time because it's unlikely anyone will set "Controversial" as the default *and* use multiple accounts, one of which is on <0.19.0.

Closes #1215


EricBAndrews (approved):
LGTM 👍 

The only instance with any significant number of users still running 0.18x is Beehaw--it's _possible_ there are going to be some users there who also maintain a more broadly federated Lemmy account, but I'm content to solve that problem if it shows up.

PR #1240, opened 2024-08-21
Account Switch bug fixes

Sjmarf:
Fixed bugs when "Reload on Switch" is off:
- The refresh popup is now hidden when viewing non-feed tabs of a `CommunityView`
- Fixed `PersonView` not always loading the new feed when "refresh" tapped
- Fixed `ExpandedPostView` not correctly highlighting a comment when the view was opened using a `Post` with a non-active `ApiClient`.

Also fixed a bug where the quick switcher would remain open after logging into a new account when "Reload on Switch" is on.

### DIFF
--- a/Mlem/App/Views/Pages/Person/PersonView.swift
+++ b/Mlem/App/Views/Pages/Person/PersonView.swift
@@ -39,7 +39,7 @@ struct PersonView: View {
     init(person: AnyPerson) {
         self._person = .init(wrappedValue: person)
         
-        if let person1 = person.wrappedValue as? any Person1Providing {
+        if let person1 = person.wrappedValue as? any Person1Providing, person1.api === AppState.main.firstApi {
             self._feedLoader = .init(wrappedValue: .init(
                 api: AppState.main.firstApi,
                 userId: person1.id,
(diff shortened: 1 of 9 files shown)


Mlem/App/Views/Pages/Person/PersonView.swift
```diff
@@ -96,8 +97,11 @@ struct PersonView: View {
                             savedOnly: false,
                             prefetchingConfiguration: .forPostSize(postSize)
                         )
-                        
                         preheatFeedLoader()
+                    } else if feedLoader?.api !== entity.api {
+                        Task {
+                            await feedLoader?.switchUser(api: entity.api, userId: entity.id)
+                        }
```
EricBAndrews: *nit*

This prevents spinning up a `Task` if we don't have a `feedLoader`--the compiler might optimize that out, but it can't hurt

```suggestion
                    } else if let feedLoader, feedLoader.api !== entity.api {
                        Task {
                            await feedLoader.switchUser(api: entity.api, userId: entity.id)
                        }
```

EricBAndrews (approved):
LGTM! 🚀 

One minor nit, non-blocking

PR #1241, opened 2024-08-22
Fix "Calling into SwiftUI on non-main thread is not supported" (maybe)

Sjmarf:
This fatal error would _occasionally_ occur on `CustomTabViewHostingController.swift` line 16. It's hard to tell whether it's been fixed or not because the bug is random and somewhat improbable.

My shaky repro for this is to switch back and forth between two accounts with "Reload on Switch" off. Eventually it would occur. I've also had this happen before when switching between tabs. This occurs on both Xcode 15 and 16.


EricBAndrews (approved):
Nice 👍

Issue #1244, opened 2024-08-24
[2.0] Improved Image Handling


EricBAndrews:
Improve how we handle images across the board:

- Use Nuke `ImageProcessors` to perform resizing upstream
- Investigate fetching full-size and resizing to all required resolutions using upstream image requests (might require custom image processor)
- Investigate lazy loading if above results in excessive network usage
- Refactor `DynamicImageView` to use `LazyImage` (we don't need the  `Image(uiImage)` construction because `DynamicImageView` never appears in animation/problem views)
- Use full-size image for `.saveImage`

Sjmarf:
Is this done?

EricBAndrews:
Yes, this is either all closed or migrated into smaller issues

PR #1245, opened 2024-08-24
Save Image

### LINKED ISSUE #1136: [2.0] Image actions ("Save", "Copy", etc)

EricBAndrews:
# Checklist
      - closes #1136 

# Pull Request Information

Adds necessary image handling for 2.0 initial TF. I have some further improvements planned in #1244, but I predict they will be slow to develop.

- Adds image saving
- Adds image sharing
- Adds QuickLook view

This implementation uses the stock SwiftUI `quickLookPreview` rather than the custom one implemented in v1. The stock implementation is much more performant at the cost of having no interactive dismissal; since quicklook is not our primary image viewer, I consider that trade (and the slimmer codebase) worthwhile.

### DIFF
--- a/Mlem/App/Views/Shared/Images/Core/DynamicImageView.swift
+++ b/Mlem/App/Views/Shared/Images/Core/DynamicImageView.swift
@@ -15,7 +15,6 @@ extension UIImage {
 
 struct DynamicImageView: View {
     @Environment(Palette.self) var palette: Palette
-    @Environment(NavigationLayer.self) private var navigation
     
     @State var loader: ImageLoader
     @State var loadingPref: ImageLoadingState?
--- a/Mlem/App/Views/Shared/Images/Core/DynamicImageView.swift
+++ b/Mlem/App/Views/Shared/Images/Core/DynamicImageView.swift
@@ -6,6 +6,7 @@
 //
 
 import Nuke
+import QuickLook
 import SwiftUI
 
 extension UIImage {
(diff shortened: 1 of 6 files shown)


Mlem/App/Views/Shared/Images/Core/DynamicImageView.swift
```diff
@@ -56,5 +57,58 @@ struct DynamicImageView: View {
...
+                        }
+                    }
+                }
+            }
+            .quickLookPreview($quickLookUrl)
+    }
+    
+    func saveImage() async {
+        do {
+            let (data, _) = try await ImagePipeline.shared.data(for: .init(url: loader.url))
```
Sjmarf: `loader.url` can include the `thumbnail=1024` query item, which needs to be excluded somehow. iirc query items can't be removed from a `URL` directly - one option would be to convert to `URLComponents`, remove the query item then convert back again

Sjmarf (changes requested):
Nice! One note on image quality

In v1.0 we also had these context menu actions on Compact/Headline mode images I believe. Maybe OOS

Mlem/App/Views/Shared/Images/Core/DynamicImageView.swift
EricBAndrews: Good catch--I've implemented a fix that saves and opens quickLook at higher resolutions. Share is a tricky case, because it needs the image to share to be readily available--that should be easy to handle with the improved image fetching in #1244. In the meantime I don't think anybody will actually notice, so I've left OOS and added a note to 1244.

EricBAndrews:
Re: context menu on thumbnails, I've opened #1247 to handle.

Mlem/App/Views/Shared/Images/Core/DynamicImageView.swift
Sjmarf: I added a way to do this programatically a while ago: `navigation.shareUrl = url`. It would need adapting to accept images too, though - maybe with an enum?

Sjmarf (approved):
Nice!

PR #1260, opened 2024-08-27
`PersistenceRepository`: Don't show error toast if file not found

Sjmarf:
Without this, an error toast will be shown the first time the user opens a post after opening the app if they're using the default interaction bar.


EricBAndrews (approved):
Nice 💯

PR #1262, opened 2024-08-27
Post Editor (initial work)

Sjmarf:
> [!note]
> Drafting until first TF out. 

You can change the community and/or account from inside the sheet. Tapping the "Crosspost" option lets you add multiple targets. I also added a "New Post" button to the profile.

You can't add links or images yet. I'll implement those in upcoming PRs.


EricBAndrews (changes requested):
Nice! This is a very smooth UX.

A couple thoughts:
- "Crosspost" should automatically open the community picker
- "No user" image should be the default person avatar
Screenshot 2024-08-29 at 5 11 01 PM

Sjmarf:
> "No user" image should be the default person avatar

Nice catch!

> "Crosspost" should automatically open the community picker

Yeah, I considered this. Doing so is nicer UX in theory, but unfortunately it doesn't cover the case where you're trying to crosspost to a community that isn't available from the active account. With the current system, you can do this by changing the crosspost account to the desired one _before_ choosing a community - the community picker will use the `api` of the account you choose when displaying results. If the community picker opened immediately, you wouldn't get an opportunity to do this.

I _could_ make it open the sheet immediately if you've only logged into Mlem with one account? Or is it better to be consistent and not show the sheet immediately in any circumstance?

EricBAndrews:
Hmm, that's an interesting edge case. I think we should always open the sheet on the basic principle of "make the common case fast"--I can't imagine it's going to be super common for a user to be crossposting to a community that isn't accessible from the original posting account. It's also unlikely that users will have a clear idea of the precise federation boundaries--even with the current UX, they may well open the community picker, see the community they want isn't accessible, and then have to close it anyway to switch accounts.

The ideal solution might be to have account selection integrated into the community picker, though I'm content to call that well out of scope for this PR.

Sjmarf:
Sure. I could have it open the sheet immediately, but if the sheet is dismissed we would leave the second crosspost option there? That way the user could close the sheet and switch accounts if needed

EricBAndrews (approved):
LGTM! 🚀 

> Sure. I could have it open the sheet immediately, but if the sheet is dismissed we would leave the second crosspost option there? That way the user could close the sheet and switch accounts if needed

Yep, that makes sense to me 👍

PR #1266, opened 2024-08-28
Default Feed

EricBAndrews:
Adds the "default feed" option. `FeedsView` intended behavior is now as follows:

- By default, use `defaultFeed` setting; if `feedSelection` provided, override with that value
- Fallback to `.local` if above logic would initialize with an authenticated account and the current account is not authenticated

The `.feeds` navigation page now takes an optional parameter to support the above logic.


Sjmarf (approved):
LGTM!

PR #1269, opened 2024-08-28
Eric/mark read on scroll

EricBAndrews:
# Checklist

# Pull Request Information

Depends on [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/53)

Implements mark read on scroll.

Most of the work is done on the middleware side--this pretty much just adds a settings toggle and a view modifier to call the middleware as appropriate. 

Flushing the read queue is now a frontend responsibility, and happens in two places:
- When the active account is changed
- When the app moves into the background


Mlem/App/Utility/Extensions/Views/View Modifiers/MarkReadOnScroll.swift
```diff
@@ -0,0 +1,46 @@
...
+                    }
+                } catch {
+                    handleError(error)
+                }
+            }
+            .onDisappear {
+                if markReadOnScroll {
+                    post.markReadIfStaged()
+                }
+            }
```
Sjmarf: I wonder if there's a way to avoid storing `readStaged` on every post - we should try to keep UI logic out of MlemMiddleware if possible. Perhaps `PostGridView` could store a `bottomAppearedPostIndex: Int` value, which is passed as a binding to this `MarkReadOnScroll` modifier? Then the above logic could look something like:

```swift
.task {
    if markReadOnScroll, try await post.api.batchMarkReadEnabled {
        bottomAppearedPostIndex = max(index, bottomAppearedPostIndex)
    }
}
.onDisappear {
    if markReadOnScroll, index <= (bottomAppearedPostIndex - postSize.markReadOffset) {
        post.markRead()
    }
}
```

It should be possible to avoid touching MlemMiddleware at all by using something like that.

Storing `readStaged` in `Post2` may also have undesired consequences if you happen to have a post displayed in two feeds in Mlem at once. Though that isn't too big of an issue, it would be nice to avoid it.

I may have overlooked a reason why that might not work. I'm happy to leave this as OOS for now if we want to prioritise shipping a working version ASAP

Sjmarf: > a bottomAppearedPostIndex: Int value, which is passed as a binding to this MarkReadOnScroll modifier?

Or instead of a binding, a wrapper class to avoid causing a view update

EricBAndrews: Yeah, that's much cleaner--the logic is a bit ugly but it's better than putting it in middleware.

Mlem.xcodeproj/project.pbxproj
```diff
@@ -2534,20 +2538,20 @@
...
-		CD4368BF2AE23FD400BD8BD1 /* XCRemoteSwiftPackageReference "Semaphore" */ = {
+		CD1B2E232C7FBBB70075C7EA /* XCRemoteSwiftPackageReference "MlemMiddleware" */ = {
 			isa = XCRemoteSwiftPackageReference;
-			repositoryURL = "https://github.com/groue/Semaphore";
+			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware.git";
 			requirement = {
-				kind = upToNextMajorVersion;
-				minimumVersion = 0.0.8;
+				branch = "eric/mark-read-on-scroll";
+				kind = branch;
```

Sjmarf (approved):
Nice! Do we want to flush the queue on a timer too, or not? Currently we only flush when the user opens the app switcher or switches accounts. In v1 I believe we did it on a timer and when a feed is closed. This means that if the user switches to desktop without closing mlem or switching accounts, their read posts will be mostly up-to-date.

PR #1283, opened 2024-08-30
Custom Dynamic Themes

### LINKED ISSUE #252: Themes

EricBAndrews:
# Checklist
      - closes #252 

# Pull Request Information

Implements the Solarized color themes. The main addition is a couple of convenience `Color` initializers for custom dynamic colors. Most of the lines are missing `palette` calls to make the app fully adhere to the new palette.

The following views have been overridden to respect `palette`:
- `Divider`
- `Form`
- `Section`

`SwiftUI.<View>` is used where `palette` is inappropriate.

The way in which palettes are previewed in settings has also changed to use the palette's `.accent` and `.background` colors, rather than a custom defined pair of colors.


Mlem/App/Views/Shared/Palette Components/PaletteDivider.swift
```diff
@@ -0,0 +1,20 @@
...
+import SwiftUI
+
+/// Divider() that colors itself appropriately to the palette.
+/// DO NOT use in UIKit environments!
+struct PaletteDivider: View {
+    @Environment(Palette.self) var palette
+    
+    var body: some View {
+        Divider()
+            .overlay(palette.neutralAccent)
```
Sjmarf: In standard theme dark mode this is way brighter than it was before. What about using a `Rectangle()` with a `palette.tertiary` fill?

Mlem/App/Views/Shared/Palette Components/PaletteDivider.swift
```diff
@@ -0,0 +1,20 @@
...
+//
+//  Created by Eric Andrews on 2024-08-30.
+//
+
+import Foundation
+import SwiftUI
+
+/// Divider() that colors itself appropriately to the palette.
+/// DO NOT use in UIKit environments!
+struct PaletteDivider: View {
```
Sjmarf: Do we want to name this `Divider` and override SwiftUI completely? We can always fallback to `SwiftUI.Divider` if we need to use the default for some reason. The same applies to `PaletteForm`

Mlem/App/Views/Shared/MarkdownEditorToolbarView.swift
Sjmarf, line 1: Can we keep this view unthemed? We can't change the theme of the keyboard, so I think the lesser of two evils is to make it match the keyboard rather than the app.

Sjmarf (changes requested):
Nice! Those `Color` extensions are super clean :D

Mlem/App/Views/Shared/MarkdownEditorToolbarView.swift
EricBAndrews: Sure, that's a good point 👍

Mlem/App/Views/Shared/Palette Components/PaletteForm.swift
```diff
@@ -0,0 +1,26 @@
...
+/// Identical to Form, but respects Palette
+struct PaletteForm<Content: View>: View {
+    @Environment(Palette.self) var palette
+    
+    @ViewBuilder let content: () -> Content
+    
+    var body: some View {
+        Form {
+            content()
+                .foregroundStyle(palette.primary)
```
Sjmarf: Doing this means that form headers and footers are no longer `.secondary`. Do we just accept that consequence, implement a custom `Section` too, or add `.foregroundStyle` to all headers and footers?

Mlem/App/Views/Shared/Palette Components/PaletteDivider.swift
EricBAndrews: Nice catch! Just using `palette.tertiary` instead of `.neutralAccent` seems to do the trick, at least to my eye

Mlem/App/Views/Shared/Palette Components/PaletteDivider.swift
EricBAndrews: Yeah, that seems cleaner.

Mlem/App/Views/Shared/Palette Components/PaletteForm.swift
EricBAndrews: Nice catch! In one iteration I actually already had a custom `Section`, I can easily dig that back up

Mlem/App/Views/Shared/Palette Components/PaletteDivider.swift
Sjmarf: Hmm, it still looks a little too bright to me. Can we go darker again?

This PR:

IMG_0868

Before:

Screenshot 2024-08-31 at 22 51 27

Sjmarf: Since `.tertiary` is semi-transparent it'll be stacking ontop of the divider and producing a more opaque shade - adding `.opacity` or `.hidden` below the `.overlay` might fix that

EricBAndrews: Yeah, that's a good call--Divider also changes between light and dark mode and is by default somewhat transparent. I've gotten a very nice approximation of system default--it shouldn't be noticeably different unless you're comparing screenshots at the pixel level, and even then it's pretty close.

Sjmarf (approved):
Nice! Looks great 👍

PR #1293, opened 2024-09-02
New Account Indicator Tweak

Sjmarf:
Shows hours or minutes rather than days if account is <1d old.

Maybe OOS but do we want the flair to show the account age at time of viewing, or their age when they posted the comment?


EricBAndrews (approved):
LGTM!

Time of viewing makes more sense IMO--it would be strange to show "new account" on old posts

PR #1297, opened 2024-09-04
Scroll-to-comment reliability

Sjmarf:
Rewrote parts of `ExpandedPostView` to make scrolling to a comment more reliable. 

This is partly done by scrolling to the comment 0.1 seconds after the view appears. During that time, a `ProgressView` is shown ontop of the view using a `ZStack`. The view also now checks whether the targeted comment is included in the `GetComments` response, and requests a second page before showing the `ExpandedPostView` if not.

Also includes a little bit of logic for the upcoming comment page (which was originally going to be in this PR, but I thought I'd split it up). Relies on [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/58) to support that logic.


Mlem.xcodeproj/project.pbxproj
```diff
@@ -2635,8 +2643,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware.git";
 			requirement = {
-				kind = upToNextMinorVersion;
-				minimumVersion = 0.34.0;
+				branch = "sjmarf/comment-tweaks";
+				kind = branch;
```

Mlem/App/Views/Shared/ExpandedPost/CommentTreeTracker.swift
```diff
@@ -9,31 +9,83 @@ import MlemMiddleware
...
+                let comment: any Comment
+                let api = root.wrappedValue.api
+                if let ensuredComment = ensuredComment as? any Comment, ensuredComment.api == api {
+                    comment = ensuredComment
+                } else {
+                    print("CommentTreeTracker: Resolving comment...")
+                    comment = try await api.getComment(actorId: ensuredComment.actorId)
+                }
+                let idsToSearch = comment.parentCommentIds + [comment.id]
+                if let parentId = idsToSearch.first(where: { id in !newComments.contains(where: { $0.id == id }) }) {
```
EricBAndrews: *nit*

This line (L74) is slick but slow to read--syntactic complexity obscures purpose. Could use an explanatory comment or refactoring out to a couple lines. The latter is perhaps preferable given that this is an n^2 operation, though since n is like 50 that doesn't matter too much.

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView.swift
```diff
@@ -20,111 +19,114 @@ struct ExpandedPostView: View {
...
+                            if post.api == appState.firstApi, tracker.loadingState == .idle {
+                                post.markRead()
+                                await load(tracker: tracker)
+                            }
+                        }
+                }
+                let showLoadingSymbol = !(
+                    highlightedComment == nil
+                        || (post.wrappedValue is any Post2Providing && scrolledToHighlightedComment)
+                )
```
EricBAndrews: This would be cleaner as a computed var unless there's a reason to put it in the view body that I'm missing

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView.swift
```diff
@@ -20,111 +19,114 @@ struct ExpandedPostView: View {
...
+        self._highlightedComment = .init(wrappedValue: highlightedComment)
+        self._scrolledToHighlightedComment = .init(wrappedValue: highlightedComment == nil)
+    }
     
     var body: some View {
         ContentLoader(model: post) { proxy in
-            if let post = proxy.entity, let tracker {
-                let showLoadingSymbol = showCommentWithActorId == nil || (self.post.isUpgraded && tracker.loadingState != .loading)
+            // Using a `ZStack` here rather than `if`/`else` because there needs to
+            // be a delay between the `content()` appearing and calling `scrollTo`
```
EricBAndrews, line 46: This can be achieved using `.overlay`:

```
            Group {
                if let post = proxy.entity, let tracker {
                    content(post: post, tracker: tracker, isLoading: proxy.isLoading)
                        .externalApiWarning(entity: post, isLoading: proxy.isLoading)
                        .task {
                            if post.api == appState.firstApi, tracker.loadingState == .idle {
                                post.markRead()
                                await load(tracker: tracker)
                            }
                        }
                } else {
                    ProgressView()
                }
            }
            .overlay {
                if showLoadingSymbol {
                    palette.background
                        .ignoresSafeArea()
                        .overlay {
                            ProgressView()
                                .tint(.secondary)
                        }
                        .transition(.opacity)
                }
            }
            .animation(.easeOut(duration: 0.1), value: showLoadingSymbol)
```

EricBAndrews (changes requested):
Looking good! I've left a couple code comments, nothing major.

It also appears that the scroll-to doesn't behave nicely on long comment sections:
https://github.com/user-attachments/assets/03a5b1b3-d325-48a8-b2aa-2c39ed9acd17

Sjmarf:
> It also appears that the scroll-to doesn't behave nicely on long comment sections:

This appears to be caused by the presence a super-long post/comment body (one that covers the entire screen). The behavior seen in that post can be reproduced for any post by replacing `ExpandedPostView.swift:95` with a long `Text`:

```diff
- LargePostView(post: post, isExpanded: true)
+ Text(verbatim: .init(repeating: "A", count: 10000))
```

I did some experimenting but couldn't find a good way to fix this 😔 It's possibly a SwiftUI bug. iOS 18 implements a new scrolling system, `.scrollPosition`, which I'm hoping might be used to fix this.

EricBAndrews (approved):
LGTM! 🚀

PR #1307, opened 2024-09-07
Sign Up

### LINKED ISSUE #1305: [2.0] Account Creation ("Sign Up")

Sjmarf:
Sign up to an instance from the "Add Account" button in the account switcher or from an `InstanceView` ellipsis menu.

When you sign up from the account switcher, it shows the basic instance picker. This is a placeholder - at some point I'd like to make this more comprehensive. Some people might want to join an interest-based instance, for example, so we should make those easily discoverable. I'd also like to discourage "just joining the largest instance" somehow; maybe by suggesting a curated list of smaller well-run instances as good general-purpose instances. `awesome-lemmy-instances` has such a list of recommended instances that we could use, though I haven't properly looked into how good that list is or whether it's kept up-to-date. We could also suggest instances that speak the same language that the user's phone is set to (e.g. put jlai.lu at the top if the user speaks French).

When you submit the sign up form, it will either:
- Log you in immediately, if the instance has no sign-up restrictions
- If the instance does not require applications but does require email registration, it will show an info screen to that effect and attempt to login every 5 seconds. When the user verifies their account using the email, it logs them in.
- If the instance requires application, it shows an info page with a "Dismiss" button.

The "Sign up" buttons we discussed putting all over the place aren't present yet.


Mlem/App/Views/Root/Onboarding/SignUpView+Logic.swift
```diff
@@ -0,0 +1,79 @@
...
+//  Created by Sjmarf on 06/09/2024.
+//
+
+import MlemMiddleware
+import SwiftUI
+
+extension SignUpView {
+    var canSubmit: Bool {
+        guard let instance = instance as? any Instance2Providing else { return false }
+        return (!instance.captchaEnabled || !captchaAnswer.isEmpty)
```
EricBAndrews: *micro-nit*

This might just be personal preference/my love of DeMorgan's laws, but I think the logic is a little clearer structured this way? It's pretty clear what the check is doing either way.

```suggestion
        return !(instance.captchaEnabled && captchaAnswer.isEmpty)
```

Mlem/App/Views/Shared/Navigation/NavigationPage.swift
```diff
@@ -55,18 +56,68 @@ enum NavigationPage: Hashable {
         Self.instance(.init(wrappedValue: instance))
     }
     
-    static func communityPicker(api: ApiClient? = nil, callback: @escaping (Community2) -> Void) -> NavigationPage {
+    static func instance(hostOf entity: any ActorIdentifiable) -> NavigationPage {
+        var instance: any InstanceStubProviding = InstanceStub(
+            api: AppState.main.firstApi, actorId: entity.actorId.removingPathComponents()
+        )
+        if let entity = entity as? any Person3Providing {
+            print("INS", entity.instance)
```
EricBAndrews: *nit*

Dead code?

EricBAndrews (approved):
LGTM! I've left a couple tiny nits. It may also be nice to have a dev-mode localhost shortcut built into the planned instance picker--I had to do perform some ungainly hacks to submit applications to localhost.

Sjmarf:
> It may also be nice to have a dev-mode localhost shortcut built into the planned instance picker--I had to do perform some ungainly hacks to submit applications to localhost.

I was doing this by tapping the instance link from the profile page, then using "Sign Up" in the ellipsis menu. Does that not work for you? In theory `nil` can be returned for a person instance, which would prevent that from working, but I'm yet to encounter a case where it returns `nil` yet.

Mlem/App/Views/Root/Onboarding/SignUpView+Logic.swift
Sjmarf: Sure 👍

PR #1312, opened 2024-09-08
Codable Settings

### LINKED ISSUE #1268: Import/Export Settings
Add a feature to save settings to a JSON file. This can also support migrating settings from v1 to v2.

### LINKED ISSUE #1298: [2.0] Buttons in form aren't correctly colored after palette implementation
Possibly fixable by creating a `ButtonStyle` and applying it to our fake `Form`

EricBAndrews:
# Checklist
      - closes #1298 
      - closes #1268 

# Pull Request Information

This PR makes `Settings` codable. This is accomplished through the intermediate `CodableSettings` struct; `Settings` should not be made directly codable because doing so results in `AppStorage` values being updated every time any `Settings` object is decoded. `CodableSettings` ultimately should replace `Settings` entirely, migrating our settings storage from `AppStorage` to JSON, but that's OOS here; this means that in the interim we do have the slightly awkward situation of duplicating default values between `CodableSettings`' decoder and `Settings` itself.

Some logic has been added to `PersistenceRepository` to save and retrieve settings files. There are two different locations settings files are stored:
- System: these are app-generated settings files used for migration and persistence. From the code, they are identified by the `SystemSetting` enum.
- User: these are user-generated settings files. Currently this system is not used, but the idea is to offer easily configurable and swappable settings profiles. These can then optionally be attached to accounts.

The new codable settings enable the following features:

- Migrate from V1 settings
- Save/Restore settings: "Save settings" saves the current settings state to System/v2. "Restore settings" reverts settings to that configuration.
- Import/Export settings: allows users to directly import and export settings JSON files.

To test migration from v1 settings:
- Reset settings state
- Build and launch `eric/v1-settings-cache`
- Change some settings
- Close the app
- Build and launch this branch

You should see a "Settings Restored" toast and any settings you changed in v1 should be preserved.


Mlem/App/Views/Root/Tabs/Settings/ImportExportSettingsPage.swift
```diff
@@ -0,0 +1,104 @@
...
+//  Mlem
+//
+//  Created by Eric Andrews on 2024-09-06.
+//
+
+import Dependencies
+import Foundation
+import SwiftUI
+
+// DO NOT use Picker in this page! It refuses to change the theme until it gets re-rendered. All other components pick up theme changes correctly.
```
Sjmarf, line 12: I've noticed the text color in comments failing to pick up theme changes on occasion too, though every time I try to reproduce it I can't xD

Mlem/App/Globals/Definitions/PersistenceRepository.swift
```diff
@@ -47,6 +49,18 @@ private enum DiskAccess {
     }
 }
 
+// Enumeration of system-managed settings
+enum SystemSetting {
+    case v_1, v_2
```
Sjmarf: *nit* Maybe `case v1, v2` is better?

Sjmarf (changes requested):
Nice work!

We aren't currently tracking interaction bar configurations here - I think we should.

EricBAndrews:
Opened #1323 to handle widget persistence

Sjmarf (approved):
LGTM 💯

PR #1316, opened 2024-09-09
Search Sorting & Filtering

Sjmarf:
Added sorting and filtering to all 3 search tabs. Filtering users by instance is only available on 0.19.4 and above.

Also made search scroll infinitely.


Mlem/App/Enums/InstanceSort.swift
```diff
@@ -0,0 +1,32 @@
...
+        case .version: "Version"
+        }
+    }
+    
+    var systemImage: String {
+        switch self {
+        case .alphabetical: "textformat"
+        case .score: "star"
+        case .users: "person.2"
+        case .version: "server.rack"
```
EricBAndrews: *nit*

These should go in `Icons`

Mlem/App/Models/MlemStats/MlemStats.swift
```diff
@@ -54,12 +54,26 @@ class MlemStats {
     }
     
     @MainActor
-    func searchInstances(query: String) async throws -> [InstanceSummary] {
+    func searchInstances(query: String, sort: InstanceSort = .score) async throws -> [InstanceSummary] {
         try await MlemStats.main.loadInstances()
```
EricBAndrews: This logic is super weird--I must have missed it last review. This depends on the assumption that `searchInstances` will be called on `MlemStats.main`, since otherwise this `loadInstances` wouldn't do anything, but given that assumption it's redundant to call `MlemStats.main.loadInstances` here instead of a simple `loadInstances`. It's also a little odd to implicitly set `instances` with this call then filter and reset `instances` in the body of the function.

A cleaner way to handle this all would be:

- Extract the actual loading (L39-42 and associated error handling) into a `fetchInstances()` helper that returns `[InstanceSummary`
- Have `loadInstances` and `searchInstances` both call `fetchInstances`, then have `searchInstances` perform filtering on the response before assigning to `instances`
- Add guards to both methods that `self === MlemStats.main` with `assertionFailures` if not main

Alternatively, and I think possibly cleaner still:
- Move all loading work into `searchInstances` and make both `query` and `sort` optional
- Either delete `loadInstances` entirely or make it a simple passthrough to `searchInstances`that provides `nil` params
- Again, add `self === MlemStats.main` guard

Mlem/App/Views/Shared/Search/SearchView+FiltersView.swift
```diff
@@ -0,0 +1,217 @@
...
+                }
+                .padding(.vertical, 12)
+                .padding(.horizontal, Constants.main.standardSpacing)
+            }
+            .scrollIndicators(.hidden)
+            if selectedTab == .communities, communityFilters.instance.isOther {
+                Label(
+                    "Subscription statuses can't be displayed when using these filters.",
+                    systemImage: Icons.warning
+                )
```
EricBAndrews, line 34: We can actually do this pretty easily by checking whether `(appState.firstSession as? UserSession)?.subscriptions.communities` contains a community with matching actor id--there's the slight challenge of tier, since that set expects `Community2`, but that should be fairly easily worked around to get the nice constant-time lookup behavior.

Mlem/App/Views/Shared/Search/SearchView+FiltersView.swift
```diff
@@ -0,0 +1,217 @@
...
+        @Environment(NavigationLayer.self) var navigation
+        @Binding var filter: InstanceFilter
+        let isForPersonSearch: Bool
+        
+        var allowActiveAccountLocalInstanceSearch: Bool {
+            !isForPersonSearch || (AppState.main.firstApi.fetchedVersion ?? .infinity) >= .v19_4
+        }
+        
+        var body: some View {
+            Menu("Instance: \(filter.label)", systemImage: Icons.instance) {
```
EricBAndrews: I don't think we need the explicit "Instance:" text--the icon should make the behavior reasonably evident, and it'll be very obvious once the user taps it once. We also don't have any additional text on the sort selector, so removing "Instance:" would be consistent.

Mlem/App/Views/Shared/Search/SearchView+FiltersView.swift
```diff
@@ -0,0 +1,217 @@
...
+                Toggle(
+                    "Any",
+                    systemImage: "point.3.filled.connected.trianglepath.dotted",
```
EricBAndrews: *nit*

`Icons`

Mlem/App/Views/Shared/Search/SearchView+FiltersView.swift
```diff
@@ -0,0 +1,217 @@
...
+    
+    private struct FilterButtonStyle: ButtonStyle {
+        @Environment(Palette.self) var palette
+        
+        let isOn: Bool
+        
+        func makeBody(configuration: Configuration) -> some View {
+            HStack(spacing: 4) {
+                configuration.label
+                Image(systemName: "chevron.down.circle.fill")
```
EricBAndrews: *nit*

`Icons`

EricBAndrews (changes requested):
Nice! The design is super clean 😁

Mlem/App/Models/MlemStats/MlemStats.swift
Sjmarf: Yeah, makes sense. I think at some point I had this function outside of `MlemStats` and then moved it inside - I must have missed changing `MlemStats.main.loadInstances()` to just `loadInstances`. I'll fix it 👍 

> It's also a little odd to implicitly set instances with this call then filter and reset instances in the body of the function.

`self.instances` isn't changed in the body of `searchInstances` - `instances` is local to the function scope.

Mlem/App/Views/Shared/Search/SearchView+FiltersView.swift
Sjmarf: Yeah, I did consider that. That fixes showing subscription status, but doesn't allow you to subscribe to or unsubscribe from the community. So, we'd still have to show some sort of warning to explain the absence of the "subscribe" action. There's probably some way to fix both at once, but I'm not quite sure exactly what it is yet and I didn't want to dedicate too much time to a relatively insignificant feature. Are you happy to skip it for now?

The answer might be to go full MVVM and implement a `CommunityViewModel` that stores a `Community` but remains fully interactable when that community's api is a guest or inactive. If it is interacted with in that state, it would fetch and replace it's own `Community` with a new one from the active api, then perform the call on that new model. If replicated for posts, this system would also allow you to upvote/downvote/save posts in the feed after switching to a new account. There's probably a whole load of headaches associated with that approach though that I don't want to get into just yet - state faking, for one. A ViewModel could also be a good place to handle multi-account stuff.

Mlem/App/Views/Shared/Search/SearchView+FiltersView.swift
Sjmarf: Sure, that's a good idea 👍

Mlem/App/Views/Shared/Search/SearchView+FiltersView.swift
EricBAndrews: Sure, that's fine--I'd rather avoid MVVM if possible, since SwiftUI is built to avoid it, so I'm content leaving OOS.

EricBAndrews (approved):
LGTM! 🔥

PR #1317, opened 2024-09-11
V1 Settings Cache

EricBAndrews:
This PR adds logic to the v1 codebase that stores the current user settings every time the app moves into the background.

Any custom types, including API types, have been captured in `CodableSettings` as `String`; there are several values that need to be converted or synthesized into v2 values, and this is simpler and more explicit than mucking around with custom `Codable` conformances.

I've also added a debug-only button to print the settings encoding to console for manual inspection.


Sjmarf (approved):
Looks good, pending SwiftLint fixes 👍

PR #1320, opened 2024-09-13
Fix comment editor sometimes not submitting correct text

Sjmarf:
Closes https://lemmy.ml/post/20201916

My repro for this was to click the Select Text button, at which point it would stop responding to changes of text. There may also be other ways to trigger this.


EricBAndrews (approved):
Nice 👍

PR #1326, opened 2024-09-15
Bubbly app design

Sjmarf:
Drafting for now - still need to work out some transitions


Sjmarf:
I've commented out the fancy transition code for now, as it requires Xcode 16 and we can't upgrade to Xcode 16 until #1171 is fixed.

Mlem/App/Views/Pages/Community/CommunityView.swift
```diff
@@ -145,9 +139,11 @@ struct CommunityView: View {
             }
             if let description = community.description {
                 Markdown(description, configuration: .default)
+                    .padding(Constants.main.standardSpacing)
+                    .background(palette.secondaryGroupedBackground, in: .rect(cornerRadius: 10))
```
EricBAndrews: We should standardize corner radii, since we're suddenly using them a _lot_. Generally we have corner radius = spacing, which is convenient and also aesthetically pleasing--we could roll it into `standardSpacing`

Mlem/App/Views/Pages/Person/PersonView.swift
EricBAndrews, line 1: I'm not sure where the precise culprit is but a little bit of padding got added that mucks up the tile layout:

Screenshot 2024-09-18 at 6 01 19 PM

EricBAndrews (changes requested):
Really nice work! Makes the app feel much more modern.

A couple notes, plus Slack discussion

Mlem/App/Views/Pages/Person/PersonView.swift
Sjmarf: Nice catch 👍

Mlem/App/Views/Pages/Community/CommunityView.swift
Sjmarf: Sure, makes sense 👍

EricBAndrews (approved):
LGTM! 🔥

PR #1327, opened 2024-09-15
iOS 18 mark read on scroll

EricBAndrews:
Improves mark read on scroll behavior using new iOS 18 tricks. Posts are now marked when they disappear off the top of the screen, no cursed index hacks required.

On iOS 17, the old behavior will be used instead.


Sjmarf (approved):
Nice! That's super clean :D

PR #1331, opened 2024-09-16
Context Menu Failures

EricBAndrews:
# Checklist
      - community feedback

# Pull Request Information

This PR fixes an issue where context menu actions were failing on images using `/api/v3/image_proxy?url=<url>` proxies. There were two related bugs:
1. `fullSizeUrl` was removing the `url=<url>` query parameter along with the `thumbnail=<size>` query param. It now only removes `thumbnail`.
2. `downloadImageToFileSystem` was using the URL's `pathExtension` to determine the image type. When given an image proxied through a URL param, `pathExtension` was `""`. This PR adds logic to extract the `pathExtension` from the url param if extraction from the original URL fails, plus an `assertionFailure` to catch empty path extensions.

## Bug Repro

To reproduce the bad behavior that this fixes, simply try to save/share/quickLook any image on lemm.ee.


Sjmarf (approved):
LGTM 👍

PR #1332, opened 2024-09-17
Xcode 16 Swipe Action Fix

### LINKED ISSUE #1171: [2.0] Scroll gestures sometimes ignored on iOS 18

### LINKED ISSUE #1183: [2.0] Swipe action gesture occurs simultaneously with horizontal scroll gesture
https://github.com/user-attachments/assets/0c9ee7d9-9430-448f-82d9-2672ca7e26a0

Sjmarf:
Fixes #1171, which only occurs when running on iOS 18 having built from Xcode 16. The fix uses a new API so the old solution will have to stick around too.


EricBAndrews (approved):
Nice! 🔥

PR #1343, opened 2024-09-19
[V1] Scroll/Swipe Conflict Fix

EricBAndrews:
Implements PanGesture on the v1 codebase.


Sjmarf (changes requested):
It appears to be working on some posts but not others. This happens both on a sim and my physical device

https://github.com/user-attachments/assets/f805774c-68ed-4976-8115-63d528ffd6cf

Sjmarf (approved):
LGTM 💯

PR #1348, opened 2024-09-20
Bubbly App Fixes

EricBAndrews:
Some minor tweaks to the bubbly app design:
- Fixed shadow not appearing on quick swipes
- Fixed inner card not always rounding to the given corner radius


Sjmarf (approved):
LGTM!

PR #1356, opened 2024-09-23
Sign Up Prompts

Sjmarf:
When the user has no permanent accounts, they'll see the following prompts:

The "Welcome to Lemmy!" prompt can be dismissed by pressing the X button, and brought back by going to Settings -> Advanced -> Developer.

If the user is using a saved guest account rather than a temporary one, they'll see this prompt in the Inbox tab:

(I also moved some of the `InboxView` code into another file; SwiftLint was complaining about it)


EricBAndrews (approved):
LGTM! 🚀

PR #1367, opened 2024-09-29
`CommentPage`

### LINKED ISSUE #1366: Assertion failure when tapping menu button to navigate to instance
Repro:
- Go to community
- Try to navigate to instance using link in ellipsis menu
- MenuButton L29 should trigger

Sjmarf:
Adds this view:

In this PR, the menu is shown anytime a comment is navigated to. What do you think about that? Are there some cases where the old behavior would be better? Theoretically this is safer than the old behaviour, because it guarantees that the comment is loaded in the comment list (the old method did not). Though, that doesn't happen particularly frequently and could be worked around by sending a second API request. The one case where we _must_ display this view is when a comment chain gets so deep that the comments become too thin.

I've also added the "Load more comments" buttons when the comments aren't initially loaded.

Here's some posts where the "Load more" button is shown:

https://lemmings.world/post/405439
https://lemmy.ml/post/13565023


Sjmarf:
Also closes #1366

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView.swift
```diff
@@ -198,6 +166,53 @@ struct ExpandedPostView: View {
...
+                    } label: {
+                        HStack {
+                            CommentBarView(depth: comment.depth + 1)
+                            HStack {
+                                Text("More Replies")
+                                Image(systemName: Icons.forward)
+                            }
+                            .frame(maxWidth: .infinity)
+                            .padding(.vertical, 2)
+                            .foregroundStyle(.secondary)
```
EricBAndrews: *nit*

Or `.accent` (see general comment)

```suggestion
                            .foregroundStyle(palette.secondary)
```

EricBAndrews (changes requested):
Code looks good!

Some design notes:

- "More Replies" can be the same font size as a comment, and maybe wants to be `.accent` to indicate that it's a button
Screenshot 2024-09-30 at 6 05 11 PM
- Selecting "View All" after navigating into `CommentPage` from `ExpandedPostView` should take you back rather than pushing a new `ExpandedPostView`
- The highlight on the navigated-to comment should fade after a moment
- The "View All"/"View Parent" buttons feel a bit heavy, though I'm not sure what a nice way to lighten them would be

Also, maybe OOS for this PR, but `CommentTreeTracker` L150 should be `@MainActor`--got a crash on that line while testing

Sjmarf:
> The "View All"/"View Parent" buttons feel a bit heavy, though I'm not sure what a nice way to lighten them would be

Maybe like this?



> Selecting "View All" after navigating into `CommentPage` from `ExpandedPostView` should take you back rather than pushing a new `ExpandedPostView`

What about just hiding the "View All" button in that context? Making it go back feels weird to me - there's already a back button in the top-left of the screen

EricBAndrews (approved):
LGTM! 🚀

PR #1382, opened 2024-10-11
OLED theme

EricBAndrews:
# Checklist
      - Community feedback

# Pull Request Information

This PR adds an OLED theme that makes all backgrounds true black. To distinguish card entities, a light border is applied to the entity using the new `.paletteBorder` view modifier. This border has the same color as the palette-aware `Divider`.

To support this, `PaletteProviding` now defines a `bordered` boolean. `ColorPalette` defaults this to `false`, as it is a non-standard option. The divider color is now also a computed property of `PaletteProviding`

`List` and `Form` do not easily support a rounded border, so these views instead have a light shadow applied to them.

This PR also makes a couple minor changes:
- Padding on collapsed comments is adjusted to perfectly center the username within the collapsed comment when compact posts are off
- Padding on some account settings entities has been updated to play nice with the new shadow

### DIFF
--- a/Mlem/App/Configuration/Colors/OledPalette.swift
+++ b/Mlem/App/Configuration/Colors/OledPalette.swift
@@ -10,6 +10,7 @@ import SwiftUI
 extension ColorPalette {
     static let oled: ColorPalette = .init(
         supportedModes: .dark,
+        bordered: true,
         primary: .primary,
         secondary: .secondary,
         tertiary: Color(UIColor.tertiaryLabel),
--- a/Mlem/App/Configuration/Colors/OledPalette.swift
+++ b/Mlem/App/Configuration/Colors/OledPalette.swift
@@ -8,38 +8,16 @@
 import SwiftUI
 
 extension ColorPalette {
-    static let oled: ColorPalette = .init(
-        supportedModes: .dark,
-        bordered: true,
-        primary: .primary,
-        secondary: .secondary,
-        tertiary: Color(UIColor.tertiaryLabel),
-        background: .black,
-        secondaryBackground: .black,
-        tertiaryBackground: .black,
-        groupedBackground: .black,
-        secondaryGroupedBackground: .black,
-        tertiaryGroupedBackground: .black,
-        thumbnailBackground: Color(UIColor.systemGray4),
-        positive: .green,
-        negative: .red,
-        warning: .red,
-        caution: .orange,
-        upvote: .blue,
-        downvote: .red,
-        save: .green,
-        read: .purple,
-        favorite: .blue,
-        selectedInteractionBarItem: .white,
-        administration: .teal,
-        moderation: .cyan,
-        federatedFeed: .blue,
-        localFeed: .purple,
-        subscribedFeed: .red,
-        inbox: .purple,
-        accent: .blue,
-        neutralAccent: .gray,
-        colorfulAccents: [.orange, .pink, .blue, .green, .purple, .indigo, .mint],
-        commentIndentColors: [.red, .orange, .yellow, .green, .blue, .purple]
-    )
+    static let oled: ColorPalette = {
+        var palette = ColorPalette.standard
+        palette.supportedModes = .dark
+        palette.bordered = true
+        palette.background = .black
+        palette.secondaryBackground = .black
+        palette.tertiaryBackground = .black
+        palette.groupedBackground = .black
+        palette.secondaryGroupedBackground = .black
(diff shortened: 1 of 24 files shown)


Mlem/App/Utility/Extensions/Views/View Modifiers/View+PaletteBorder.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+//  Created by Eric Andrews on 2024-10-10.
+//
+
+import SwiftUICore
+
+import Foundation
+import SwiftUI
+
+private struct PaletteBorder: ViewModifier {
```
Sjmarf, line 13: This needs to be applied to the crosspost list in `PostPage`

Edit: and the "New post" button in Profile, and the post in `ExpandedPostView`, and the Large mode website preview

Mlem/App/Views/Shared/Palette Components/Form.swift
```diff
@@ -22,7 +22,9 @@ struct Form<Content: View>: View {
                 .tint(palette.accent)
                 .buttonStyle(PaletteButton())
         }
+        .listStyle(InsetListStyle())
         .scrollContentBackground(.hidden)
         .background(palette.groupedBackground)
+        .shadow(color: palette.primary.opacity(palette.bordered ? 0.2 : 0.0), radius: 5)
```
Sjmarf, line 28: This shadow shows up on `Form` sections for which we removed the background using `.listRowBackground`. Possibly fixable by applying another `.shadow` on those views or something?

Mlem/App/Configuration/Colors/OledPalette.swift
```diff
@@ -0,0 +1,45 @@
...
+        moderation: .cyan,
+        federatedFeed: .blue,
+        localFeed: .purple,
+        subscribedFeed: .red,
+        inbox: .purple,
+        accent: .blue,
+        neutralAccent: .gray,
+        colorfulAccents: [.orange, .pink, .blue, .green, .purple, .indigo, .mint],
+        commentIndentColors: [.red, .orange, .yellow, .green, .blue, .purple]
+    )
```
Sjmarf: I wonder if it's better to do something like this? place

```swift
static let oled: ColorPalette = {
    var palette = ColorPalette.standard
    palette.background = .black
    // etc
    return palette
}()
```

That avoids defining the standard accents in more than one

Sjmarf (changes requested):
Looks good! A few thoughts

Mlem/App/Globals/Definitions/Palette.swift
```diff
@@ -67,6 +69,7 @@ enum PaletteOption: String, CaseIterable, Codable {
     var label: LocalizedStringResource {
         switch self {
         case .standard: "Default"
+        case .oled: "Oled"
```
Sjmarf: "OLED" would be better than "Oled" imo

Mlem/App/Views/Shared/Palette Components/Form.swift
EricBAndrews: Nice catch--this one was a little hacky, I had to move the content into the header of an empty Section. This does seem to fix the spacing issues that we were previously solving with negative padding.

Mlem/App/Utility/Extensions/Views/View Modifiers/View+PaletteBorder.swift
EricBAndrews: Nice catch!

Mlem/App/Configuration/Colors/OledPalette.swift
EricBAndrews: Yeah, that's cleaner 👍

Mlem/App/Views/Root/Tabs/Settings/AccountSettingsView.swift
```diff
@@ -14,17 +14,17 @@ struct AccountSettingsView: View {
     
     var body: some View {
         Form {
-            Section {
+            // empty section disables background
+            Section {} header: {
```
Sjmarf, line 18: This makes the banner view thinner than it was before - possibly `.listRowInsets` would fix that?

We also use the empty section in the header of the "About" page

Sjmarf (changes requested):

Sjmarf:
The text fields in the sign-in flow also need `PaletteBorder`; I missed that one earlier :D

Sjmarf (approved):
LGTM! 💯

PR #1383, opened 2024-10-11
Expanded Post View Polish

Sjmarf:
- Added the "no comments" text
- Added a slight fade transition when the comments load in
- Added an `ErrorView` that is shown if the comments fail to load

### DIFF
--- a/Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView+Views.swift
+++ b/Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView+Views.swift
@@ -1,0 +1,94 @@
+//
+//  ExpandedPostView+Views.swift
+//  Mlem
+//
+//  Created by Sjmarf on 11/10/2024.
+//
+
+import MlemMiddleware
+import SwiftUI
+
+extension ExpandedPostView {
+    @ViewBuilder
+    var noCommentsView: some View {
+        VStack(spacing: 5) {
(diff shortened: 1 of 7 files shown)


Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView+Views.swift
```diff
@@ -0,0 +1,94 @@
...
+    var noCommentsView: some View {
+        VStack(spacing: 5) {
+            Image(systemName: Icons.noContent)
+                .font(.title)
+                .foregroundStyle(.tertiary)
+            VStack {
+                Text("No comments to be found")
+                    .fontWeight(.semibold)
+                Text("Why not post the first one?")
+                    .font(.footnote)
```
EricBAndrews: Not sure we need the cutesy language--we've been moving away from it across the codebase, and personally I never really liked it in V1 either. Maybe just "No comments found?"

```suggestion
                Text("No comments found")
                    .fontWeight(.semibold)
```

EricBAndrews (approved):
LGTM! 🔥 

Minor note on wording.

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView+Views.swift
Sjmarf: Sure 👍

PR #902, opened 2024-02-05
Easy Tap Link Settings

EricBAndrews:
# Checklist
- [x] I have read CONTRIBUTING.md
- [x] I have described what this PR contains
- [x] If this PR alters the UI, I have attached pictures/videos
- [ ] This PR addresses one or more open issues that were assigned to me:

# Pull Request Information

This PR addresses two pieces of community feedback regarding easy-tap links:
- They're too big in compact comments
- There should be an option to disable them

I have added a setting to customize the behavior of the link, with four options:
- Disabled: easy tap links will not display
- Compact: easy tap links will only display the title, and not the caption
- Large: easy tap links will always display both title and caption
- Contextual: easy tap links will display as compact in compact environments (i.e., compact comments) and expanded otherwise

PR #1384, opened 2024-10-13
Tappable Links

### LINKED ISSUE #1072: [2.0] Tappable links
In v1.0, we extracted these using Regex. For 2.0, [I've implemented the link extraction logic in LemmyMarkdownUI](https://github.com/mlemgroup/LemmyMarkdownUI/pull/9). This has several advantages:

- It won't look for links inside of code blocks.
- We can choose not to render links inside of spoilers.
- We can extract the title of each link as a series of nodes, allowing the tappable link titles to be rendered in the same markdown format as the links in the original text.
- It's about 8-10x faster than our regex solution (tested on the markdown in the Profile tab), because it simply traverses an existing markdown tree.

## LemmyMarkdownUI implementation

Get the links contained within some markdown like so:

```swift
let blocks: [BlockNode] = .init(markdown: string)
let links = blocks.links
```

`links` is an array of `LinkData`, which is defined as follows:

```swift
struct LinkData {
   let title: [InlineNode]
   let url: URL
   var insideSpoiler: Bool
}
```

## Using this in Mlem

LemmyMarkdownUI's `Markdown` view has an initialiser that accepts `[BlockNode]`. This means that we can create the markdown tree in Mlem, extract the links from it, then pass the tree to `Markdown` to render it. This should be pretty performant.

Notice that `LinkData.title` is an array of inline nodes. When rendering the title of each tappable link, we can pass this array to LemmyMarkdownUI's `MarkdownText` view to render the title with markdown features. Or, you can use `title.stringLiteral` to get the raw string with markdown delimiters removed.

We should probably not include links that are inside spoilers in the tappable links list. To exclude these, we can use something like `links.filter { !$0.insideSpoiler }`.

Sjmarf:
Added tappable links (closes #1072)

In v1 we had an option for compact links. Is that something we'd want to keep? Personally I think that's too complicated - the whole point of tappable links is that they're large enough to tap easily. Hiding the URL and making them smaller seems to defeat that purpose. A simple on/off toggle would be better imo.

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -2812,7 +2812,7 @@
 			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware";
 			requirement = {
 				kind = upToNextMinorVersion;
-				minimumVersion = 0.43.0;
+				minimumVersion = 0.44.0;
 			};
 		};
 		CD4368BF2AE23FD400BD8BD1 /* XCRemoteSwiftPackageReference "Semaphore" */ = {
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -184,7 +184,7 @@
 		039F589D2C7B6C0A00C61658 /* FormChevron.swift in Sources */ = {isa = PBXBuildFile; fileRef = 039F589C2C7B6C0A00C61658 /* FormChevron.swift */; };
 		03A82FA12C0D1E8500D01A5C /* ApiClient+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03A82FA02C0D1E8500D01A5C /* ApiClient+Extensions.swift */; };
 		03A82FA32C0D1F2400D01A5C /* View+ExternalApiWarning.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03A82FA22C0D1F2400D01A5C /* View+ExternalApiWarning.swift */; };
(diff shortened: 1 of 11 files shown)


EricBAndrews:
Per #902, we got community feedback on v1 that the links are too large in compact comments.

> the whole point of tappable links is that they're large enough to tap easily. Hiding the URL and making them smaller seems to defeat that purpose.

Even without the URL it's a much easier hitbox than the bare Markdown, especially if the link is on a small character or there are several links closely packed together.

Sjmarf:
Makes sense; I've added a setting for it 👍 The setting layout is a little tricky - ideally I'd like to avoid using the word "contextual" as an option, as it isn't immediately obvious what that means. What do you think of this layout?

With "compact comments" off:


With "compact comments" on:

EricBAndrews:
That's definitely more comprehensible than "contextual," but the settings behavior is now somewhat unintuitive--having a setting in one page affect the options available in another isn't great. Some other ideas:
- "Automatic" instead of "Contextual"--slightly more normal verbiage, and we can add an explanatory footer. I like this approach the best, since it keeps the behavior on one page.
- Move the option for compact behavior into the comments settings and label it something like "hide link URL," though this still splits the behavior across multiple views and we'd probably have to add a note in the links section when this is on indicating that the behavior will be overridden in comments.

EricBAndrews (commented):
Code looks good! Will approve once we find a good way to display the compact comments link settings.

EricBAndrews:
Might also be convenient to roll a MlemMiddleware version bump to 0.44.0 into this PR

Sjmarf:
> "Automatic" instead of "Contextual"--slightly more normal verbiage, and we can add an explanatory footer. I like this approach the best, since it keeps the behavior on one page.

Sure, I'll do that. I like "automatic" better than "contextual" 👍 

>  Might also be convenient to roll a MlemMiddleware version bump to 0.44.0 into this PR

Will do 👍

EricBAndrews (approved):
LGTM! 🔥

PR #1408, opened 2024-10-26
"Swipe anywhere to Navigate" setting

### LINKED ISSUE #130: Swipe to navigate
## Swipe navigation
Allow swipe from anywhere to go back similar to Apollo / Relay

## Why
Swiping from the edge of the screen or reaching to the top left corner can be awkward with one hand

## What I Need

- Option to enable / disable the feature (to allow use of swipe to vote for those who prefer that)
- Ability to swipe anywhere to go back

Sjmarf:
If `General` -> `Swipe anywhere to Navigate` is enabled, you can swipe anywhere on the screen to go back rather than only at the edges.

Also fixed a bug in which feed post corners weren't clipped if Swipe Actions were turned off.

Closes #130


EricBAndrews (changes requested):
Code looks good, works well!

If swipe actions are on, it may not be immediately clear that they need to be disabled for this setting to be activated--we should at least offer an explanatory footer, and possibly flip the swipe actions toggle off when swipe navigation is enabled (the same as is done when going the other way)

Sjmarf:
Hm, I can't think of a good footer that's not too verbose. Any ideas? Alternatively we could swap out the two toggles for a single picker or something? That might not be as clear though

EricBAndrews:
The best I can come up with for footer is "Only one of these can be active at a time"--I agree that a picker isn't as clear. We might not need a footer if they just automatically swap--that behavior _should_ be clear enough, though you can never be too sure

Sjmarf:
I've made both toggles enabled at all times, and turning one on turns the other off. Maybe that's clear enough without needing a footer?

EricBAndrews:
Works for me 👍

EricBAndrews (approved):

Issue #1416, opened 2024-10-29
Token unable to be read from keychain for specific accounts


nocturne213:
**Describe the bug**
What did you do to trigger the bug?
Added a second account on ipad

**Expected Behavior**
What did you expect to happen?
I would be able to swap between accounts

**Actual Behavior**
What actually happened?
The next time i opened mlem both of my accounts were removed and i had to log in again.

**To Reproduce**
What steps do I need to take to trigger the bug?
Log into an account (i was using sipuli.xyz) then add a second account (i used mtgzone.com) then when opening a link from within a post mlem fully closed, when i reopened mlem neither account was available and mlem was showing a guest Lemmy.world account.

**Mlem Information:**
Version + Build number
If this started happening recently, write which version introduced the bug
I remember this happening when version 2.0 first came out, i almost exclusively use my sopuli account so I never tried again after the first time until tonight. I am using ver 2.0 (114) on ipad pro iOS 18.1.

**Additional Context**
Any other relevant information

nocturne213:
It appear to not be limited to iPad, but possibly certain instances. I was logged into mtgzone on both my iPhone and iPad and was logged out of all accounts on both devices. I previously had 2 or 3 different accounts logged in on the iPhone. The only difference was adding the mtgzone account.

Sjmarf:
Hmm, that's interesting. I've not seen this issue so it being related to mtg.zone is plausible. Accounts being wiped is a result of a decoding error when opening the app, normally... we should really make Mlem log those errors somewhere permanent to make debugging easier. 

When the accounts gets wiped, an "error" popup might be shown - if it happens again and you tap on the "error" popup before it goes away, it'll show the exact error. On the off chance it happens again, having that error information would be useful for us.

nocturne213:
The operation couldn’t be completed. (Mlem.UserAccount.DecodingError error 1.)

after more testing it happens on iPhone as well as iPad. It happens when I am logged in only on mtgzone.com as well as logged into other instances as well. It was more noticeable on iPad because pretty much every time you minimize Mlem on iPad it crashes. But if you log in on iPhone, then swipe away Mlem and reopen it it gives the above error and logs out all accounts.

Sjmarf:
Thanks, that's narrowed it down a lot.

[...]

PR #1419, opened 2024-10-30
Don't wipe all accounts if keychain token cannot be found

Sjmarf:
Previously if a token wasn't found in the keychain Mlem would reset the JSON account list file. Instead of resetting the JSON file, I've made it fallback to setting an arbitrary string as the token. This will cause Mlem to pick up on the token being invalid when it tries to perform an authenticated request, and show the re-auth sheet.

In normal circumstances, a token should always be present in the keychain. However, in #1416 an account's token is consistently missing from the keychain. I haven't been able to reproduce this exact issue myself yet.


EricBAndrews (approved):
LGTM 🚀

PR #1430, opened 2024-11-10
Show post context for comments in feeds

### LINKED ISSUE #1390: Show post and community information on comments view in user profiles
## Introduction
The current comment view on any user profile only shows the comment as it would appear under the post. It makes sense to not include the post or community information when viewing the comment in the context of the post. However, when viewing comments under a user's profile, it would make sense to include post information: post title, poster@instance, and community@instance.

Sjmarf:
Closes #1390

Do you think tapping the link should open the `CommentPage` (current behavior) or the parent `PostPage`? I'm not sure.


EricBAndrews (approved):
LGTM! 🚀 

Re: behavior, I think it makes more sense for tapping the link to navigate to the post while tapping the comment itself goes to the comment page--that also gives the user easy options to go to either as they wish.

Issue #832, opened 2023-12-28
Disable Infinite Scrolling Option


Sjmarf:
Instead of infinite scrolling, there'd be an option to load the next n pages at the bottom of a page. Apollo had this. Helps you manage how much time you spend on Lemmy haha.

PR #1434, opened 2024-11-13
Actor Feed Loaders

### LINKED ISSUE #1402: Moderated feed slow to load
The moderated feed can reportedly be very slow to load, perhaps when it contains fewer than 50 items?

EricBAndrews:
Frontend changes for [the real PR](https://github.com/mlemgroup/MlemMiddleware/pull/70).

Contains two minor changes of note:
- `EndOfFeedView` can now take in an optional `loadMore` callback; if available, a "Load More" button will be rendered at the bottom of the feed if the `loadingState` is `idle`. Sometimes hiding read posts creates a feed that won't trigger the infinite load; this handles that case without needing to do weird threshold hacks. This will also make implementing #832 trivial.
- `loadFeed` now triggers if the loader is empty regardless of `loadingState`. The middleware PR changes the initial `loadingState` to be `.loading`, which in turn results in a spinner being immediately rendered when an empty feed is opened.

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -128,8 +128,8 @@
 		037331A42C9CB12D00C826E1 /* EnvironmentValues+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 037331A32C9CB12D00C826E1 /* EnvironmentValues+Extensions.swift */; };
 		037386472BDAFE81007492B5 /* LemmyMarkdownUI in Frameworks */ = {isa = PBXBuildFile; productRef = 037386462BDAFE81007492B5 /* LemmyMarkdownUI */; };
 		037658DF2BE7D9EF00F4DD4D /* Community1Providing+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 037658DE2BE7D9EF00F4DD4D /* Community1Providing+Extensions.swift */; };
-		037DE07A2CE108D9007F7B92 /* FooterLinkView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 037DE0792CE108D9007F7B92 /* FooterLinkView.swift */; };
 		037DE0752CE023E3007F7B92 /* BlockListView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 037DE0742CE023E3007F7B92 /* BlockListView.swift */; };
+		037DE07A2CE108D9007F7B92 /* FooterLinkView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 037DE0792CE108D9007F7B92 /* FooterLinkView.swift */; };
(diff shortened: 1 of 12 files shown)


Mlem.xcodeproj/project.pbxproj
```diff
@@ -2859,20 +2859,20 @@
...
 			};
 		};
-		CD64A9182CA37D63007CA7E6 /* XCRemoteSwiftPackageReference "Gifu" */ = {
+		CD555F802CE56CB200E763E7 /* XCRemoteSwiftPackageReference "MlemMiddleware" */ = {
 			isa = XCRemoteSwiftPackageReference;
-			repositoryURL = "https://github.com/kaishin/Gifu";
+			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware";
 			requirement = {
-				branch = master;
+				branch = "eric/actor-feed-loaders";
```

Mlem/App/Views/Shared/Search/SearchView.swift
Sjmarf, line 1: There's a couple of bugs related to post searching in this PR:

1) Changing any of the post filters doesn't refresh the post feed. Possibly missing `@Observable` somewhere in MlemMiddleware?
2) With the search bar unfocused, tapping on the "Posts" tab in Search loads the results for the search query `""`. Before this PR, it would wait until the user submits their search query before loading anything.

Mlem/App/Views/Shared/PersonContentGridView.swift
Sjmarf, line 1: In the saved feed, switching account without reloading and then tapping the "refresh" button loads the previous account's saved content rather than the saved content of the new account.

Sjmarf (changes requested):

Mlem/App/Views/Shared/PersonContentGridView.swift
EricBAndrews: Nice catch! I've implemented some custom refresh logic for `PersonContentFeedLoader` that should fix this.

Mlem/App/Views/Shared/Search/SearchView.swift
EricBAndrews: (2) was due to the change in default loading state causing that first load--I've changed that behavior for the `SearchPostFeedLoader` on the middleware PR. I could only repro (1) using the erroneous load from (2)--the filters trigger refreshes properly for me when using a real search.

Sjmarf: Yep, it looks like it was only broken for that first erroneous load 👍

Sjmarf (approved):
LGTM! Closes #1402 😄

Sjmarf (changes requested):
Hmm I might have found another bug actually. Searching for "Mlem" in community search doesn't bring anything up anymore 🤔 

Edit: Maybe it's related to having < pageSize items in the first page of results? This also seems to happen in the post feed - going to mlemtesting with 'Hot' selected also displays no results.

EricBAndrews:
> Maybe it's related to having < pageSize items in the first page of results?

Yep, that was it--forgot to return the undersized page instead of the default empty array when I made that change lol

Sjmarf (approved):
LGTM 👍

PR #1464, opened 2024-11-28
Expand comments inline

### LINKED ISSUE #1451: When clicking "more replies", expand comments inline in most cases
We should show a new page only if one of the expanded comments has a depth greater than n

Sjmarf:
Closes #1451


EricBAndrews (approved):
LGTM! 🔥

PR #1468, opened 2024-11-30
OLED theme tweaks

### LINKED ISSUE #1462: Easy tap link title alignment
Should be leading rather than centered

Sjmarf:
- Changed the form shadow radius to 1px so that it looks like a solid border
- Added missing `.paletteBorder` cases
- Fixed tappable link text alignment (closes #1462)


EricBAndrews (approved):
LGTM! ✅ 

That's a pretty clean hack to fix the "glow"

PR #1471, opened 2024-12-01
Redact user token from error toast

### LINKED ISSUE #1453: Redact user token from error popup
Under certain circumstances the error popup can include the URL which caused the error, which in turn contains the user's token. We should redact the token from the error message if so, in case they post the error message somewhere.

Sjmarf:
Closes #1453


EricBAndrews (approved):
LGTM! 💯

PR #1476, opened 2024-12-03
Hide downvote button if disabled on instance

### LINKED ISSUE #1475: Hide downvote button if downvotes disabled on instance

Sjmarf:
Closes #1475

### DIFF
--- a/Mlem/App/Utility/Extensions/Content Models/Comment1Providing+Extensions.swift
+++ b/Mlem/App/Utility/Extensions/Content Models/Comment1Providing+Extensions.swift
@@ -32,7 +32,7 @@ extension Comment1Providing {
             leadingActions: {
                 if api.canInteract {
                     upvoteAction(feedback: [.haptic])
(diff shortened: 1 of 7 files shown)


Mlem/App/Utility/Extensions/Content Models/Comment1Providing+Extensions.swift
```diff
@@ -32,7 +32,9 @@ extension Comment1Providing {
             leadingActions: {
                 if api.canInteract {
                     upvoteAction(feedback: [.haptic])
-                    downvoteAction(feedback: [.haptic])
+                    if api.myInstance?.downvotesEnabled ?? true {
```
EricBAndrews: *nit*

This `api.myInstance?.downvotesEnabled ?? true` phrase gets repeated a lot--it would be cleaner to have it as a computed var `api.downvotesEnabled`.

EricBAndrews (approved):
Code looks good!

One nit re: duplication--should be addressed or dismissed before merge but not important enough to warrant re-review

PR #1494, opened 2024-12-12
Sort Picker Changes

### LINKED ISSUE #1212: Hide most sorting options
Lemmy has a *lot* of sorting options, but many people will never use most of them. We could hide the lesser-used ones, and make them accessible via a "More..." button at the bottom of the sort picker. Tapping "More..." would open a sheet with the hidden sort modes. In that sheet, users would also be able to change which sort modes are shown/hidden. By default, we could show:

- Hot
- New
- Top Day
- Top Week
- Top Month
- Top All Time

### LINKED ISSUE #1217: Scaled Sort explanation
The mechanism of this sort mode isn't obvious. When the user first switches to Scaled, we could show a Tip (possibly using TipKit) at the top of the feed that explains what it is.

Sjmarf:
This is just the sort picker stuff - I haven't moved filters (e.g. "Hide read" etc) here yet.

Closes #1212
Closes #1217

If there are any unavailable sort modes, they're shown at the bottom of the list.


Mlem/App/Views/Pages/Community/AdvancedSortView+SortButton.swift
```diff
@@ -0,0 +1,101 @@
...
+                                Text("Requires Lemmy \(type.minimumVersion) or later")
+                                    .multilineTextAlignment(.leading)
+                                    .foregroundStyle(palette.warning)
+                                    .font(.footnote)
+                            }
+                        }
+                        .padding(.vertical, Constants.main.halfSpacing)
+                        Spacer()
+                        if selectedSort == type {
+                            Image(systemName: "checkmark")
```
EricBAndrews: *nit*

`Icons`

Mlem/App/Views/Pages/Community/TopSortPicker.swift
```diff
@@ -0,0 +1,85 @@
...
+            .filter { includeAll ? true : PinnedSortTracker.main.pinnedSortTypes.contains($0) }
+            .filter { (appState.firstApi.fetchedVersion ?? .infinity) >= $0.minimumVersion }
+    }
+    
+    var body: some View {
+        VStack(spacing: 10) {
+            Text("Choose Timeframe...")
+                .font(.footnote)
+                .foregroundStyle(palette.secondary)
+                .fontWeight(.semibold)
```
EricBAndrews: Unnecessary--clear from context

EricBAndrews (changes requested):
Code looks good!

The design is very nice too--my only note is that the pin button being in a separate column feels a bit off. Some alternative ideas:
- Color/fill the icon of the selected sort instead of the checkmark; we'd need a colorblind mode alternative, since fill is not available for scaled (or we find a new icon--megaphone, maybe?)
- Shade the background of the selected sort

Sjmarf:
> my only note is that the pin button being in a separate column feels a bit off. Some alternative ideas:

Sure, I can try those. How do you propose that the user changes whether an item is pinned or not? Via context menu? Currently, tapping the pin button toggles the pin state and tapping the sort mode itself selects it.

EricBAndrews:
Tapping the pin still seems like the way to go--we can also put it on a swipe action

Sjmarf:
How's this?



I'd rather avoid using only icon fill state to denote selection when we've got lots of different icons - it's much more difficult at a glance to see which is selected (to my eye, anyway).

EricBAndrews (approved):
LGTM! 🔥

Issue #1509, opened 2024-12-17
Static webp sometimes laggy


EricBAndrews:
Some static webp images are very slow in the image viewer

EricBAndrews:
Haven't noticed this since #1510. Tentatively marking closed.

PR #1510, opened 2024-12-17
Improve animated webp detection

EricBAndrews:
# Checklist
      - Hopefully deals with #1509

# Pull Request Information

This PR improves how webp images are handled. Previously, every webp would go through the `NukeWebpBridgeDecoder`, which would both decode the image and, if the image was animated, populate the `animated` field. This had multiple problems:
- The `BridgeDecoder` was scanning the entire data to find the `ANMF` header
- The `BridgeDecoder` was therefore very slow, even for static images, and because the check was so slow it couldn't defer to the standard Nuke decoder for static images
- The `BridgeDecoder` was poorly optimized for static images, since it was built to handle animated images
- The `BridgeDecoder` could throw a false positive if `ANMF` happened to appear in the body of the image's data

This PR implements a much more efficient method of checking whether a webp is animated; it now scans only the critical bytes, rather than the whole data. This allows the `NukeWebpBridgeDecoder` to check at `init` whether a webp is animated and defer to the more efficient Nuke default decoder for static images.

Hopefully using Nuke for the static webps fixes the lagginess, which I believe may have been caused by the custom decoding. It may just be good luck, but I haven't encountered lag on this branch.


Sjmarf (approved):
LGTM! 💯

PR #1519, opened 2024-12-17
Mod Mail Reports

### LINKED ISSUE #1504: Mod Mail: View Reports

Sjmarf:
Relies on [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/83).

Closes #1504. There is no content loader model yet - just some temporary code to display the first page of results. [Issue here](https://github.com/mlemgroup/mlem/issues/1520).

Reports look like this. In future the interaction bar will be customizable, but I haven't implemented this yet. [Issue here](https://github.com/mlemgroup/mlem/issues/1521).

On versions prior to 0.19.4, significantly less information is included in the report responses. This makes us unable to create `Post2`/`Comment2` models on those versions, and instead we have to fallback to `Post1`/`Comment1` models.

As a result, on 0.19.3 and below the reports instead look like this:

The biggest downgrade here is that there is no ellipsis menu from which you can take action on the content. To do that, you have to tap on the report to open the post/comment and take action on it from there. Now, this downgrade isn't actually a limitation of the missing data - we *could* still include those ellipsis menus, but doing so would require a significant amount of work and I don't think it's worth it.


Mlem/App/Views/Root/Tabs/Feeds/Feed Comments/FeedCommentView.swift
```diff
@@ -9,26 +9,48 @@ import Foundation
...
+    @Setting(\.postSize) var postSize
+    
+    let comment: any Comment
+    var overrideIsTiled: Bool?
+    @ViewBuilder var embeddedContent: () -> EmbeddedContent
     
-    let comment: Comment2
+    init(
+        comment: any Comment,
+        overrideIsTiled: Bool? = nil,
```
EricBAndrews: *nit*

It may be a little cleaner/more intuitive to simply have this be `overriddenSize: PostSize?`, rename `postSize` to `settingsPostSize`, and declare `var postSize: PostSize { overridenSize ?? postSize }` in place of the current computed `overridenSize`

EricBAndrews (changes requested):
Looks good overall! One minor code note, plus design discussion from Slack

Mlem/App/Views/Root/Tabs/Feeds/Feed Comments/FeedCommentView.swift
Sjmarf: Sure 👍

Sjmarf:
Tweaked the resolve button design pre 0.19.4.

EricBAndrews (approved):
LGTM! 🚀

Issue #720, opened 2023-10-14
Mark read when reply is visited?


Sjmarf:
When the user clicks on a reply in the inbox, the parent post is opened and scrolled to show the comment. When this happens, the reply is not marked as read. Should we mark the comment as read in this case? Or at least provide a setting that does so?

EricBAndrews:
This setting should also govern marking messages as read when entering the chat view

PR #1530, opened 2024-12-23
Send and reply to private messages

### LINKED ISSUE #622: Send message not from reply
Introduce a flow to send a private message to a user without already having a message to reply to

### LINKED ISSUE #1130: iMessage-like Private Messages View
We could have a "Private Message" action in `PersonView` that opens an iMessage-like view of messages with that user. This is possible to do with the API, though we'd have to be refreshing every few seconds in the background if we want new messages to pop up live. Related to #622. Apollo had this.

### LINKED ISSUE #1482: Reply to private messages
See also #622 which is not a regression (we didn't have that in v1)

Sjmarf:
Tap on a private message in the inbox to open the messages page. To start a new conversation, tap on "Send Message" in the ellipsis menu of `PersonView`.

Relies on [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/85).

Closes #622
Closes #1482
Closes #1130

Still to do:
- Allow editing Messages (#1529)
- Find a way to display read status (#1531). This could be tricky because you can technically have an older message be unread, a newer message be read.


Mlem.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -51,8 +51,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/mlemgroup/MlemMiddleware",
       "state" : {
-        "revision" : "dede345d6a2063d08fd94922cfafffb8e63359f0",
-        "version" : "0.55.0"
+        "branch" : "sjmarf/message-feed-loader",
+        "revision" : "6bfeb850a12c74869cff78b72d2719e80d97b36b"
```

Mlem/App/Utility/Extensions/Content Models/Message1Providing+Extensions.swift
```diff
@@ -75,11 +87,17 @@ extension Message1Providing {
...
        
-    func replyAction() -> BasicAction {
-        .init(
+    func replyAction(navigation: NavigationLayer) -> BasicAction {
+        let callback: (() -> Void)?
+        if let creator = creator_, api.canInteract {
+            callback = { navigation.push(.messageFeed(creator, focusTextField: true)) }
+        } else {
+            callback = nil
+        }
```
EricBAndrews: *nit*

```suggestion
    func replyAction(navigation: NavigationLayer) -> BasicAction {
        let callback: (() -> Void)? = nil
        if let creator = creator_, api.canInteract {
            callback = { navigation.push(.messageFeed(creator, focusTextField: true)) }
        }
```

Mlem/App/Views/Pages/MessageFeedView/MessageFeedView.swift
```diff
@@ -0,0 +1,220 @@
...
+    func content(person: any Person) -> some View {
+        ScrollViewReader { scrollProxy in
+            ScrollView {
+                if let feedLoader {
+                    LazyVStack(spacing: 0) {
+                        ForEach(feedLoader.items.reversed()) { message in
+                            if !message.deleted {
+                                bubbleView(message: message, feedLoader: feedLoader)
+                                    .id(message.id)
+                            }
```
EricBAndrews, line 90: Deleted messages should be filtered out at the middleware level--if a deleted post happens to be the loading threshold, this will break the infinite loading

Mlem/App/Views/Pages/MessageFeedView/MessageFeedView.swift
```diff
@@ -0,0 +1,220 @@
...
+            .onAppear {
+                if feedLoader == nil {
+                    feedLoader = .init(person: person, pageSize: 50)
+                    Task { @MainActor in
+                        do {
+                            try await feedLoader?.loadMoreItems()
+                        } catch {
+                            handleError(error)
+                        }
+                    }
```
EricBAndrews, line 126: What's the reason for this being here instead of `.init()`?

Mlem/App/Views/Pages/MessageFeedView/MessageBubbleView.swift
```diff
@@ -0,0 +1,127 @@
...
+        Group {
+            let blocks: [BlockNode] = .init(message.content)
+            if blocks.isSimpleParagraphs {
+                MarkdownText(blocks, configuration: message.isOwnMessage ? .inverted : .default)
+            } else {
+                Markdown(blocks, configuration: message.isOwnMessage ? .inverted : .default)
+            }
+        }
+        .padding(Constants.main.standardSpacing)
+        .padding(message.isOwnMessage ? .trailing : .leading, 7)
```
EricBAndrews, line 28: *nit*

This `message.isOwnMessage ? .trailing : .leading` phrase is repeated a _lot_--probably nicer to extend `Message` to provide this as something like `ownerAwareAlignment`

Mlem.xcodeproj/project.pbxproj
```diff
@@ -2725,7 +2749,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = "local build";
 				DEVELOPMENT_ASSET_PATHS = "\"Mlem/Preview Content\"";
-				DEVELOPMENT_TEAM = 8B9GNJW88W;
+				DEVELOPMENT_TEAM = V4747CQLU2;
```
EricBAndrews:
```suggestion
				DEVELOPMENT_TEAM = 8B9GNJW88W;
```

Mlem.xcodeproj/project.pbxproj
```diff
@@ -2745,7 +2769,7 @@
 					"@executable_path/Frameworks",
 				);
 				MARKETING_VERSION = 2.0;
-				PRODUCT_BUNDLE_IDENTIFIER = com.hanners.Mlem;
+				PRODUCT_BUNDLE_IDENTIFIER = com.sjmarf.mlem1;
```
EricBAndrews:
```suggestion
				PRODUCT_BUNDLE_IDENTIFIER = com.hanners.Mlem;
```

EricBAndrews (changes requested):
Nice work! We've wanted this UX _forever_, it's great to see it up and running.

A couple design notes:
- The UX around marking messages read is a bit odd (either swipe from the main feed then navigate in to reply or mark from the context menu of the chat view)--imo opening the chat should mark all messages in that chat read. Alternatively, the functionality should be available from some top-level gesture (tap or swipe) in the chat feed.
- Do we need the user link in the header of the sheet view? Since you get there from profile, that page is already one swipe away. We certainly don't need the ellipsis menu in that view--removing it would let us put the close button in that high-priority trailing spot
- Content to leave OOS, but since we have an auto-refreshing unread count already we should have an auto-fetch for messages in the current chat view. It's a bit inconsistent to see the unread counter increment and have to navigate back to get the new message. This should be pretty straightforward with a simple `.onChange` listener inside the chat view that picks up when the unread count changes and triggers a new load.
- It might be nice to keep the reply swipe action and just have it pop the chat sheet

Mlem/App/Views/Pages/MessageFeedView/MessageFeedView.swift
Sjmarf: The feedloader can't be made if we've only got a `PersonStub` - it requires a `Person`.

Mlem/App/Views/Pages/MessageFeedView/MessageFeedView.swift
Sjmarf: Deleted messages aren't returned by the API - this `if` statement is for hiding messages as soon as they are deleted by the user. This is the same setup that we use for deleting posts and comments. I don't think this breaks the loading threshold because the view is guaranteed to have already appeared before potentially being hidden by the `!message.deleted` if the user deletes the message.

Mlem/App/Views/Pages/MessageFeedView/MessageBubbleView.swift
Sjmarf: Yeah, I did try this but unfortunately half of the types are `Alignment` and the other half are `HorizontalAlignment`, so we'd need two computed properties. And in some cases I want `.leading` for `isOwnMessage`, and for others I need `.trailing`... so we'd either need 4 computed properties or some sort of `flip()` method of `Alignment` and `HorizontalAlignment` (which don't exist yet so I'd need to implement them). I thought it would be simpler to do it this way.

Sjmarf:
> The UX around marking messages read is a bit odd (either swipe from the main feed then navigate in to reply or mark from the context menu of the chat view)--imo opening the chat should mark all messages in that chat read. Alternatively, the functionality should be available from some top-level gesture (tap or swipe) in the chat feed.

Sure, I've made it mark all messages as read when the chat is opened. However, this is arguably inconsistent with the current behaviour of visiting a reply, which doesn't mark it as read. Is that something we want to change? #720

> Do we need the user link in the header of the sheet view? Since you get there from profile, that page is already one swipe away. We certainly don't need the ellipsis menu in that view--removing it would let us put the close button in that high-priority trailing spot

I've removed it 👍 

> Content to leave OOS, but since we have an auto-refreshing unread count already we should have an auto-fetch for messages in the current chat view. It's a bit inconsistent to see the unread counter increment and have to navigate back to get the new message. This should be pretty straightforward with a simple .onChange listener inside the chat view that picks up when the unread count changes and triggers a new load.

Makes sense 👍 I've made it refresh every 5 seconds.

> It might be nice to keep the reply swipe action and just have it pop the chat sheet

Maybe - that would be consistent with iMessage, but I'm cautious that it might give the impression that Lemmy has an actual "reply" system for individual messages (like iMessage does), which is not the case.

EricBAndrews:
> this is arguably inconsistent with the current behaviour of visiting a reply, which doesn't mark it as read

Fair point. I've added a comment to the linked issue--the behavior should be consistent between the two once we have that toggle available. Until then, sacrificing consistency for the better message UX feels like a good trade.

> it might give the impression that Lemmy has an actual "reply" system for individual messages (like iMessage does)

Yeah, that's a worthy concern. I think that without the UI cues that iMessage provides to support the reply system it's probably fine, but I'm content to leave OOS here and refine later.

Mlem/App/Views/Pages/MessageFeedView/MessageBubbleView.swift
EricBAndrews: Makes sense to me 👍

Mlem/App/Views/Pages/MessageFeedView/MessageFeedView.swift
EricBAndrews: Ah, I see 👍

EricBAndrews:
LGTM! 🔥

Sjmarf:
Needs approval 😄

EricBAndrews (approved):

PR #1539, opened 2024-12-24
Use MainActor for NavigationModel mutation

Sjmarf:
There's a TF crash related to mutating `NavigationModel.layers` on a background thread; this should hopefully fix it


EricBAndrews (approved):
Nice 👍

PR #1543, opened 2024-12-26
Accessible Mark Read Indicator

### LINKED ISSUE #1448: Accessible post read status indicator
We had this in v1, it needs to be copied over to v2

EricBAndrews:
# Checklist
      - closes #1448 

# Pull Request Information

Adds an accessible read post indicator. There are two options:
- Outline: unread posts are outlined. Thickness is configurable from 1 to 5px
- Checkmark: read posts show a checkmark next to the ellipsis menu

These markers should only be visible when the system-wide "differentiate without color" option is enabled.

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Settings/AccessibilitySettingsView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/AccessibilitySettingsView.swift
@@ -9,6 +9,8 @@ import SwiftUI
 
 struct AccessibilitySettingsView: View {
     
+    @Environment(\.accessibilityDifferentiateWithoutColor) var differentiateWithoutColor
+    
     @Setting(\.readPostIndicator) var readPostIndicator
     @Setting(\.readOutlineThickness) var readOutlineThickness
     
@@ -49,6 +51,8 @@ struct AccessibilitySettingsView: View {
                         }
                     }
                 }
+            } footer: {
+                Text("These settings control how Mlem behaves when the system-wide \"Differentiate Without Color\" option is enabled.")
             }
         }
     }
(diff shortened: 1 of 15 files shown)


Mlem/App/Views/Root/Tabs/Settings/AccessibilitySettingsView.swift
```diff
@@ -0,0 +1,59 @@
...
+                        Slider(
+                            value: $readBarThicknessSlider,
+                            in: 1 ... 5,
+                            step: 1
+                        ) {
+                            Text("Outline Thickness")
+                        } minimumValueLabel: {
+                            Text("1")
+                        } maximumValueLabel: {
+                            Text("5")
```
Sjmarf:
```suggestion
                            Text(verbatim: "1")
                        } maximumValueLabel: {
                            Text(verbatim: "5")
```

Mlem/Localizable.xcstrings
```diff
@@ -96,6 +96,12 @@
     },
     "%lld on your instance" : {
 
+    },
+    "1" : {
+
+    },
+    "5" : {
+
```
```suggestion
```

Sjmarf (changes requested):
Nice, the outline option works really well :D

I don't think we should have a "None" option in the accessibility setting. Did we have that in v1? Imo, if the user has "differentiate without color" enabled then we should accommodate that without requiring the user to change any settings manually, which they are currently required to do. Might defaulting to "checkmark" and removing the "none" option be better?

EricBAndrews:
> if the user has "differentiate without color" enabled then we should accommodate that without requiring the user to change any settings manually, which they are currently required to do

Absolutely agreed, I've updated the default to checkmark. Nice catch!

> I don't think we should have a "None" option in the accessibility setting

I think we should keep that as an option because I don't want to make assumptions about why a user has the system-wide option enabled--e.g., a user may have it on because they are red/green colorblind but have no trouble differentiating `.primary` and `.secondary`, in which case they may not want the added visual clutter of the accessible indicator.

Sjmarf (approved):
Works for me 👍

Issue #1488, opened 2024-12-07
Keyword filters on/off toggle


Sjmarf:
It isn't possible to turn off keyword filtering temporarily - you have to delete all your keywords, which isn't ideal.

Child of #1175

ghost:
I was about to comment this, right now the temporary solution is to keep a JSON backup to pull out when needed.

PR #1553, opened 2024-12-28
Import/Export Filtered Keywords

### LINKED ISSUE #1533: Import/Export keyword filters

EricBAndrews:
# Checklist
      - closes #1533 

# Pull Request Information

This PR enhances the existing import/export settings feature to also include keyword filters. This is done instead of making filters independently exportable because filters are presented as settings and so it doesn't make much sense to handle them as a special case; this also makes the common case of moving settings from one device to another smooth.

Since filters are stored in a file and other settings in AppStorage, this requires some slight architectural rearranging to make smooth:
- All filter-related persistence now happens in a private extension of `PersistenceRepository` available only to `FiltersTracker`. This makes `FiltersTracker` a reliable source of truth for the current persisted filters state.
- When settings are exported, the `CodableSettings` object is instantiated with the current state of `FiltersTracker`
- When settings are imported, `FiltersTracker` is updated with the imported keywords and the underlying file is automatically updated

### DIFF
--- a/Mlem/App/Configuration/User Settings/CodableSettings.swift
+++ b/Mlem/App/Configuration/User Settings/CodableSettings.swift
@@ -11,6 +11,7 @@ import UIKit
 
 /// Mirror of Settings but without any AppStorage complexity and fully optionalized.
 struct CodableSettings: Codable {
+    // MARK: Settings mirrored in AppStorage
     var a11y_markReadType: String // TODO: pending a11y mark read
     var a11y_readBarThickness: Int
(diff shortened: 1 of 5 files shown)


Sjmarf (approved):
Nice! Works great 😄 

Incorporating the filters into the wider settings export works great, but slightly breaks [the workaround](https://github.com/mlemgroup/mlem/issues/1488#issuecomment-2538352596) someone was using to toggle keyword filters on/off. Imo we should implement #1488 before 2.0 release to cover this.

PR #1565, opened 2024-12-30
"Sign Out" button in account switcher context menu

### LINKED ISSUE #1562: Add "Sign out" action to context menus in account list settings
The swipe action doesn't work when "Swipe to navigate" is on; we should have an alternative

Sjmarf:
Closes #1562


EricBAndrews (approved):
Nice 👍

Issue #1196, opened 2024-08-07
Community categories in Search (like Apple Music)


Sjmarf:
Instead of having a list of communities in the search homepage, we could have several category buttons akin to Apple music:
<img src="https://github.com/user-attachments/assets/718689a4-db4d-4caf-a0fe-ad0d1ab54008" width=50%>

We could have "Top Communities" and "Top Local Communities" to replace the existing community list. We could also have a "Trending communities" section - this could be done using Lemmy's trending communities API endpoint, though in my opinion their algorithm kinda sucks. The (imo better) alternative would be to pull data from !trendingcommunities@feddit.nl. 

We could also have a number of preset lists, which could be fetched from a server (#1313) so that we can edit them without updating Mlem. The downside of doing this is that we'd have to manage the lists ourselves. There may exist some service that categorises communities already, in which case we could utilise that existing service instead.

Here's a post with some communities sorted into categories. https://sh.itjust.works/post/1777994

Sjmarf:
Piefed (#1824) has a list of communities sorted by category, but it's quite sparse and there's not yet an API endpoint for accessing it.

PR #1571, opened 2025-01-03
Recent Searches

### LINKED ISSUE #1556: Recent Searches list

Sjmarf:
> [!note]
> Relies on [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/91).

Closes #1556

Adds search history for community, user and instance search. Search history can be cleared or turned off completely under Settings -> Account -> Local Settings.

Still to do:
- Allow removing individual items from search history (#1570)
- Post search history (#1569)

## Serializing entire models vs just IDs

In Mlem v1, we stored just the IDs of the recently searched objects and sent individual `GET` requests for each object when the account was made active. This could get quite spammy when the search history fills up (the maximum here is 30 items; in Mlem v1 it was 15 iirc). 

I think it's better to instead serialize *all* of the data we need to create `Community2`/`Person2` models when the app opens. I've done this by allowing `Community2`/`Person2` to be turned back into `ApiCommunityView`/`ApiPersonView`. Those API models are stored in a JSON file and used to initialize `Community2`/`Person2` when the app opens.

One downside of doing this is that if one of the recent searches changes state (e.g. the Mlem user subscribes to a community on another client), that change won't be reflected in the recent search list until the user taps on the navigation link and the model updates. There are potentially ways that this could be worked around (e.g. updating the `subscribed` state of all cached communities when the subscription list is refreshed), but no complete fix.

## Recording all visits

The `VisitHistory` class keeps track of which communities/users/instances have been visited. It keeps track of *all* visits to those entity pages, not just those from the search tab. This is because, in future, I want to show recently visited communities in the community search sheet (such as when choosing a community in the post editor) when the user hasn't typed anything yet.


Mlem.xcodeproj/project.pbxproj
```diff
@@ -3070,8 +3086,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware";
 			requirement = {
-				kind = upToNextMinorVersion;
-				minimumVersion = 0.59.0;
+				branch = "sjmarf/recent-searches";
+				kind = branch;
```

Mlem/App/Globals/Definitions/PersistenceRepository.swift
```diff
@@ -155,13 +162,13 @@ class PersistenceRepository {
...
-        try await save(settings, to: PersistencePath.userSettings.appendingPathComponent(name, conformingTo: .json))
+    func saveUserSettings(_ settings: CodableSettings, for account: any Account) async throws {
+        try await save(settings, to: PersistencePath.accountSettings(for: account))
     }
     
     /// Loads given user settings, if present
-    func loadUserSettings(name: String) -> CodableSettings? {
-        load(CodableSettings.self, from: PersistencePath.userSettings.appendingPathComponent(name, conformingTo: .json))
+    func loadUserSettings(for account: any Account) -> CodableSettings? {
+        load(CodableSettings.self, from: PersistencePath.accountSettings(for: account))
```
EricBAndrews: These should be renamed `loadAccountSttings` and `saveAccountSettings`.

As a side note, I had originally intended these to be a way for users to save/restore custom named settings profiles without needing to manually muck around with files (hence the `String` name). That still seems like a useful (if low priority) feature. Those old methods weren't used and are trivial to reintroduce so it's fine to remove them here, but we should keep the function names tidy in case we want to add that functionality back in.

EricBAndrews (changes requested):
Nice work! Everything works very nicely 😁 

Some of the behavior is a bit unexpected:
- "Recent Searches" doesn't really show searches, but recently visited search results--just renaming it to "Recents" would smooth that over
- Navigating to a suggested search result doesn't put it into recents, which feels inconsistent

Sjmarf:
> "Recent Searches" doesn't really show searches, but recently visited search results--just renaming it to "Recents" would smooth that over

The title reads "Recently Searched" because that's what Apple uses in Music/Podcasts/TV. IMO we should keep it that way for consistency. "Recently Searched" reads like "recently searched for" to me, which makes sense. I don't mind "Recents" either though so if you prefer that I'm OK with it

> Navigating to a suggested search result doesn't put it into recents, which feels inconsistent

Are you referring to tapping on a recent search or tapping one of the results on the Search homepage?

Mlem/App/Globals/Definitions/PersistenceRepository.swift
Sjmarf: Yeah, I noticed that they weren't used anywhere. I'll rename them and leave them in for now 👍

EricBAndrews:
> The title reads "Recently Searched" because that's what Apple uses in Music/Podcasts/TV.

Hmm, I guess they do. It still feels like a misnomer to me but not enough to break from consistency with Apple patterns--we can leave as Recently Searched.

> Are you referring to tapping on a recent search or tapping one of the results on the Search homepage?

One of the results on the Search homepage.

Sjmarf:
Hmm, I'm not sure whether we should be making the homepage results appear in the list. The title of the recents section being "recently searched" implies that it contains only items that you've "searched", which doesn't apply to the homepage items. 

Also, if you want to go back to a homepage items that you've visited recently I think it's more likely that you'll go straight there the same way that you did before rather than opening the search bar despite not actually wanting to search anything.

Eventually I'd like to replace the homepage list with a grid of categories instead, like Apple Music does (#1196).

EricBAndrews:
Sure, that reasoning makes sense to me

EricBAndrews (approved):

PR #1572, opened 2025-01-04
Fix flair assertion failure

Sjmarf:
The `assert` on `Person1Providing+Extensions` line 41 would occasionally fail. 

This seems to be because somehow the wrong `Comment` models ends up being put in the `commentContext` environment for `FullyQualifiedLabelView`. Looking at the code, this mismatch shouldn't be possible - I think it's a SwiftUI bug. I wasn't able to directly fix the issue, so instead have made `FullyQualifiedLabelView` ignore the `commentContext` if the comment creator doesn't match the `person` it was given.

The wrong environment appears to only occur momentarily when the view appears, and then fixes itself shortly after. So I don't think the bug causes any visual issues.


EricBAndrews (approved):
LGTM! 🚀

Issue #1470, opened 2024-12-01
Inline images sometimes fail to load, showing placeholder


Sjmarf:
This is an issue with state update I think - the inline images load fine if they're already cached.

Sjmarf:
NOT fixed. I've got an idea of what needs changing in LemmyMarkdownUI to fix this; I'll get on it at some point

PR #1576, opened 2025-01-05
Fix user-created comments not always being inserted into the comment tree

Sjmarf:
Also added a possible fix for #1470, haven't spent that long trying to reproduce it though. I'll close that issue later if I think it's fixed


EricBAndrews (approved):
Nice 👍

PR #1579, opened 2025-01-07
Insert new posts into feed

### LINKED ISSUE #1456: When new post created, insert the post at the top of the current feed
The post should only be inserted if it's meant to be in the current feed. We'd have to implement some logic to determine this. Alternatively, we could just refresh the current feed (without clear) and let the backend determine that.

In v1, I think (?) we just inserted into the feed regardless of whether it belonged there or not, which isn't ideal UX.

EricBAndrews:
> [!NOTE]
> Depends on [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/93)

# Checklist
      - closes #1456

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 55;
+	objectVersion = 60;
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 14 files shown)


Mlem.xcodeproj/project.pbxproj
```diff
@@ -3090,19 +3090,19 @@
...
-		CD4A6BCC2D1A1E4E0046468E /* XCRemoteSwiftPackageReference "MlemMiddleware" */ = {
+		CD64A9182CA37D63007CA7E6 /* XCRemoteSwiftPackageReference "Gifu" */ = {
 			isa = XCRemoteSwiftPackageReference;
-			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware";
+			repositoryURL = "https://github.com/kaishin/Gifu";
 			requirement = {
-				kind = upToNextMinorVersion;
-				minimumVersion = 0.60.0;
+				branch = master;
+				kind = branch;
```

Mlem/App/Views/Pages/PostEditor/PostEditorTargetView.swift
```diff
@@ -160,4 +160,30 @@ class PostEditorTarget: Identifiable {
...
+            Task { @MainActor in
+                withAnimation {
+                    communityFeedLoader.prependItem(post)
+                }
+            }
+            return
+        }
+        
+        if let personContentFeedLoader = feedLoader as? PersonContentFeedLoader,
+           personContentFeedLoader.userId == account.id {
```
Sjmarf: It might be worthwhile comparing the `ApiClient` here too - theoretically we could be comparing two different users with the same ID on different instances and get a false positive (though that is unlikely).

Sjmarf (approved):
Looks good! One minor point

PR #1585, opened 2025-01-10
Add and Remove Administrators

### LINKED ISSUE #965: Add and remove administrators

EricBAndrews:
> [!NOTE]
> Depends on [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/94)

# Checklist
      - closes #965 

# Pull Request Information

Administrators can now be appointed from the "Administration" tab of the instance view or appointed/removed from `Person1Providing` menus.

Also updates the `SearchSheetView` so it can be opened with a filter applied. Note that this **does not work** on locally hosted instances (at least, not on mine). To test it properly, you can remove the admin check at `InstanceView:139` and open the sheet using the "Add Administrator" button on a real instance.

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 60;
+	objectVersion = 55;
 	objects = {
 
 /* Begin PBXBuildFile section */
@@ -331,7 +331,6 @@
 		CD3153072C38421B00BC5FBE /* View+LoadFeed.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD3153062C38421B00BC5FBE /* View+LoadFeed.swift */; };
 		CD317D4C2BE97FED008F63E2 /* Palette.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD317D4B2BE97FED008F63E2 /* Palette.swift */; };
 		CD317D4F2BE983ED008F63E2 /* MonochromePalette.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD317D4E2BE983ED008F63E2 /* MonochromePalette.swift */; };
-		CD32680C2D308CEC0072AC1C /* MlemMiddleware in Frameworks */ = {isa = PBXBuildFile; productRef = CD32680B2D308CEC0072AC1C /* MlemMiddleware */; };
 		CD332D792CA7175500A53988 /* PlayButton.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD332D782CA7175200A53988 /* PlayButton.swift */; };
 		CD332D7C2CA71E6F00A53988 /* GifView.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD332D7B2CA71E6E00A53988 /* GifView.swift */; };
 		CD332D7E2CA7486000A53988 /* String+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD332D7D2CA7485D00A53988 /* String+Extensions.swift */; };
@@ -392,6 +391,7 @@
 		CD9CFA302C2E22F600739BBC /* FeedDescription.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD9CFA2F2C2E22F600739BBC /* FeedDescription.swift */; };
 		CD9CFA372C306DAB00739BBC /* PostGridView.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD9CFA362C306DAB00739BBC /* PostGridView.swift */; };
 		CD9D243D2CC1DF59006E5F3F /* AccountType.swift in Sources */ = {isa = PBXBuildFile; fileRef = CD9D243C2CC1DF55006E5F3F /* AccountType.swift */; };
+		CDA38A6F2D31C4E6000C5B20 /* MlemMiddleware in Frameworks */ = {isa = PBXBuildFile; productRef = CDA38A6E2D31C4E6000C5B20 /* MlemMiddleware */; };
 		CDA683F82C77E577000C4486 /* NsfwBlurBehavior.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDA683F72C77E577000C4486 /* NsfwBlurBehavior.swift */; };
 		CDAA02DB2C810DB200D75633 /* Calendar+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDAA02DA2C810DB200D75633 /* Calendar+Extensions.swift */; };
 		CDAA02DD2C81792500D75633 /* SolarizedPalette.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDAA02DC2C81792500D75633 /* SolarizedPalette.swift */; };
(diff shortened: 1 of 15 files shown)


Mlem.xcodeproj/project.pbxproj
```diff
@@ -3098,12 +3098,12 @@
...
 			};
 		};
-		CDB190B02D2CB4090044B05B /* XCRemoteSwiftPackageReference "MlemMiddleware" */ = {
+		CDA38A6D2D31C4DD000C5B20 /* XCRemoteSwiftPackageReference "MlemMiddleware" */ = {
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware";
 			requirement = {
-				kind = upToNextMajorVersion;
-				minimumVersion = 0.61.0;
+				branch = "eric/add-admin";
```

Mlem/App/Utility/Extensions/Content Models/Person1Providing+Extensions.swift
```diff
@@ -197,4 +204,26 @@ extension Person1Providing {
...
+                    handleError(error)
+                }
+            }
+        }
+        
+        return .init(
+            id: "addAdmin\(uid)",
+            appearance: .addAdmin(isOn: isOn),
+            callback: callback
+        )
```
Sjmarf: IMO both adding and removing admins should have confirmation dialogues. It's not entirely unlikely that someone might accidentally tap the wrong user in the `SearchSheetView`, so I think we should display a confirmation dialogue after that.

Mlem/App/Views/Shared/Search/Searchable.swift
```diff
@@ -8,25 +8,30 @@
...
+    static func search(api: ApiClient, query: String, page: Int, limit: Int, filter: ApiListingType) async throws -> [Community2] {
+        try await api.searchCommunities(query: query, page: page, limit: limit, filter: filter)
     }
 }
 
 extension Person2: Searchable {
-    static func search(api: ApiClient, query: String, page: Int, limit: Int) async throws -> [Person2] {
-        try await api.searchPeople(query: query, page: page, limit: limit)
+    static func search(api: ApiClient, query: String, page: Int, limit: Int, filter: ApiListingType) async throws -> [Person2] {
+        try await api.searchPeople(query: query, page: page, limit: limit, filter: filter)
```
Sjmarf, line 22: Btw, filtering the person search results by `.local` is only available on 0.19.4 and newer. Before that I believe it just shows all users and ignores the `ApiListingType`. On earlier versions, do we want to make it obvious in the UI somewhere that only local users can be added? Might not be worth the time though, considering this is a feature only admins will be using

Sjmarf (changes requested):
Code looks good! A couple of design points, plus some questions in the other PR

Mlem/App/Views/Pages/Instance/InstanceView.swift
```diff
@@ -134,6 +135,16 @@ struct InstanceView: View {
...
             }
+            
+            if appState.firstApi.isAdmin {
+                Button("Add Administrator", systemImage: Icons.add) {
+                    navigation.openSheet(.personPicker(filter: .local) { person in
+                        Task { await addAdmin(person, added: true) }
+                    })
+                }
+                .buttonStyle(.capsule)
+                .padding(.top, Constants.main.halfSpacing)
```
Sjmarf: This padding is inconsistent with the rest of the list; IMO we should keep everything equally spaced

```suggestion
```

EricBAndrews: I did it this way to follow the profile "New Post" button--that's differentiated with a full space even when the content below is tighter.

Screenshot 2025-01-11 at 2 46 29 PM

Without the spacing feels a little tight to me--the proximity makes the different corner radii stand out in a way that feels just a bit awkward:

Screenshot 2025-01-11 at 2 58 13 PM

It looks nicer in the same spot as "New Post"? I initially didn't do that because it feels like a pretty low-priority button for that spot, but on the other hand an admin is probably only going here to manipulate the admin list so perhaps it's higher priority than I initially thought.

Screenshot 2025-01-11 at 2 59 45 PM

Sjmarf: Hmm, good point. What if we kept it at the bottom with the wider spacing, and also make the spacing between the modlog button and the first administrator row wider? That might make it look more symmetrical. 

Alternatively we could add an "Administrators" heading above the list of admins and put the + button inline with the heading, like:

```
[Modlog                             ]

Administrators (5)                (+)
[Admin1                             ]
[Admin2                             ]
[Admin3                             ]
```

EricBAndrews: I like changing the modlog spacing--that also has the added bonus of making modlog a little more prominent/easy to tap.

Screenshot 2025-01-11 at 3 20 53 PM

Sjmarf: Nice, LGTM 👍

Mlem/App/Views/Shared/Search/Searchable.swift
EricBAndrews: Nice catch--I've added an informative failure toast, which should be sufficient for an admin.

Sjmarf: Nice 👍

Mlem/App/Views/Pages/Instance/InstanceView.swift
```diff
@@ -130,20 +133,24 @@ struct InstanceView: View {
...
-                }
-                .buttonStyle(.capsule)
-                .padding(.top, Constants.main.halfSpacing)
+                Button("Add Administrator", systemImage: Icons.add, action: openAddAdminSheet)
+                    .buttonStyle(.capsule)
+                    .padding(.bottom, Constants.main.halfSpacing)
+                    .confirmationDialog("", isPresented: $showingConfirmation) {
+                        Button("Yes", action: addNewAdmin)
+                    } message: {
+                        Text("Really appoint \(newAdmin?.displayName ?? "this user") as administrator of \(instance.displayName)?")
```
Sjmarf: For localization we should avoid unwrapping to a fallback in this way because it might not work with all languages

```suggestion
    if let displayName = newAdmin?.displayName {
        Text("Really appoint \(displayName) as an administrator of \(instance.displayName)?")
    } else {
        Text("Really appoint this user as an administrator of \(instance.displayName)?")
    }
```

Mlem/App/Views/Pages/Instance/InstanceView.swift
```diff
@@ -130,20 +133,24 @@ struct InstanceView: View {
...
-                    navigation.openSheet(.personPicker(filter: .local) { person in
-                        Task { await addAdmin(person, added: true) }
-                    })
-                }
-                .buttonStyle(.capsule)
-                .padding(.top, Constants.main.halfSpacing)
+                Button("Add Administrator", systemImage: Icons.add, action: openAddAdminSheet)
+                    .buttonStyle(.capsule)
+                    .padding(.bottom, Constants.main.halfSpacing)
+                    .confirmationDialog("", isPresented: $showingConfirmation) {
```
Sjmarf: The title of this dialog isn't displayed by default, but I'm not sure whether accessibility might rely on it. We should either provide an actual title here, or replace `""` with `String()` to avoid it being localized.

Mlem/App/Utility/Extensions/Content Models/Person1Providing+Extensions.swift
```diff
@@ -223,6 +223,9 @@ extension Person1Providing {
         return .init(
             id: "addAdmin\(uid)",
             appearance: .addAdmin(isOn: isOn),
+            confirmationPrompt: isOn
+            ? "Really remove \(displayName) from administrators of \(instance.displayName)?"
+            : "Really appoint \(displayName) as administrator of \(instance.displayName)?",
```
Sjmarf: Wording suggestion
```suggestion
            ? "Really remove \(displayName) as an administrator of \(instance.displayName)?"
            : "Really appoint \(displayName) as an administrator of \(instance.displayName)?",
```

Sjmarf (approved):
A few wording suggestions, otherwise LGTM!

PR #1595, opened 2025-01-11
Fix second render on sheet dismissal

### LINKED ISSUE #1492: Interactive sheet dismissal triggers second render
The body of `NavigationLayerView` gets rendered a second time when swiping a sheet down. This may be the root cause of #1354. The behavior does not manifest when dismissing using the button.

### LINKED ISSUE #1593: Image viewer close button flashes when dismissing with swipe gesture

Sjmarf:
Closes #1492
Closes #1593

Because the view body to which the `.sheet` modifier was attached accessed properties from `NavigationModel`, changing a `NavigationModel` property would cause the view body to be evaluated again. I've refactored it to avoid being dependent on `NavigationModel`.

### DIFF
--- a/Mlem/App/Views/Root/ContentView.swift
+++ b/Mlem/App/Views/Root/ContentView.swift
@@ -63,7 +63,10 @@ struct ContentView: View {
                         try await (appState.firstSession as? UserSession)?.unreadCount?.refresh()
                     }
                 }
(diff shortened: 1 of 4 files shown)


EricBAndrews (approved):
LGTM! 🔥 

Nice work on this one, that was a nasty little bug

PR #1609, opened 2025-01-16
Fix bubble picker crash

Sjmarf:
We've had a few TF crashes related to this (most recently from Nocturne on Jan 6).

I can't see what causes this so have just added an `if` statement to prevent a crash. If you see a better solution lmk.


EricBAndrews (approved):
LGTM! 🔥

PR #1612, opened 2025-01-16
New Media View

### LINKED ISSUE #1219: Limit link preview images height in Large mode
We currently limit the height of post images in Large mode. If a post exceeds this height, it is shrunk to the maximum height whilst maintaining aspect ratio. We should also do this for website preview images.

Example of the current (incorrect) behavior:

EricBAndrews:
# Checklist
      - closes #1219 

# Pull Request Information

Replaces `DynamicMediaView` and `LargeImageView` with `MediaView`. `MediaView` is built on an opt-in functionality paradigm; by default it returns a plain image for the URL, and the caller may enable additional complexities as needed.

This architecture offers several advantages over the previous one:
- Image behavior is now obvious at the point where `MediaView` is invoked
- Removes the "bucket brigade" anti-pattern that was emerging in the old image view hierarchy
- Improves code consistency (e.g., previously the post used `LargeImageView`, but the image viewer used `DynamicMediaView` directly to bypass default behaviors on `LargeImageView`)
- Enables granular control over image behavior without bolt-on default behavior overrides

The new `MediaView` also has some new features:
- Aspect ratio can be vertically bounded with the `verticalAspectRatioBounds` parameter (previously this was handled by tacking `.aspectRatio` to one of the old views)
- `ContentMode` can now be specified; previously only `.fit` was allowed.

`WebsitePreviewView` has been updated to use `verticalAspectRatioBounds`, closing #1219. It uses `.fill` to always take up the full width, since `.fit` looks bad with the website preview complex.


Mlem/App/Views/Shared/Images/Core/MediaView+Views.swift
```diff
@@ -0,0 +1,183 @@
...
+            playing: playing,
+            aspectRatio: uiImage.verticallyBoundedAspectRatio(bounds: aspectRatio),
+            contentMode: contentMode)
+    }
+    
+    @ViewBuilder
+    var nsfwOverlay: some View {
+        if enableNsfwBlur {
+            NsfwOverlay(blurred: $blurred)
+        }
```
Sjmarf, line 87: There was previously a short animation when blurring/unblurring an image which doesn't seem to be working anymore

Mlem/App/Views/Shared/Images/Wrappers/ThumbnailImageView.swift
```diff
@@ -8,6 +8,7 @@
 import Foundation
 import MlemMiddleware
 import SwiftUI
+import _QuickLook_SwiftUI // https://github.com/swiftlang/swift/issues/73048
```
Sjmarf: I believe frameworks formatted like `_QuickLook_SwiftUI` are imported automatically if you import both `QuickLook` and `SwiftUI`. Would importing `QuickLook` here instead also work?

```suggestion
import QuickLook
```

Mlem/App/Views/Shared/WebsitePreviewView.swift
```diff
@@ -59,24 +57,22 @@ struct WebsitePreviewView: View {
...
-                            .background {
-                                Capsule()
-                                    .fill(.regularMaterial)
-                                    .overlay(Capsule().fill(palette.background.opacity(0.25)))
-                            }
-                            .padding(Constants.main.halfSpacing)
-                    }
+                MediaView(
+                    url: thumbnailUrl,
+                    verticalAspectRatioBounds: .init(width: 4, height: 5),
```
Sjmarf: Imo we should use a shorter aspect ratio here - the website preview as a whole (including the site title) ends up taller than 4:5 which looks too tall to me. Maybe 1:1 would be better?

Mlem/App/Views/Shared/Images/Core/MediaView.swift
```diff
@@ -2,58 +2,142 @@
...
-                    }
+        content
+            .blur(radius: blurValue, opaque: true)
+            .overlay(animationControlOverlay)
+            .overlay(nsfwOverlay)
+            .overlay(developerOverlay)
+            .overlay(errorOverlay)
+            .clipShape(.rect(cornerRadius: cornerRadius))
+            .withContextMenu(menuContent: contextMenuContent, isEnabled: enableContextMenu)
+            .gesture(TapGesture().onEnded(tapActions), isEnabled: onTapActions != nil)
```
Sjmarf: This causes the image viewer to not open when no additional `onTapActions` are specified (e.g. in inline markdown images).

```suggestion
            .gesture(TapGesture().onEnded(tapActions), isEnabled: (onTapActions != nil) || enableImageViewer)
```

Sjmarf (changes requested):
Nice, this is super clean 😄 Found a few bugs

Mlem/App/Utility/Extensions/MarkdownConfiguration+Extensions.swift
```diff
@@ -68,9 +68,12 @@ extension MarkdownConfiguration {
 
 private func imageView(_ inlineImage: InlineImage, shouldBlur: Bool) -> AnyView {
     AnyView(
-        LargeImageView(url: inlineImage.url, shouldBlur: shouldBlur)
-            .aspectRatio(CGSize(width: 1, height: 1.2), contentMode: .fill)
-            .frame(maxWidth: .infinity)
+        MediaView(
+            url: inlineImage.url,
+            verticalAspectRatioBounds: .init(width: 4, height: 5),
```
Sjmarf, line 73:
```suggestion
            verticalAspectRatioBounds: .init(width: 4, height: 5),
            cornerRadius: Constants.main.mediumItemCornerRadius,
```

Mlem/App/Views/Shared/Images/Core/MediaView+Views.swift
EricBAndrews: Nice catch!

Fixing this I also realized the `blurValue` I was using was scaling based on underlying image size, not visual size--I've reverted that back to `.dynamicBlur`.

Mlem/App/Views/Shared/WebsitePreviewView.swift
EricBAndrews: Works for me 👍

Mlem/App/Views/Shared/Images/Core/MediaView.swift
EricBAndrews: Nice catch!

EricBAndrews:
Implementation note:

I've updated the behavior of blurred animated content so that unblurring automatically plays. There are two reasons for this:
- Smoother UX (one tap instead of two in the common case)
- The material of the `PlayButton` and the blur animation create some nasty artifacts together. I've added some code to `AnimationControlLayer` to fix this on the blur case by hiding the play button immediately when blurring so that it isn't present when the blur animation happens. This doesn't work in the unblur case, however, because the play button would need to be shown after the unblur animation completes and SwiftUI's timing code is not that precise; the play button either appears slightly before, causing artifacts, or slightly after, which looks bad.

Sjmarf (approved):
Nice, LGTM 🔥

PR #1613, opened 2025-01-17
Unread count settings

### LINKED ISSUE #1597: Include reports and registration applications in inbox unread count

Sjmarf:
Added a setting to control which notifications are included in the Inbox tab badge. We had this in v1


Mlem.xcodeproj/project.pbxproj
```diff
@@ -3122,8 +3134,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware";
 			requirement = {
-				kind = upToNextMinorVersion;
-				minimumVersion = 0.65.0;
+				branch = "sjmarf/moderator-unread-count";
+				kind = branch;
```
EricBAndrews: reminder

EricBAndrews (changes requested):
This works really nicely!

A couple notes:
- Mod mail tabs don't show unread count
- IMO there should be a nav link to the badge settings page from Inbox settings as well--that would also be a reasonable place for a user to look for them
- Inbox badge doesn't update properly when switching from an admin account to a non-admin, though this may be because mod mail feed loading isn't implemented yet

Sjmarf:
Fixed 👍

Sjmarf:
Actually hold on a min. Found a bug

Sjmarf:
Fixed 👍

EricBAndrews (approved):
LGTM! 🔥

PR #1617, opened 2025-01-17
Image Controls

EricBAndrews:
# Checklist
      - closes #issue
      - progress towards #issue

# Pull Request Information

Replaces the image viewer context menu with a control layer.

Intended behavior:
- Control layer is shown when viewer first opens
- For non-animated media: control layer appears/disappears on tap
- For animated media: 
  - When playing: control layer appears/disappears on tap outside of media's frame (i.e., in the black margins)
  - When not playing: control layer appears/disappears on tap anywhere but the play button

This PR also removes the `ProgressView` from `AnimationControlLayer`. Media loading is almost always fast enough that the `ProgressView` is just visual clutter, and in rare cases a gif advertises itself as "animated" but isn't and leaves the `ProgressView` spinning on top of it (I don't have a repro example handy, but I have seen it a couple times).

### DIFF
--- a/Mlem/App/Views/Pages/ImageViewer.swift
+++ b/Mlem/App/Views/Pages/ImageViewer.swift
@@ -22,6 +22,7 @@ struct ImageViewer: View {
     @State var offset: CGFloat = 0
     @State var isDismissing: Bool = false
     @State var opacity: CGFloat = 0
+    @State var showingControlLayer: Bool = true
     
     init(url: URL) {
         var components = URLComponents(url: url, resolvingAgainstBaseURL: false)!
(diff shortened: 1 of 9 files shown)


Mlem/App/Views/Pages/ImageViewer.swift
```diff
@@ -65,23 +86,93 @@ struct ImageViewer: View {
...
+                }
+                .padding(Constants.main.standardSpacing)
+                .contentShape(.rect)
+            }
+            .padding(.horizontal, Constants.main.halfSpacing)
+            .background {
+                Capsule().fill(.ultraThinMaterial)
+                    .environment(\.colorScheme, .dark)
+            }
+            .offset(y: controlOffset)
```
Sjmarf: I wonder if also adding an opacity animation for the controls would look nice?

Mlem/App/Views/Pages/ImageViewer.swift
```diff
@@ -31,32 +50,34 @@ struct ImageViewer: View {
...
+        .onChange(of: showControls) {
+            withAnimation(.easeOut(duration: duration)) {
+                controlOffset = showControls ? 0 : 100
+            }
+        }
+        .onChange(of: isZoomed) {
+            if isZoomed {
+                showControls = false
+            }
+        }
```
Sjmarf: The controls only start to hide when I finish the zoom gesture - it would be nicer if they were hidden as soon as the gesture starts imo.

Sjmarf (changes requested):
Looks good! 😄 A couple of design questions

EricBAndrews:
Per Slack, controls now (should) work as follows:
- On single tap gesture: hide/show with opacity animation
- On zoom in: hide with opacity + slide animation the moment the zoom starts
  - If the user has tapped to show controls while the image is already zoomed, controls should not auto-hide (unless zoom scale resets to 1.0 and the user zooms back in from there)
- On zoom out: show with opacity + slide animation
  - On double tap gesture, this happens immediately
  - On pinch gesture, this happens when zoom scale hits 1.0; this prevents the controls from popping in and out if the user is pinching around an image
- On interactive dismiss: slide behavior unchanged, fade now matches slide-and-fade (i.e., fades to 0 at `maxControlOffset`)

I've also refactored the control layer view code into `ImageViewer+Views`.

Sjmarf (approved):
Nice, works beautifully 😄

PR #1622, opened 2025-01-18
Image Control Tweaks

EricBAndrews:
Minor changes to the image controls:
- Fixed a bug where zooming in and then dragging would cause the controls to appear
- When interactively dismissing, the controls now move at 1/3 the speed of the image--this feels smoother to me


Sjmarf (approved):
LGTM 👍

PR #1627, opened 2025-01-19
Move Accessible Read Indicator Setting to Post Settings

Sjmarf:
I think this is better because it means that we have one less link on the main settings page, and users who want to change this setting will still be able to find it in Post Settings pretty easily. The setting is only shown if Differentiate Without Color is on.

If we end up adding more accessibility settings we could always add the page back and duplicate the setting into that page too.

Also minor tweaks to animations for the other post settings


EricBAndrews:
I like duplicating this setting into post settings, but don't want to get rid of the Accessibility page--there are a couple more accessibility settings we had in v1 that we should include for 2.0 (#1629, #1630)

EricBAndrews (changes requested):
I like duplicating this setting into post settings, but don't want to get rid of the Accessibility page--there are a couple more accessibility settings we had in v1 that we should include for 2.0 (#1629, #1630)

Sjmarf:
Fair enough; I'll put it back 👍

EricBAndrews (approved):
LGTM! 🔥

PR #1628, opened 2025-01-19
Fix "data could not be read because it is missing"

### LINKED ISSUE #1445: "The data couldn’t be read because it is missing." when opening app
From TF feedback:

> When opening the app (after it has been force quit), the feeds tab gives a generic error briefly. This occurs most of the time: The data couldn’t be read because it is missing.
> Nothing appears wrong with the app itself when this occurs.

and:

> Every time I load Mlem fresh or after time, I get this error. It almost feels like it has a content delay? I initially see nothing but this error then content populates.

and on Dec 24 build 180:

> Request error message when opening the app or refreshing/loading the All Feed. “The data couldn’t be read because it is missing.”

I've seen this before at one point but can't reproduce it at the moment. I think it might be related to `CodableSettings`?

Sjmarf:
Closes #1445

Also replaced cases where we print an error using `error.localizedDescription` with `String(describing: error)` instead - this seems to give a more descriptive error message


EricBAndrews (approved):
LGTM! 👍

Issue #1393, opened 2024-10-18
"Error" while scrolling timeline


RayJW:
**Describe the bug**
It regularly happens while I'm scrolling my timeline. Just a red popup saying error with the description "The data couldn’t be read because it isn’t in the correct format." when I click on the message.

**Expected Behavior**
Not getting the error I guess.

**Actual Behavior**
Regularly seeing the error. 

**To Reproduce**
Scrolling any timeline for me. The instance I see it on is sh.itjust.works, but not sure how related that is.

**Mlem Information:**
2.0 (104), I also checked the more stable build but I'm getting the same behaviour.

Sjmarf:
Possibly #1445

RayJW:
> Possibly [#1445](https://github.com/mlemgroup/mlem/issues/1445)

I can still reliably reproduce the error with 2.0 (180). Let me know if there is anything you would like me to try.

Sjmarf:
@RayJW

Does going to `Settings` -> `General` -> `Import/Export Settings` and tapping "Reset Settings State" fix it? This won't reset your settings, it just deletes some of the files related to migrating from v1.

RayJW:
> Does going to `Settings` -> `General` -> `Import/Export Settings` and tapping "Reset Settings State" fix it? This won't reset your settings, it just deletes some of the files related to migrating from v1.

Sorry, I don't have this option in my settings. I checked both provided beta builds (180) and (205). I also tried toggling `Developer Mode` on. The only available options in this menu are `Save Settings`, `Restore Settings`, `Export Settings`, and `Import Settings`. None of these seem to help alleviate the issue.

Sjmarf:
Ah sorry I didn't notice that it's only shown on debug builds. I'll add an option for it in developer settings in the next weekly TF build on Friday.

RayJW:
> Ah sorry I didn't notice that it's only shown on debug builds. I'll add an option for it in developer settings in the next weekly TF build on Friday.

With the new build `Restore Settings` is now visible. While the initial error is gone, I am now getting a new error on every start-up, even after restoring the settings.

> The data couldn’t be read because it isn’t in the correct format.

Edit: Ok, it seems like the copy button in the error log is not working correctly either. This is the correct error:

> Optional(Swift.DecodingError.typeMismatch(Swift.-Dictionary<Swift.String, Any>,
Swift.DecodingError.Context(codingPath:
[...]

PR #1634, opened 2025-01-19
Add "reset settings state" button to developer settings

Sjmarf:
See #1393


EricBAndrews (changes requested):
Looks good! We should also remove this button from Import/Export Settings

EricBAndrews (approved):
LGTM! 🔥

PR #1637, opened 2025-01-20
More Image Control Bugfix

EricBAndrews:
Fixes an issue where image controls would reappear when interactively dismissing the image viewer after tapping to dismiss the controls


Sjmarf (approved):
Nice!

PR #1649, opened 2025-01-24
Use types for issue templates (maybe)

Sjmarf:
No idea whether this will work or not; it isn't documented.


EricBAndrews (approved):
Let's try it!

PR #1654, opened 2025-01-24
Hide status bar in image viewer

Sjmarf:
What do you think? I think this is cleaner, and Apple hides the status bar when you tap on an image in Photos.

I had to rearrange some of the color scheme code to get this to work properly


EricBAndrews (approved):
LGTM! 🚀 

This is much cleaner 😁

PR #1656, opened 2025-01-24
Safety & Filtering Settings

Sjmarf:
Organised some settings into Safety and Filtering

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Settings/PostReadIndicatorSettingsView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/PostReadIndicatorSettingsView.swift
@@ -25,7 +25,7 @@ struct PostReadIndicatorSettingsView: View {
             SettingsHeaderView(
                 title: "Read Indicator",
                 // swiftlint:disable:next line_length
-                description: "When you've read a post already, its title will appear dimmed. If you like, you can choose an additional way of indicating read status.",
+                description: "Read posts are indicated with dimmed title text. If you like, you can choose an additional way of indicating read status.",
                 systemImage: Icons.read
             )
             .tint(palette.secondary)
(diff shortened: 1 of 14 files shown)


Mlem/App/Views/Root/Tabs/Settings/PostReadIndicatorSettingsView.swift
```diff
@@ -25,14 +25,10 @@ struct PostReadIndicatorSettingsView: View {
...
                 // swiftlint:disable:next line_length
-                description: "When you've read a post already, its title will appear dimmed. If you like, you can choose an additional way of indicating read status."
-            ) {
-                Image(systemName: Icons.read)
-                    .resizable()
-                    .aspectRatio(contentMode: .fit)
-                    .frame(height: 60)
-                    .foregroundStyle(.tertiary)
-            }
+                description: "When you've read a post already, its title will appear dimmed. If you like, you can choose an additional way of indicating read status.",
```
EricBAndrews: *nit* slightly more concise
```suggestion
                description: "Read posts are indicated with dimmed title text. If you like, you can choose an additional way of indicating read status.",
```

Mlem/App/Views/Root/Tabs/Settings/SafetyBlurNsfwSettingsView.swift
```diff
@@ -0,0 +1,38 @@
...
+        }
+        .contentMargins(.top, 16)
+        .labelStyle(.conditional)
+    }
+    
+    @ViewBuilder
+    var headerView: some View {
+        SettingsHeaderView(
+            title: "Blur NSFW Content",
+            description: "Choose which contexts Not Safe For Work content should be blurred in.",
```
EricBAndrews: *nit*

```suggestion
            description: "Choose when Not Safe For Work content should be blurred.",
```

Alternatively:
`Choose contexts in which to blur Not Safe For Work content`, if we want to keep the "in which contexts" phrasing--the basic revision here is to keep "in" and "which" together

EricBAndrews: "When" and "where" are both good contenders here--I picked "when" because it feels slightly more natural ("When should I blur? -> Always/outside NSFW comm/never" vs "Where should I blur? -> Everywhere/outside NSFW comm/nowhere")

EricBAndrews (approved):
LGTM! 🔥 

Some minor wording suggestions

PR #1657, opened 2025-01-25
Localization Fixes

### LINKED ISSUE #1379: Include `EndOfFeedView` image names in Localization
The `EndOfFeedView` messages are English-specific idioms that may not exist in other languages. The images we use in that view reflect the message. E.g, the "It's turtles all the way down" uses the turtle symbol. 

Rather than directly translating these messages, translators should be encouraged to use a suitable idiom from their language if one exists. The images we use are unlikely to match those idioms, so we should make the image names localized too.

### LINKED ISSUE #1601: Make `BasicAction` accept a `LocalizedStringResource` for `confirmationDialog`

Sjmarf:
- Closes #1379
- Closes #1601
- Fixes the "Content warning" navigation link value in Settings (I missed this in my last PR, oops)

### DIFF
--- a/Mlem/App/Models/Action/BasicAction.swift
+++ b/Mlem/App/Models/Action/BasicAction.swift
@@ -24,13 +24,17 @@ struct BasicAction: Action {
     init(
         id: String,
         appearance: ActionAppearance,
-        confirmationPrompt: String? = nil,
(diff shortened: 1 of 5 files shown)


EricBAndrews (approved):
LGTM! 💯

PR #1659, opened 2025-01-25
Change password

### LINKED ISSUE #1427: Change password

Sjmarf:
> [!note]
> Relies on [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/103)

Closes #1427

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Settings/AccountSignInSettingsView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/AccountSignInSettingsView.swift
@@ -9,6 +9,7 @@ import SwiftUI
 
 struct AccountSignInSettingsView: View {
     @Environment(AppState.self) var appState
+    @Environment(NavigationLayer.self) var navigation
     @Environment(Palette.self) var palette
     
     var body: some View {
@@ -23,7 +24,13 @@ struct AccountSignInSettingsView: View {
                     }
                 }
             }
+            Section {
+                Button("Change Password", systemImage: "key") {
+                    navigation.openSheet(.settings(.accountChangePassword))
+                }
+            }
         }
+        .labelStyle(.conditional)
         .navigationTitle("Sign-In & Security")
     }
 }
--- a/Mlem/App/Views/Root/Tabs/Settings/AccountSignInSettingsView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/AccountSignInSettingsView.swift
@@ -25,7 +25,7 @@ struct AccountSignInSettingsView: View {
                 }
             }
             Section {
-                Button("Change Password", systemImage: "key") {
+                Button("Change Password", systemImage: Icons.security) {
                     navigation.openSheet(.settings(.accountChangePassword))
                 }
             }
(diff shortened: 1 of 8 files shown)


Mlem/App/Views/Root/Tabs/Settings/AccountSignInSettingsView.swift
```diff
@@ -23,7 +24,13 @@ struct AccountSignInSettingsView: View {
                     }
                 }
             }
+            Section {
+                Button("Change Password", systemImage: "key") {
```
EricBAndrews: *nit*

```suggestion
                Button("Change Password", systemImage: Icons.security) {
```

(this also needs to be added to `Icons`, and the "Sign in & Security" settings page should be updated to use `Icons.securityFill`)

EricBAndrews (changes requested):
Looks good and works well! A couple notes:

- The password fields should have a toggleable hide/show
- Passwords must be between 10 and 60 characters; we should prevent submission of passwords that don't meet that criteria. The length bound is [hard-coded in the backend](https://github.com/LemmyNet/lemmy/blob/8ee81a396727018034438e568f74b67d4219ba1a/crates/api_common/src/utils.rs#L400) so we don't need to worry about fetching it.
- Invalid passwords return `400: "invalid_password"`. We should catch this and display an "invalid password" status rather than popping the error toast. This also serves as a fallback for the password length check in case either Lemmy updates their requirements or an instance is running a spiked version with custom requirements.

Sjmarf:
> The password fields should have a toggleable hide/show

I don't think this is feasible - `SecureField` doesn't support it, and Apple themselves don't use this design pattern afaik.

> Passwords must be between 10 and 60 characters; we should prevent submission of passwords that don't meet that criteria. The length bound is [hard-coded in the backend](https://github.com/LemmyNet/lemmy/blob/8ee81a396727018034438e568f74b67d4219ba1a/crates/api_common/src/utils.rs#L400) so we don't need to worry about fetching it.

Sure, nice catch 👍 

> Invalid passwords return 400: "invalid_password". We should catch this and display an "invalid password" status rather than popping the error toast. This also serves as a fallback for the password length check in case either Lemmy updates their requirements or an instance is running a spiked version with custom requirements.

Done 👍

Mlem/App/Views/Root/Tabs/Settings/ChangePasswordView.swift
```diff
@@ -0,0 +1,127 @@
...
+                }
+                .frame(maxWidth: .infinity)
+                .disabled(viewState != .initial)
+            } footer: {
+                Group {
+                    if !newPassword.isEmpty {
+                        if newPassword != confirmNewPassword {
+                            Text("Passwords don't match.")
+                        } else if !(10 ... 60 ~= newPassword.count) {
+                            Text("New must be between \(10) and \(60) characters long.")
```
EricBAndrews: *nit*
```suggestion
                            Text("New password must be between \(10) and \(60) characters long.")
```

EricBAndrews (approved):
LGTM! 💯 

Shame about the `SecureField` behavior--that's been a standard QoL feature for years now, a bit odd that Apple doesn't like it but I guess they're really pushing "just use a password manager"

PR #1675, opened 2025-01-28
Shields.io badges

### LINKED ISSUE #1581: Render shields.io badges
Badges from https://shields.io/ are commonly used in instance descriptions

Sjmarf:
Closes #1581 

Known issues:
- The spacing between badges is far too large. Fixing this requires significant changes to LemmyMarkdownUI; I've created an issue [here](https://github.com/mlemgroup/LemmyMarkdownUI/issues/49) to keep track of it.
- Tapping on a badge doesn't open a link yet. Created an issue [here](https://github.com/mlemgroup/LemmyMarkdownUI/issues/50).


EricBAndrews (changes requested):
Nice work!

LemmyMarkdownUI got moved into the project tree--it should go back into Package Dependencies.

Badges are also all displayed with `.accent`--I'm content to leave as-is, but as a luxury feature it seems feasible to parse the color in the badges spec.

Sjmarf:
> Badges are also all displayed with .accent--I'm content to leave as-is, but as a luxury feature it seems feasible to parse the color in the badges spec.

Whoops, I meant to mention this. We parsed the colors in v1, but it's not quite as easy in v2 because of the theming system. If a badge is colored "blue", we can't guarantee that the current theme has a "blue" equivalent. I'm not sure what the best way to do that would be.

EricBAndrews (approved):
LGTM! 🔥

PR #1684, opened 2025-01-31
Fix inline images being displayed in feed

### LINKED ISSUE #1679: Inline images are displayed full-size in the feed, rather than as an icon
NGL I do kinda like this, but it could be inconvenient. Maybe we should have a setting for it

See also #1095

Sjmarf:
Closes #1679


EricBAndrews (approved):
LGTM! 🔥

PR #1687, opened 2025-01-31
Render read pinned posts as compact

### LINKED ISSUE #1586: Render read pinned posts as compact

EricBAndrews:
## Issues

- closes #1586


Sjmarf (changes requested):
In Tiled mode we shouldn't switch to compact

--------

If you tap on an unread post, you can see the post size change to compact as the new page is sliding onto the navigation stack. That's also how it worked in v1, but I wonder if there's a less jarring way of handling the switch. Maybe:
- A short animation from one to the other
- Or, only switch to compact when the user refreshes (could be tricky to implement)

Feel free to leave OOS

EricBAndrews:
> In Tiled mode we shouldn't switch to compact

Nice catch!

> A short animation from one to the other

This seems the nicest--implementing it isn't the prettiest thing in the world but it's not too bad.

Sjmarf (approved):
LGTM! 🔥

PR #1700, opened 2025-02-03
Drag-To-Zoom

EricBAndrews:
## Description

Adds an accessibility feature to drag the side of the image viewer to zoom. This can be toggled in Settings -> Accessibility -> Drag-To-Zoom Images.

Also makes some minor enhancements to the image viewer:
- Double tap to zoom out no longer zooms in first
- While actively changing scale, a zoom indicator appears in the top left.

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Settings/AccessibilitySettingsView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/AccessibilitySettingsView.swift
@@ -50,7 +50,7 @@ struct AccessibilitySettingsView: View {
                     "Drag-To-Zoom Images",
                     value: .init(localized: zoomDraggerLocation.label),
                     fallbackValue: "",
-                    systemImage: "arrow.up.and.down.and.sparkles",
+                    systemImage: Icons.zoomDragger,
                     destination: .settings(.zoomDragger)
                 )
             } header: {
--- a/Mlem/App/Views/Root/Tabs/Settings/AccessibilitySettingsView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/AccessibilitySettingsView.swift
@@ -47,7 +47,7 @@ struct AccessibilitySettingsView: View {
             
             Section {
                 NavigationLink(
-                    "Drag-To-Zoom Images",
+                    "Slide to Zoom Images",
                     value: .init(localized: zoomDraggerLocation.label),
                     fallbackValue: "",
                     systemImage: Icons.zoomDragger,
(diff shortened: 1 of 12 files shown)


Mlem/App/Views/Root/Tabs/Settings/AccessibilitySettingsView.swift
```diff
@@ -32,12 +33,28 @@ struct AccessibilitySettingsView: View {
...
             Section {
                 Toggle("Website Thumbnail Indicator", systemImage: Icons.browser, isOn: $websiteThumbnailIcon)
                 Toggle("Settings Icons", systemImage: Icons.icon, isOn: $showSettingsIcons)
+            } header: {
+                Text("Non-Text Indicators")
+            }
+            
+            Section {
+                NavigationLink(
+                    "Drag-To-Zoom Images",
```
Sjmarf: IMO we shouldn't hyphenate this. Apple doesn't include hyphens in similar circumstances:

```suggestion
                    "Drag to Zoom Images",
```

I also wonder whether "Swipe to Zoom" might be better? "Drag" suggests that you're moving something, which isn't the case here. Though "Swipe" sounds like one smooth motion after which you let go, which isn't the case here either... I'm not sure.

Mlem/App/Views/Root/Tabs/Settings/ZoomDraggerSettingsView.swift
```diff
@@ -0,0 +1,112 @@
...
+            Picker("Location", selection: $zoomDraggerLocation) {
+                ForEach(ZoomDraggerLocation.allCases, id: \.self) { location in
+                    Label(String(localized: location.label), systemImage: location.systemImage)
+                        .tag(location)
+                }
+            }
+            .labelsHidden()
+            .pickerStyle(.inline)
+        }
+        .labelStyle(.conditional)
```
Sjmarf: Anywhere we use `SettingsHeaderView` we need to add this `.contentMargins` line to make the padding above the section a little smaller. Annoyingly this can't be put inside of the `SettingsHeaderView` itself (I couldn't find a way to make it work, anyway).

```suggestion
        .labelStyle(.conditional)
        .contentMargins(.top, 16)
```

Mlem/App/Views/Root/Tabs/Settings/ZoomDraggerSettingsView.swift
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+struct ZoomDraggerSettingsView: View {
+    @Setting(\.zoomDraggerLocation) var zoomDraggerLocation
+    
+    var body: some View {
+        Form {
+            SettingsHeaderView(
+                title: "Drag-To-Zoom",
```
Sjmarf: See above
```suggestion
                title: "Drag to Zoom",
```

Sjmarf (approved):
LGTM! I love the preview animation in Settings 🔥 A few minor wording comments

iirc Apollo had a feature where you could double tap and hold, then swipe vertically and zoom in and out. That could be a good feature to implement in future 🤔

PR #1701, opened 2025-02-03
add slur filter warning

### LINKED ISSUE #1458: Show warning in post/comment editor if text matches instance slur filter
We had this in v1; I'm not particularly concerned with whether this gets put into v2 or not. Would be nice to re-implement this in future though.

EricBAndrews:
## Issues

- closes #867 

## Description

Adds a filter warning to the post and comment editor. 

## Implementation Notes

- Grouped editor views in `Pages/Editors`


Mlem/App/Views/Pages/Editors/FilterViolationWarning.swift
```diff
@@ -0,0 +1,38 @@
...
+                .padding(.horizontal, Constants.main.halfSpacing)
+                .background {
+                    Capsule()
+                        .fill(palette.warning.opacity(0.2))
+                        .stroke(palette.warning)
+                }
+            
+            Text(context) +
+            Text(failingText).fontWeight(.semibold) +
+            Text(" are forbidden by this instance's filters.")
```
Sjmarf: We can use string interpolation here to make localizing this easier. Something like

```swift
let innerText = Text(failingText).fontWeight(.semibold)
if isPost {
    Text("Posts containing \(innerText) are forbidden by this instance's filters")
} else {
    Text("Comments containing \(innerText) are forbidden by this instance's filters")
}
```

Mlem/App/Views/Pages/Editors/FilterViolationWarning.swift
```diff
@@ -0,0 +1,38 @@
...
+    let failingText: String
+    let isPost: Bool
+    
+    var context: LocalizedStringResource {
+        isPost ? "Posts containing " : "Comments containing "
+    }
+    
+    var body: some View {
+        VStack(alignment: .leading, spacing: Constants.main.standardSpacing) {
+            (Text(Image(systemName: Icons.warning)) + Text(" Filter violation"))
```
Sjmarf: Ideally we'd avoid having a leading space in the localization string. Would a label work here?

```suggestion
            Label("Filter violation", systemImage: Icons.warning)
```

Mlem/App/Views/Pages/Editors/FilterViolationWarning.swift
```diff
@@ -0,0 +1,38 @@
...
+                .foregroundStyle(palette.warning)
+                .padding(Constants.main.standardSpacing)
+                .padding(.horizontal, Constants.main.halfSpacing)
```
Sjmarf: IMO we should make this badge a little smaller

```suggestion
                .font(.footnote)
                .foregroundStyle(palette.warning)
                .padding(.vertical, 5)
                .padding(.horizontal, 7)
```

Sjmarf (changes requested):
Nice, works great 👍 

Ideally, the slur filter would use the regex from the instance of the selected account *within* the comment editor, rather than the currently active account. Doing that would mean fetching the instance for the selected account, which I don't believe we do currently.

In the post editor, the user can crosspost to several different communities using several different accounts. So, we'd need to check the filter on each instance and show a warning that specifies which instance's slur filter the text is in violation of.

Sjmarf (approved):
Nice! Works perfectly 🚀

PR #1702, opened 2025-02-03
Fix Toast crash

Sjmarf:
Fixed a crash that sometimes occurred when adding/removing toasts on a background thread


EricBAndrews (approved):
Nice! 👍

PR #1704, opened 2025-02-04
Add previews

Sjmarf:
> [!note]
> Relies on [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/111).

Added SwiftUI previews to some views:
- `LargePostView`
- `HeadlinePostView`
- `CompactPostView`
- `PersonListRowBody`
- `PersonListRow`
- `CommunityListRowBody`
- `CommunityListRow`

In order to accomplish this, I've added a `.mock` static method to _some_ middleware models:
- `Post1` and `Post2`
- `Person1` and `Person2`
- `Community1` and `Community2`

I'll be expanding this in future to include the other models. The definition of `.mock` only compiles in `DEBUG` mode, so should be used in SwiftUI previews only.

To create a mocked model, use this syntax:

```swift
#Preview(traits: .sampleEnvironment, .sizeThatFitsLayout) {
    CompactPostView(post: Post2.mock(.generic))
}
```

`.generic` creates a model with some sample text:

For `Person` and `Community` models, you can alternatively create a realistic-looking mock model like so:

```swift
Community.mock(.realistic(.news))
```

I will be adding `.realistic` versions of `Post` too in future.

At some point, I intend to redo the screenshots of Mlem that we use on the App Store, the website and the GitHub README. When we do this, I think it would be best to use entirely fake screenshots of the UI generated using SwiftUI previews. Doing this will allow us to easily recreate the same screenshots later. This would be especially useful if we ever wanted to localize the App Store screenshots into multiple languages.

I've included the names of the `.realistic` mock models in a new `PreviewLocalizable` file, which we could send for translation. That means that any community names, post body content etc in the App Store screenshots would also be properly localized.

All of the preview-related stuff, including `PreviewLocalizable` and the avatar images used in the previews, are only compiled in `DEBUG` mode.


Mlem.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -60,8 +60,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/mlemgroup/MlemMiddleware",
       "state" : {
-        "revision" : "d3752b6176e10f5cef56913915c85d62e9bb6b15",
-        "version" : "0.72.2"
+        "branch" : "sjmarf/mock",
+        "revision" : "124c760d49208714ba5b1b2f6ef9c29251d93a01"
```
EricBAndrews: reminder

EricBAndrews (approved):
Fantastic work! 😁  

App Store screenshots have been a pain point for a while, this ought to clear that right up.

PR #1712, opened 2025-02-05
Shields.io badges now correctly open link when tapped

Sjmarf:
I mentioned this issue in [the original Shields.io PR](https://github.com/mlemgroup/mlem/pull/1675). 

This doesn't do anything with links that surround non-shield images, because the "tap to open link" gesture would conflict with tapping to open the image viewer. We could put an "open link" button in the image viewer, maybe.


EricBAndrews (approved):
LGTM! 🔥

PR #1714, opened 2025-02-05
Audio Fixes

### LINKED ISSUE #1708: "Mute Videos" sometimes doesn't work
### Requirements

### Description

Sometimes videos play audio anyway.

### Mlem Version

_No response_

### LINKED ISSUE #1709: Mute button/zoom slider conflict
### Requirements

### Description

Enabling the right-side zoom slider overrides the mute tap gesture with the slide gesture

### Mlem Version

_No response_

EricBAndrews:
## Issues

- closes #1708 
- closes #1709 

## Description

Fixes the above bugs. In the image viewer, the mute and play buttons are now part of the viewer control layer rather than embedded within the video view, and taps on the video hide/show controls the same as when viewing images.

## Implementation Notes

Media state is now handled with `MediaControlState`, which exists as a unified control state abstraction and is provided environmentally to the concrete media view types. This will also provide a bridge between a future common scrubbing control interface and the different scrubbing implementations that the different media types will require.

Mute is now handled using `player.volume` instead of `player.isMuted`; this is slightly more performant and way more reliable, since it doesn't hook into system-side audio APIs. The only behavior change here is that changing system volume won't automatically unmute a currently playing video.


Mlem/App/Views/Pages/ImageViewer+Views.swift
```diff
@@ -98,9 +128,26 @@ extension ImageViewer {
...
+            controlState.muted.toggle()
+        } label: {
+            Image(systemName: controlState.muted ? Icons.muted : Icons.unmuted)
+                .scaledToFit()
+                .frame(width: 22, height: 22)
+                .contentTransition(.symbolEffect(.replace, options: .speed(2)))
+                .padding(Constants.main.standardSpacing)
+                .background(.ultraThinMaterial, in: .circle)
+                .padding(Constants.main.standardSpacing)
+                .padding([.top, .leading], Constants.main.standardSpacing)
```
Sjmarf: This `.top` padding causes the play and mute buttons to appear lower down than the share/save buttons. Is this intentional? I think I'd prefer it if they were centered vertically. What do you think?

Sjmarf (approved):
Nice work! Having a real pause button in the image viewer is a great improvement :) One minor design question

Mlem/App/Views/Pages/ImageViewer+Views.swift
EricBAndrews: Nice catch!

EricBAndrews:
Pls disregard re-review request lol

PR #1716, opened 2025-02-05
padding fix

EricBAndrews:
Previous padding forgot to account for the -2 y offset applied to icons in the central bar so was off by 2px


Sjmarf (approved):
LGTM 👍

Issue #1717, opened 2025-02-06
Keyboard volume inverse of system volume


EricBAndrews:
As of #1719, using `sharedInstance()` fixes the old loud keyboard bug, but causes this behavior.

Turning system volume up turns keyboard volume down, and vice versa.

EricBAndrews:
Cannot reproduce on iOS 26

PR #1719, opened 2025-02-06
Fix loud keyboard

EricBAndrews:
## Issues

- progress towards #1717 (mostly, see below)

## Description

Fixes the keyboard clicks being way too loud. 

## Implementation Notes

The cause of this was initializing the `CHHapticengine` with `audioSession: nil`. I don't know why this caused that behavior, since that's how Apple says you're supposed to do it if you aren't going to play audio, but so it goes. Using the `sharedInstance` resolves the issue.

**HOWEVER:** keyboard click volume is now the _inverse_ of the system audio (though it does mute if system audio is 0). This is still a significant improvement over the deafening clicks before. I'll update #1717 when this fix merges to reflect that behavior.

Things I tried while fixing this:
- Create a separate `AVAudioSession` for the haptic engine: no effect
- Start the engine to play the haptic and stop it after: doesn't fix the issue _and_ makes haptics laggy
- Set `CHHapticEngine.playsHapticsOnly`/`CHHapticEngine.mutesAudio`: no effect, when combined with the above techniques sometimes makes notifications silent too
- Using players instead of the engine itself: no effect

### DIFF
--- a/Mlem/App/Globals/Definitions/HapticManager/HapticManager.swift
+++ b/Mlem/App/Globals/Definitions/HapticManager/HapticManager.swift
@@ -53,7 +53,9 @@ class HapticManager {
     func initEngine() -> CHHapticEngine? {
         if CHHapticEngine.capabilitiesForHardware().supportsHaptics {
             do {
-                let ret = try CHHapticEngine(audioSession: nil)
+                let ret = try CHHapticEngine(audioSession: .sharedInstance()) // (audioSession: nil)
+                ret.playsHapticsOnly = true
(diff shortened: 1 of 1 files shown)


Sjmarf (approved):
LGTM 👍 We used to use `.sharedInstance()`, but [I switched it](https://github.com/mlemgroup/mlem/commit/935deca78e8238522a7c3041a04431f740db9a3f) to `nil` a while back to fix an issue where music would pause every time you switch to Mlem. I can no longer reproduce that issue with this PR, though 🤔

PR #1720, opened 2025-02-06
Stop Printing Errors

EricBAndrews:
## Issues

- progress towards #124 

## Description

Adds a `silent` toggle to `handleError` that suppresses the toast and replaces all error prints with calls to `handleError` with `silent: true`. This gives us consistent, detailed error prints through `handleError`, plus better insight into errors that may happen outside of the debugger mode through the error log.

### DIFF
--- a/Mlem/App/Models/ErrorDetails.swift
+++ b/Mlem/App/Models/ErrorDetails.swift
@@ -14,6 +14,7 @@ struct ErrorDetails: Hashable {
     var title: String?
     var body: String?
     var error: Error?
+    var location: String?
     var systemImage: String?
     var buttonText: String?
     var refresh: (() async -> Bool)?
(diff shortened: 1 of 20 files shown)


Mlem/App/Models/ErrorDetails.swift
```diff
@@ -62,12 +66,18 @@ struct ErrorDetails: Hashable {
...
-        if let error = error as? ApiClientError {
-            output = error.description
+    var errorDescription: String {
+        if let error = error as? CustomStringConvertible {
+            return error.description
         } else {
-            output = String(describing: error)
+            return String(describing: error)
+        }
+    }
```
Sjmarf: `String(describing: error)` returns error.description` if `error` conforms to `CustomStringConvertible`; we don't need this extra check

Sjmarf (approved):
Nice! This is much nicer. One minor nit

Mlem/App/Models/ErrorDetails.swift
EricBAndrews: Nice one 👍

PR #1721, opened 2025-02-06
TF sheet crash fix

Sjmarf:
Possible fix for a TF crash. Not entirely sure that this fixes it - added a main thread check to be safe


EricBAndrews (approved):
Nice 👍

PR #1726, opened 2025-02-07
Fix post/comment sometimes not editable

Sjmarf:
See here: https://lemmy.ml/post/19667291/16576138

It looks like lemmy-ui sends `nil` for `languageId` when editing a comment unless you actually change the language. We were sending the language id back every time. I assume this is what was causing the issue.


EricBAndrews (approved):
LGTM! 🔥

PR #1729, opened 2025-02-09
Possible fix for inline image crash

Sjmarf:
We've had a few TF crashes related to `UIGraphicsBeginImageContextWithOptions`. I can't repro the crash, but I'm guessing it might be because it needs to run on `@MainActor`


EricBAndrews (approved):
LGTM 👍

PR #1734, opened 2025-02-10
Markdown fixes

### LINKED ISSUE #1447: Certain markdown blocks cause app to freeze
Via TF Feedback. This post causes a freeze: https://sopuli.xyz/post/19489280

Possibly a result of the upvote/downvote characters in the post content? I did some brief testing and it looks like this is a problem with the `Markdown` view rather than the parsing logic, because replacing `Markdown` with `MarkdownText` fixes it.

Sjmarf:
Closes #1447. The app seemed to freeze forever, but this wasn't actually the case - it would unfreeze eventually. 

It turns out that the `VStack` used in the `Markdown` view was spending all that time doing layout calculations for the large number of complex child views. To fix this, I simplified the logic by replacing the `VStack` with a custom `Layout` that only properly does calculations for the x axis and just always gives maximum space on the y axis. We're able to do this because the height of `Markdown` is always guaranteed to be infinite.

 There is still a small hitch when opening the post mentioned in the issue but it's no-where near as bad as before.

Also fixed shields.io badges being spaced super far apart.

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -3294,7 +3294,7 @@
 			repositoryURL = "https://github.com/mlemgroup/LemmyMarkdownUI";
 			requirement = {
 				kind = upToNextMinorVersion;
-				minimumVersion = 0.8.0;
+				minimumVersion = 0.9.0;
 			};
 		};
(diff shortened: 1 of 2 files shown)


EricBAndrews (approved):
Nice work! 🔥

PR #1745, opened 2025-02-14
Remember last active account when opening app

### LINKED ISSUE #39: Ability to set default account for app launch
ability to set a default account, so app automatically loads feed from the default account

Sjmarf:
When you open the app, the account you were using when you last closed the app is now made active. I think this is better UX than making the user explicitly set a default account.

Closes #39

### DIFF
--- a/Mlem/App/Globals/Definitions/AccountsTracker.swift
+++ b/Mlem/App/Globals/Definitions/AccountsTracker.swift
@@ -36,7 +36,7 @@ class AccountsTracker {
         if let lastUsedAccount = sorted.last {
             return lastUsedAccount
         }
-        return defaultGuestAccount
+        return userAccounts.first() ?? defaultGuestAccount
     }
     
     var defaultGuestAccount: GuestAccount {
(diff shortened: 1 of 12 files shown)


Mlem/App/Globals/Definitions/AccountsTracker.swift
```diff
@@ -25,11 +25,24 @@ class AccountsTracker {
...
+    func mostRecentAccount() -> any Account {
+        var allAccounts: [any Account] = userAccounts + guestAccounts
+        if let activeAccount = allAccounts.first(where: { $0.activityState == .active }) {
+            return activeAccount
+        }
+        let sorted = allAccounts.sorted(by: { $0.activityState.lastUsed ?? .distantPast < $1.activityState.lastUsed ?? .distantPast })
+        if let lastUsedAccount = sorted.last {
+            return lastUsedAccount
+        }
+        return defaultGuestAccount
```
EricBAndrews: This should be the first user account, if present--otherwise the app puts you on a guest account at first launch following this patch
```suggestion
        return userAccounts.first() ?? defaultGuestAccount
```

EricBAndrews (changes requested):
Nice work! This is a smooth ux

EricBAndrews (approved):
LGTM! 🚀

PR #1747, opened 2025-02-15
Display error when `ContentLoader` fails

### LINKED ISSUE #1746: Image failure background not distinct in dark mode
Looks like it should be `.tertiaryGroupedBackground rather than `.secondaryBackground`, or something like that

Sjmarf:
Previously it would just show a `ProgressView` indefinitely when it fails. The test case here is #1735.

Also fixed #1746

See also [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/118) - it's not directly linked, but if you aren't using that version of MlemMiddleware you'll get an `assertionFailure` rather than seeing the error details

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -3357,8 +3357,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware";
 			requirement = {
-				branch = "sjmarf/bug-fix";
-				kind = branch;
+				kind = upToNextMinorVersion;
+				minimumVersion = 0.76.0;
 			};
 		};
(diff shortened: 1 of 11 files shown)


Mlem.xcodeproj/project.pbxproj
```diff
@@ -3357,8 +3357,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/mlemgroup/MlemMiddleware";
 			requirement = {
-				kind = upToNextMinorVersion;
-				minimumVersion = 0.75.0;
+				branch = "sjmarf/bug-fix";
+				kind = branch;
```
EricBAndrews: reminder

EricBAndrews (approved):
LGTM!

One minor nit--`ErrorView` should read "Hide Details" when details are shown.

PR #1748, opened 2025-02-15
Refactor `CommentWrapper`

Sjmarf:
No functional changes, just rearranged some logic. In hindsight I think conforming `CommentWrapper` to `Comment2Providing` was the wrong choice, so I have removed that conformance and renamed it to `CommentTreeNode`. This more clearly separates the UI state (e.g. `collapsed`) from the actual Lemmy state of `Comment2` (e.g. `saved` etc).


EricBAndrews (approved):
LGTM! Much cleaner 😁

PR #1750, opened 2025-02-15
Remove "Quick Look" button from context menu

### LINKED ISSUE #1742: "Quick Look" button in image context menu doesn't work
It works correctly in the image viewer, but not in the context menu. Tbh we could just remove it from the context menu

Sjmarf:
It's still accessible from the image viewer

Closes #1742


EricBAndrews (approved):
Nice 👍

PR #1752, opened 2025-02-15
Localize LemmyMarkdownUI

Sjmarf:
Localized the few strings that LemmyMarkdownUI uses:
- The default spoiler title (`"Spoiler"`)
- The `"Table"` text that is shown when you try to render a table using `MarkdownText` rather than `Markdown`

I've also added localization for the slur filter replacement text. Cases of `*removed*` are now replaced with a special `.censored` node:

This should make it more clear to users that the content was removed by the slur filter. The `"Censored"` text is localized. Lemmy themselves don't localize the `*removed*` text for non-English instances, so this is an improvement.

Of course, there's a chance of false positives if someone writes `*removed*` of their own accord. But I think it's worth doing anyway.

Also fixed a bug where community links would sometimes go onto more than one line like this:


EricBAndrews (approved):
LGTM! 🔥

PR #1760, opened 2025-02-16
Fix dual load bug

### LINKED ISSUE #1405: Feed-loading API calls performed twice when feed opened

Sjmarf:
Closes #1405

This was caused by this `if`/`else` block:

```swift
if let feedLoader {
    content
        .refreshable {  await refresh(feedLoader)  }
        // ...
} else {
    content
}
```

When `feedLoader` changed from `nil` to not `nil`, the view was entirely replaced because it switches from one `content` declaration to the other. I've refactored it to avoid doing this


EricBAndrews (approved):
LGTM! 💯

PR #1762, opened 2025-02-17
Interaction bar editor comment preview

### LINKED ISSUE #1690: Comment interaction bar editor preview should be a comment rather than a post

EricBAndrews:
## Issues

- closes #1690

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Settings/InteractionBarEditor/InteractionBarEditorView+Views.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/InteractionBarEditor/InteractionBarEditorView+Views.swift
@@ -65,7 +65,7 @@ extension InteractionBarEditorView {
                 .foregroundStyle(palette.selectedInteractionBarItem, palette.neutralAccent.gradient)
                 .frame(width: Constants.main.smallAvatarSize, height: Constants.main.smallAvatarSize)
             
-            MockTextView()
+            MockTextView(beginOpacity: 0.4, endOpacity: 0.3)
                 .frame(maxWidth: 200)
                 .frame(height: 13)
         }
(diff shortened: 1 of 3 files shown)


Sjmarf:
Nice! I wonder whether making the avatar placeholder a simple circle might be better than using the default avatar image? I think it stands out a little too much at the moment considering that it's not actually relevant to the settings page context.

I also wonder whether giving the username label `MockText` a lower opacity would look good, to match the opacity difference between the comment body and username label in the actual view.

EricBAndrews:
Circle is a bit strange IMO:
image

To my eye the concrete anchor point of the icon makes the preview much more immediately legible as a comment rather than a bunch of stacked bars

Mlem/App/Views/Root/Tabs/Settings/InteractionBarEditor/InteractionBarEditorView+Views.swift
```diff
@@ -47,6 +32,54 @@ extension InteractionBarEditorView {
...
+    
+    @ViewBuilder
+    var commentPreviewBody: some View {
+        HStack(spacing: 7) {
+            Image(systemName: Icons.personCircleFill)
+                .resizable()
+                .scaledToFit()
+                .symbolRenderingMode(.palette)
+                .foregroundStyle(palette.selectedInteractionBarItem, palette.neutralAccent.gradient)
+                .frame(width: Constants.main.smallAvatarSize, height: Constants.main.smallAvatarSize)
```
Sjmarf, line 66: Fair enough. What do you think about lower opacity?

```suggestion
                .frame(width: Constants.main.smallAvatarSize, height: Constants.main.smallAvatarSize)
                .compositingGroup()
                .opacity(0.5)
```

EricBAndrews: Works for me 👍

Sjmarf (approved):
LGTM!

PR #1765, opened 2025-02-17
Better haptic preheating

### LINKED ISSUE #1763: Haptics hitch sometimes
### Requirements

### Description

_No response_

### Mlem Version

_No response_

EricBAndrews:
## Issues

- closes #1763 

## Description

Removes some hitches that would happen when playing haptics (especially noticeable when using a swipe action):
1. Engine startup. If the engine is not started when a haptic is played, the main thread hitches while the engine starts. `preheatEngine` handles this on the first launch, but the engine stops when the app moves into the background. I've added a `scenePhase` listener to `ContentView` to start the engine when the app moves into the foreground.
2. File read. Previously, playing a haptic required reading the pattern from the `.ahap` file. This was sometimes slow on the first read for a given pattern, again hitching the main thread. The `HapticManager` now loads each file into a `CHHapticPatternPlayer` at `init` so they're ready to go.

The best repro for the hitches is:
- Start the app
- Swipe a post
- Close the app (just move to background, no need to full quit), wait for the "haptic engine shut down" log message
- Start the app again and swipe a post

## Implementation Notes

I've also refactored the `HapticManager`:
- Rearranged functions into a slightly more sensical order (init/api/core/helpers)
- Removed `preheat()`. The preheating work is now done automatically in `init`
- Streamlined error handling

### DIFF
--- a/Mlem/App/Globals/Definitions/HapticManager/Haptic.swift
+++ b/Mlem/App/Globals/Definitions/HapticManager/Haptic.swift
@@ -8,7 +8,7 @@
 import Foundation
 
 /// Enumerates all custom defined haptics used in the app. The raw value of a case corresponds to the name of the file it is stored in.
-enum Haptic: String {
+enum Haptic: String, CaseIterable {
     /// Very gentle tap. Used for subtle feedback--things like crossing a swipe boundary
     case gentleInfo = "Gentle Info"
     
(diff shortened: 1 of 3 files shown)


Sjmarf (approved):
Nice! 💯

PR #1766, opened 2025-02-17
Clean up warnings

EricBAndrews:
Removes compiler warnings. The code changes are pretty trivial. I have pasted the relevant sections of the [build settings reference](https://developer.apple.com/documentation/xcode/build-settings-reference) docs here for the `project.pbxproj` changes, with my comments in italics:

- `CLANG_ANALYZER_LOCALIZABILITY_NONLOCALIZED`: Warn when a nonlocalized string is passed to a user interface method expecting a localized string.
- `ENABLE_USER_SCRIPT_SANDBOXING`: If enabled, the build system will sandbox user scripts to disallow undeclared input/output dependencies. *(needs to be `NO` for Swiftlint to work)*
- `ASSETCATALOG_COMPILER_GLOBAL_ACCENT_COLOR_NAME`: The name of a color resource to use as a the target’s accent color, used as the default tint color on iOS and watchOS, and accent color on macOS. *(functionality overridden by the palette system. Removing it silences the 'AccentColor not defined' warning)*
- `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES`: Always embed the Swift standard libraries in the target’s products, even if the target does not contain any Swift code. For example, this should be enabled if the target is embedding other products which contain Swift, or if it is a test target which does not contain Swift but which is testing a product which does. This setting only applies to wrapped products, not to standalone binary products. *(irrelevant for Mlem)*


Sjmarf:
> `CLANG_ANALYZER_LOCALIZABILITY_NONLOCALIZED`

I can't actually get the compiler to generate a warning with this enabled. Did you have any luck? I've got a feeling that this only applies to Objective-C code, and not Swift. I couldn't find any mention of which languages it applies to online, but all the examples I saw were in Objective-C.

EricBAndrews:
Interesting! Xcode thought it should be enabled, but it doesn't complain if it's removed again.

Sjmarf (approved):
LGTM 👍

Issue #1768, opened 2025-02-17
Add missing items to interaction bar


EricBAndrews:
### Requirements

- [x] There are no existing issues for this feature.
- [x] This is a request for a **single** feature (create multiple issues for multiple feature requests).

### Description

Some items are missing from v1:
- Moderator bar: ban, resolve
- Comment bar: saved readout

The legacy decoding logic will also need to be updated to ensure these are mapped properly from their v1 encodings.

PR #1769, opened 2025-02-18
Legacy Widget Decoding

### LINKED ISSUE #1393: "Error" while scrolling timeline

EricBAndrews:
## Issues

- closes #1393

## Description

Implements decoding for the v1 interaction bar widget format. 

v1 used both post and comment configurations in the inbox according to the underlying inbox item type; `ReplyBarConfiguration` defaults to the v1 comment configuration, since that's by far the most common inbox item.

I've grandfathered in duplicate widgets to reduce transition pains; per Slack, I think we should re-enable them in v2.

## Implementation Notes

I've isolated almost all the legacy compatibility code in the new `Legacy/` group so it should be super easy to implement the changes needed for #1768 and to rip out the legacy logic once 1.3 -> 2.0 upgrade issues aren't a concern.


Sjmarf (approved):
LGTM! 🚀

PR #1776, opened 2025-02-19
Feed fade-in animation

Sjmarf:
Added a small fade-in animation to feeds. We already had this for the comments feed in `ExpandedPostView`


EricBAndrews (approved):
LGTM! 🚀

PR #1777, opened 2025-02-19
Improve keychain access error handling

Sjmarf:
Replaced the `keychain[key]` syntax we used for Keychain access with `get`/`set` methods. These methods throw errors when something goes wrong rather than returning `nil`. This should make it easier to debug issues like #1416


EricBAndrews (approved):
LGTM! 💯

PR #1779, opened 2025-02-20
More Moderator Widgets

### LINKED ISSUE #1768: Add missing items to interaction bar

EricBAndrews:
## Issues

- closes #1768 

## Description

Adds the resolve and ban widgets to the report interaction bar and the saved readout to the comment interaction bar.

Additional changes:
- Instance ban now uses square instead of circle to avoid confusion with resolve/unresolve
- Updated report defaults to use new widgets

### DIFF
--- a/Mlem/App/Models/Action/ActionAppearance+StaticValues.swift
+++ b/Mlem/App/Models/Action/ActionAppearance+StaticValues.swift
@@ -204,6 +204,15 @@ extension ActionAppearance {
         )
     }
     
+    static func resolve(isOn: Bool) -> Self {
+        .init(
+            label: isOn ? "Unresolve" : "Resolve",
+            color: isOn ? Palette.main.negative : Palette.main.positive,
+            icon: isOn ? Icons.failureCircle : Icons.successCircle,
+            swipeIcon2: isOn ? Icons.failureCircleFill : Icons.successCircleFill
+        )
+    }
+    
     /// Adds or removes a user as administrator
     /// - Parameter isOn: true when user is admin, false otherwise
     static func addAdmin(isOn: Bool) -> Self {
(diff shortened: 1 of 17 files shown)


Mlem/App/Models/Action/ActionAppearance+StaticValues.swift
```diff
@@ -204,6 +204,15 @@ extension ActionAppearance {
         )
     }
     
+    static func resolve(isOn: Bool) -> Self {
+        .init(
+            label: isOn ? "Unresolve" : "Resolve",
```
Sjmarf, line 209:
```suggestion
            label: isOn ? "Unresolve" : "Resolve",
            isOn: isOn,
```

Mlem/App/Models/Action/ActionAppearance+StaticValues.swift
```diff
@@ -204,6 +204,15 @@ extension ActionAppearance {
...
+            color: isOn ? Palette.main.negative : Palette.main.positive,
+            icon: isOn ? Icons.unresolve : Icons.resolve,
```
Sjmarf, line 212: Imo we shouldn't toggle to `unresolve` when on the bar
```suggestion
            icon: isOn ? Icons.unresolve : Icons.resolve,
            barIcon: isOn ? Icons.resolveFill : Icons.resolve,
```

Sjmarf (changes requested):
Code looks good - a couple of minor notes on resolve button styling.

I also noticed that the "reset" button in the mod mail interaction bar editor resets back to the regular default configuration rather than the default mod mail configuration. Though this probably wasn't introduced in this PR, so feel free to leave OOS

EricBAndrews:
I took a couple stabs at the reset behavior, but it's actually pretty nasty due to how intricate the typing structure of the whole thing is. I couldn't find a particularly clean solution that wasn't complicated enough to require its own PR, so I'm leaving as-is and opened #1781 to track it.

Sjmarf (approved):
LGTM 🚀

PR #1782, opened 2025-02-21
Update `FixedImageView` when URL updates

### LINKED ISSUE #1263: FixedImageView doesn't update if url is changed by parent view after it appears

EricBAndrews:
## Issues

- closes #1263 

## Description

Refactors `FixedImageView` and the `FixedImageLoader` to change the image when the url changes. This previously did not work because state vars can't be updated at `init`. To resolve this, `FixedImageLoader` no longer accepts a url at init; the url is instead passed into the `load` function, which triggers loading. This is in turn called in a `.task(id: url)` within `FixedImageView`. This is basically the same architecture used by Nuke's `LazyImage`.

I've also cleaned up the structure of `FixedImageView` to make the logic a bit more sane.

[d977d1f](https://github.com/mlemgroup/mlem/pull/1782/commits/d977d1fd82a8fee496a169f7ee24b17008be4c38) has a super crude testbed in `DeveloperSettingsView` that can be used to test `FixedImageView`.

### DIFF
--- a/Mlem.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Mlem.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -25,7 +25,7 @@
       "location" : "https://github.com/kaishin/Gifu",
       "state" : {
         "branch" : "master",
-        "revision" : "639e40d5dbd86dfa94fc61296c270b49af6f8a2a"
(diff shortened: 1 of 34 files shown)


Sjmarf (approved):
Awesome! It's great to get this one fixed; it's been plaguing us for a while. There's a few more `.id` workarounds that we can remove:

- `PostEditorTargetView:59`
- `PostEditorTargetView:100`
- `ProfileSettingsView:134`

PR #1783, opened 2025-02-21
Save accounts to file when token is refreshed

### LINKED ISSUE #1416: Token unable to be read from keychain for specific accounts

Sjmarf:
Found this whilst looking into #1416. I don't think this'll fix that issue though

### DIFF
--- a/Mlem/App/Globals/Definitions/AccountsTracker.swift
+++ b/Mlem/App/Globals/Definitions/AccountsTracker.swift
@@ -142,6 +142,7 @@ class AccountsTracker {
             $0.name.caseInsensitiveCompare(username) == .orderedSame && $0.api.baseUrl == url
         }) {
             account.updateToken(token)
+            saveAccounts(ofType: .user)
             return account
         } else {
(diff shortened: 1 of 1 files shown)


EricBAndrews (approved):
Nice 👍

PR #1793, opened 2025-02-22
Widget Editor Refinements

### LINKED ISSUE #1749: Info stack in interaction bar editor should be aligned left/right if positioned on left/right
At the moment it's always aligned to center

EricBAndrews:
## Issues

- closes #1749

## Description

The info stack now properly aligns left/right when needed. Getting this to animate smoothly requires a really obnoxious amount of logic because the animation needs to happen immediately, which requires computing the projected info stack location when there's an item being removed.

Duplicate widgets are now also allowed.


Sjmarf (approved):
LGTM! 🚀

PR #1796, opened 2025-02-23
fix eof infinite load

### LINKED ISSUE #1795: Posts just keep loading
### Requirements

### Description

Due to the new EOF fallback. The `.task` gets triggered regardless of whether the EOF is actually visible or not, causing an infinite load loop.

### Mlem Version

_No response_

EricBAndrews:
## Issues

- closes #1795 

## Description

Fixes the previous end-of-feed logic causing an infinite load. The "Load More" button is back, and now appears after 0.5s in the idle state.


Sjmarf (approved):
LGTM 👍

Issue #1737, opened 2025-02-12
Use `MediaView` to back `FixedImageView`


EricBAndrews:
Or, depending on how ergonomic the API would be, just use `MediaView` directly.

### Description

There is a lot of duplicated logic. Unique functions of `FixedImageView`:
- Nuke resizing logic
- Fallback image
- `load(url: URL)` architecture (causes image to update when url updated)

Related:
- #1263 -- this probably won't fix it outright but will move the problem to `MediaView`
- #1511 
- #1743

PR #1798, opened 2025-02-23
Eric/thumbnail load state

### LINKED ISSUE #1797: Tapping thumbnail doesn't always open link
### Requirements

### Description

Seems especially bad on videos

### Mlem Version

_No response_

EricBAndrews:
## Issues

- closes #1797

## Description

Fixes an issue where `loadingState` would not always propagate up through the loading preference key. This caused thumbnails to not open the media view on tap because the `ThumnailImageView` thought the loading state was done. I have implemented two changes to solve this:
- `loader.load()` is now called in a `Task` in an `onChange` instead of in a `task`, which should prevent it from getting cancelled
- Loading state is now handled using an environment object, `MediaLoadingStateTracker`. In #1737 I will unify this into `MediaControlState`


Sjmarf (approved):
Nice 👍

PR #1803, opened 2025-02-26
[#1124] Add french localization

### LINKED ISSUE #1124: French localisation
## Introduction

Today the app provides only english texts. All the strings are hardcoded and not extracted in localizables files.
I propose ot bring localizaiton for both english and french by:
1. extracting all hard-coded strings to a localizable strings dict
2. add french translations

## Requirements

- Extract all hard-coded values
- Define localizable string dict
- Add french translations

## Discussion

Could be quite easy, I can do it in one evening.

## What I Need

Nothing particular, just get your agreement before starting dev, it can induce conflcits between branches so I need to ship this evolution fast.

WDYT @buresdv @EricBAndrews?

pylapp:
## Issues

- Closes #1124 

## Description

Add french support for application.
Translate wordings in french.
Modify a bit Swift code so as to load localized wordings.

## Implementation Notes

You may notice some parts of Swift source code have been modified a bit so as to use localized versions of wordings.
In fact, SwftUI computes the wording to translate at runtime mostly but depending to how you add strings in your views, these strings can be considered as pure string to keep asis, or no localized string key to translated then. The more indirectly you add the string, the string won't get translated. Thus a `localized` method has been defined to force to use the localized version considering the string as wording key. Otherwise it will imply to refactor a lot of views to use the string keys the closest to the views. I use that tricky since eons :-D

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -346,6 +346,7 @@
 		03FE14082BF94FFB00A8377F /* ErrorView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03FE14072BF94FFB00A8377F /* ErrorView.swift */; };
 		03FE140C2BF953B000A8377F /* HandleError.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03FE140B2BF953B000A8377F /* HandleError.swift */; };
 		50C99B562A61D792005D57DD /* Dependencies in Frameworks */ = {isa = PBXBuildFile; productRef = 50C99B552A61D792005D57DD /* Dependencies */; };
+		515B78D82D6F6F4E0027D6D4 /* String+extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 515B78D72D6F6F480027D6D4 /* String+extension.swift */; };
 		6332FDBD27EFAF7C0009A98A /* Settings.bundle in Resources */ = {isa = PBXBuildFile; fileRef = 6332FDBC27EFAF7B0009A98A /* Settings.bundle */; };
 		636250DC2A18111400FC59B4 /* KeychainAccess in Frameworks */ = {isa = PBXBuildFile; productRef = 636250DB2A18111400FC59B4 /* KeychainAccess */; };
 		6363D5C527EE196700E34822 /* MlemApp.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6363D5C427EE196700E34822 /* MlemApp.swift */; };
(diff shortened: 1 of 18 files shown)


Sjmarf (commented):
Nice work :)

It looks like you might have missed translating `"Yes"`?



The label for the "Inbox" tab is quite long, which causes the letters to be squished on some devices.



You're the expert, of course, but I wonder whether there's a better alternative? It doesn't need to be a direct translation. Reddit translates "Inbox" as "Notifications" in French, for example:




If you think "Boîte de réception" is the best option, I'm happy to stick with that :)

pylapp:
Ou that's interesting ideas! I will have a look tomorrow and update the PR with your suggestions.
Let's keep in touch!

Sjmarf:
Sounds good :)

I also noticed a few places where we forgot to include some strings in the localization file (oops). 

I've implemented fixes for these on the `sjmarf/localization fixes2` branch of this repository. If you merge that branch into your PR branch, the additional strings will appear in the Localizable file. You can then filter the Localizable file by choosing "State ≠ Translated" in the search bar to see the new strings that need to be translated.

Screenshot 2025-02-26 at 21 48 57

You'll probably need to build the app (⌘B) in order for the new strings to appear in the Localizable file.

pylapp:
@Sjmarf Ok, seems good now, I added the fix you mentioned and merged your branch before addind new translations.
I compiled the app ([on my own branch](https://github.com/pylapp/mlem/tree/feat/1124-add-french-l10n)), everything is ok.
The [PR branch](https://github.com/pylapp/mlem/tree/pr/1124-add-french-l10n) is updated :)

pylapp:
@Sjmarf Ok, seems good now, I added the fix you mentioned and merged your branch before addind new translations.
I compiled the app ([on my own branch](https://github.com/pylapp/mlem/tree/feat/1124-add-french-l10n)), everything is ok.
The [PR branch](https://github.com/pylapp/mlem/tree/pr/1124-add-french-l10n) is updated :)

Sjmarf (approved):
Nice, looks good to me! I'm happy for you to merge this pull request, if you've got the option to? If not, let me know and I can merge it. Thanks for contributing to Mlem! :D

pylapp:
> Nice, looks good to me! I'm happy for you to merge this pull request, if you've got the option to? If not, let me know and I can merge it. Thanks for contributing to Mlem! :D

Nope I cannot merge this PR as I don’t have *write* access in the repository because I am not part of its team. That’s normal.
Only you and your colleagues are able to do that, I could be nice 😄


IMG_1451

Sjmarf:
Thanks, I've merged it 😄 I'd forgotten how it works lol

PR #1804, opened 2025-02-26
Interaction Bar Report Defaults

### LINKED ISSUE #1781: "Reset" resets to non-report default on report interaction bar editor
### Requirements

### Description

_No response_

### Mlem Version

_No response_

EricBAndrews:
## Issues

- closes #1781 

## Description

The interaction bar editor now properly resets to report defaults in report contexts. This requires a couple of changes:
- `InteractionBarConfiguration` now has `reportDefaults` defined in the protocol. Because this is not meaningful for all configurations, this property is optional.
- `PostBarConfiguration` and `CommentBarConfiguration` expose a non-optional `reportDefaults_`, which backs their protocol-conformant `reportDefaults` and can be used directly in contexts where the type is known
- `InteractionBarEditorView` now takes in an `isReport` flag. When true, it will attempt to reset to `reportDefaults`.


Sjmarf (approved):
Nice 👍 

Just noticed that the resolve button is red rather than green. Might be worth fixing quickly in this PR

IMG_1447

PR #1811, opened 2025-02-28
reset info stack alignment

### LINKED ISSUE #1808: Resetting interaction bar layout doesn't recompute info stack alignment

EricBAndrews:
## Issues

- closes #1808

## Description

Info stack alignment is now recomputed when the configuration is reset.

## Implementation Notes

There is an argument to be made for just setting it to `.center` because we know all of the defaults have widgets on both sides. I've done it with an explicit computation because (a) I don't think the extra compute is significant and (b) it's _technically_ possible that in some future build we might end up with the reset code being able to reset to a "lopsided" configuration, at which point that clever shortcut turns into a gotcha.


Sjmarf (approved):
LGTM 👍

PR #1823, opened 2025-03-02
Fix search refreshing when navigating back

### LINKED ISSUE #1822: Search resets back to first page when navigating back
Repro:

- Search for a community/user
- Scroll down so that the second page loads
- Tap on any item from the second page of results
- Navigate back
- Observe that the second page gets wiped

Sjmarf:
Closes #1822


EricBAndrews (approved):
LGTM! 👍

PR #1827, opened 2025-03-04
Keyword settings import/export

### LINKED ISSUE #1821: Import/Export Keyword Filters
### Requirements

### Description

This was a feature in v1. This should also happen as part of settings export (probably best to handle that after #1820), but I think there is value in having this individually exportable as well--users may have lists they have curated across multiple platforms

Sjmarf:
Closes #1821

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
@@ -49,7 +49,7 @@ struct FiltersSettingsView: View {
                             ) {
                                 navigation.shareInfo = .init(url: url)
                             } else {
-                                ToastModel.main.add(.failure(String("Failed to share error log")))
+                                ToastModel.main.add(.failure())
                             }
                         }
                     }
--- a/Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
@@ -13,8 +13,9 @@ struct FiltersSettingsView: View {
     
     @Setting(\.keywordFilterEnabled) var keywordFilterEnabled
     
-    @Environment(Palette.self) var palette
     @Environment(FiltersTracker.self) var filtersTracker
+    @Environment(Palette.self) var palette
+    @Environment(NavigationLayer.self) var navigation
     
     @State var newKeyword: String = ""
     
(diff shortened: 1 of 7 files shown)


Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
```diff
@@ -37,6 +38,32 @@ struct FiltersSettingsView: View {
...
+                Menu("More...", systemImage: Icons.menuCircle) {
+                    Button("Export...", systemImage: Icons.share) {
+                        Task {
+                            if let url = await downloadTextToFileSystem(
+                                fileName: "keywords.txt",
+                                text: filtersTracker.filteredKeywords.joined(separator: "\n")
+                            ) {
+                                navigation.shareInfo = .init(url: url)
+                            } else {
+                                ToastModel.main.add(.failure(String("Failed to share error log")))
```
EricBAndrews:
```suggestion
                                ToastModel.main.add(.failure(String("Failed to export keyword filters")))
```

Mlem/App/Views/Shared/Navigation/NavigationLayer.swift
```diff
@@ -173,6 +175,27 @@ class NavigationLayer: Identifiable {
...
     
+    @MainActor
+    func showFilePicker(types: [UTType], callback: @escaping (Data) async -> Void) {
+        model?.contentPickerTracker.showingFilePicker = true
+        model?.contentPickerTracker.filePickerContentTypes = types
+        model?.contentPickerTracker.filePickerCallback = { url in
+            Task {
+                do {
+                    guard url.startAccessingSecurityScopedResource() else {
+                        throw ApiClientError.insufficientPermissions
```
EricBAndrews: This shouldn't be an `ApiClientError` since it's not `ApiClient`-related--I think `MlemError` would be more appropriate

EricBAndrews (changes requested):
Nice work! Putting the actions in the nav bar is much cleaner than the v1 buttons.

If feasible, there should be a confirmation toast when the export succeeds, but I'm fine leaving that OOS if it's complicated--the share logic is already a bit gnarly, hooking up a callback to it may be more work than it's worth right now.

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
Sjmarf: Oops, missed this one lol. Can't replace it with that because we've gone through localization already... just "Failed" is probably enough though, given the context

Mlem/App/Views/Shared/Navigation/NavigationLayer.swift
Sjmarf: Makes sense 👍

Sjmarf:
> If feasible, there should be a confirmation toast when the export succeeds, but I'm fine leaving that OOS if it's complicated--the share logic is already a bit gnarly, hooking up a callback to it may be more work than it's worth right now.

The time-consuming part of the export (writing the keyword to a file) occurs *before* the share sheet opens. I don't think we need to show a "success" toast, because the share sheet opening implies that it's succeeded.

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
EricBAndrews: Works for me

EricBAndrews (approved):
LGTM! 🚀

Issue #1828, opened 2025-03-04
Developer Mode QoL


EricBAndrews:
### Requirements

- [x] There are no existing issues for this feature.
- [x] This is a request for a **single** feature (create multiple issues for multiple feature requests).

### Description

Minor QoL changes to make developer mode easier to daily drive:
- Move image type label out of the way
- Add option to disable image type label entirely

Issue #1790, opened 2025-02-22
Haptic player error


Sjmarf:
Got this when switching back to Mlem from another app

```
Optional(Error Domain=com.apple.CoreHaptics Code=-4810 "(null)" UserInfo={Error =_player != nil}) (Mlem/HapticManager.swift, startEngine():51)
```

EricBAndrews:
Did haptics continue to work after this error?

Sjmarf:
Didn't check, sorry :/

EricBAndrews:
I'm having some trouble getting this to happen--I think it's probably one of those things where you have to catch the engine in just the wrong state. If the haptics continue to work, we should just silence the error so it doesn't pop a toast; otherwise the engine needs some more sophisticated failure recovery

EricBAndrews:
I got a repro! Haptics stop working after this error occurs.

EricBAndrews:
Moving this off 2.0 because #1829 should at least band-aid fix it, and even if it doesn't the error is super rare and doesn't break core functionality (if a user _needs_ those haptics, they can always just restart the app).

Keeping this issue open until either we figure out the root cause or it doesn't show up for long enough that we can call it squashed.

EricBAndrews:
I haven't seen this in over a month--closing for now.

PR #1829, opened 2025-03-04
Better haptic failure recovery

EricBAndrews:
## Issues

- progress towards #1790 

## Description

Adds some more sophisticated failure recovery to `startEngine()`--if the engine fails to start, the catch block now re-creates the engine from scratch. The initial error is now silent, but will show a haptic if developer mode is enabled. I also have [a couple changes planned](https://github.com/mlemgroup/mlem/issues/1828) to make developer mode less obnoxious to daily drive, so if it pops up again it should be easy enough to spot.

This issue is insanely hard to reproduce, so I don't know if any of this will actually fix it.


Sjmarf (approved):
LGTM 👍

PR #1831, opened 2025-03-05
Minor l12n fix

Sjmarf:
We were using the string `Reason: ` (with a trailing space). The french translation didn't have the trailing space, so it was displayed incorrectly. I've fixed this by removing the trailing space for the english translation, and hard-coding that space instead.


EricBAndrews (approved):
LGTM! 👍

PR #1838, opened 2025-03-08
Rewrite palette system

Sjmarf:
Rewrote the palette system and moved palette-related code into a package called `Theming`.

There is a new `ThemedColor` type. Previously, we defined quite a few enums with a computed property `color`, like this:

```swift
// This is just an example and doesn't actually exist in the codebase
enum PositiveOrNegative {
    case positive, negative

    var color: Color {
        switch self {
            case positive: Palette.main.positive
            case negative: Palette.main.negative
        }
    }
}
```

In this PR, we now use `ThemedColor` instead of `Color` as the return type.

```swift
enum PositiveOrNegative {
    case positive, negative

    var color: ThemedColor {
        switch self {
            case positive: .themedPositive
            case negative: .themedNegative
        }
    }
}
```

`ThemedColor` conforms to `ShapeStyle`, so can be passed as an argument to `.foregroundStyle()`, `.background()`, `.fill()`, etc.

Before, we'd write something like this to apply a color to a view:

```swift
struct TestView: View {
    @Environment(Palette.self) var palette

    var body: some View {
        Text("Hello world")
            .foregroundStyle(palette.accent)
    }
}
```

After this PR, we write this instead:

```swift
struct TestView: View {
    var body: some View {
        Text("Hello world")
            .foregroundStyle(.themedAccent) // themedAccent is a static property defined on ThemedColor
    }
}
```

When the `ThemedColor` is resolved to a `Color`, it reads a `Palette` object from the environment. You can access this object directly in situations where a `Color` object is needed rather than a `ThemedColor`.

```swift
@Environment(\.palette) var palette
```

`Palette` is essentially just a renaming of the previous `ColorPalette` type. In future, we might want to allow custom themes. `Palette` could be made `Codable` reasonably easily, so that we can create one from a JSON file.

There is no longer a `Palette.main` singleton - the active `Palette` may only be accessed from the view environment.


EricBAndrews (approved):
LGTM! 🔥

PR #1854, opened 2025-03-11
Sharing options

### LINKED ISSUE #639: Advanced sharing options
Setting to change the instance with which posts/users/communities are shared. The available options could be something like:

- Instance the post was made on (this is what is used right now)
- The user's instance
- ~~lemmy.guide~~ Edit: This closed down :(
- Ask every time

### LINKED ISSUE #1395: Share URL should share from user's instance
Currently the share URL appears to be the link from the poster's instance. It should instead share from the user's.

Sjmarf:
> [!note]
> Relies on [this MlemMiddleware PR](https://github.com/mlemgroup/MlemMiddleware/pull/133).

Closes #639, closes #1395.

Settings -> Links -> Share Links

 The default behavior is now to share from the user's instance rather than the author's instance as it was before.


Mlem.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -59,8 +59,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/mlemgroup/MlemMiddleware",
       "state" : {
-        "revision" : "bd13585a4941ef04cad6943a3ed576c9f4904b11",
-        "version" : "0.83.0"
+        "branch" : "sjmarf/sharing",
+        "revision" : "edfba5745c5ddbcd0137afd24687db85e168a32e"
```

Mlem/App/Views/Shared/ShareInstancePickerView.swift
```diff
@@ -0,0 +1,138 @@
...
+            NavigationModel.main.shareInfo = .init(
+                url: resolvedEntity.url(),
+                actions: entity.shareSheetActions()
+            )
+            ToastModel.main.removeToast(id: toastId)
+            print("END")
+        } catch {
+            print("ERROR", error)
+            ToastModel.main.removeToast(id: toastId)
+            handleError(error)
```
EricBAndrews: *nit*

Debug prints?

Mlem/App/Views/Root/Tabs/Settings/SharingLinksSettingsView.swift
```diff
@@ -0,0 +1,38 @@
...
+
+struct SharingLinksSettingsView: View {
+    @Setting(\.linkSharingMode) var linkSharingMode
+    
+    var body: some View {
+        Form {
+            Section("Share links using...") {
+                Picker("Share links using...", selection: $linkSharingMode) {
+                    ForEach(LinkSharingMode.allCases, id: \.self) { mode in
+                        Text(mode.label)
```
EricBAndrews: *nit*

These should have conditional icons--"house," "signature," "questionmark.circle"?

EricBAndrews (approved):
LGTM! 🚀 

A couple nits, nothing blocking.

Mlem/App/Views/Root/Tabs/Settings/SharingLinksSettingsView.swift
Sjmarf: Now that I'm looking at this again, I realise that "Author's instance" isn't the best wording because it doesn't work in the context of sharing a community or a user. I've renamed it to "host instance" instead.

I'm struggling to think of a good way to separate "My instance" and "host instance" using icons... any ideas? `"signature"` doesn't work as well anymore, now that I've renamed "Author's instance" -> "Host instance". The other option is just to leave the icons out.

EricBAndrews: "House" and "building.2"? Maybe confusing since "house" is often community but contextually probably clear enough

Sjmarf: Which is which? 😅

The point of the icons setting is that it makes it easier to scan for the right option. If it's as ambiguous as it is here, that isn't possible. It might be better to not have icons at all. Or, use the same icon for both?

Sjmarf: Now that I think of it, we could use the avatars of the instances

Sjmarf: Thoughts?

EricBAndrews: Instance avatars makes sense to me 👍

PR #1856, opened 2025-03-11
Use `MediaView` for all* media

### LINKED ISSUE #1665: Partially transparent gifs aren't rendered properly
This happens in gifs where some pixels are transparent, and which pixels are transparent changes each frame. The first frame of the gif appears to be underlaid underneath the gif for the duration of playback. 

I've seen this a couple of times. Next time I see it I'll add an example link

### LINKED ISSUE #1737: Use `MediaView` to back `FixedImageView`

### LINKED ISSUE #1743: Use thumbnail URL for compact media thumbnails
### Requirements

### Description

Videos currently don't show a thumbnail. If we include the original thumbnail URL in the `.media` media type, we can use that instead of the fallback image.

EricBAndrews:
*except `SimpleAvatarView`

## Issues

- closes #1737 
- closes #1743 (renders irrelevant)
- closes #1665 
- progress towards #1828 

## Description

This PR removes `FixedImageView` and instead uses `MediaView` for all image rendering. It also contains some substantial `MediaView` changes.

### Video Decoding

The standard NukeVideo decoder has been replaced with `MlemNukeVideoDecoder`, a wrapper that ensures that a preview image is always generated. This means that movie-type videos now have thumbnails! It also improves performance somewhat by moving thumbnail generation and scaling to the image pipeline.

### Animated Media

This PR gets rid of the old system where the animated media was an overlay atop the preview image. This fixes the transparent gifs problem and also simplifies some of the `MediaView` logic since we no longer need to track `playing` and `animating` separately.

### `MediaControlState`

`MediaControlState` has been significantly expanded. It is now responsible for tracking blurring, loading state, and overlays. The core criterion of what gets tracked by `MediaControlState` is whether that piece of information needs to be available in deeply nested views (note that overlay information is important for `AnimationControlLayer`).

The relevant `MediaView` parameters have accordingly been moved into `MediaControlState`.

The overlay API is now simplified as well; instead of parameters for each overlay, overlays are now specified using a set of the new `MediaView.Overlay` enum.

### MediaView Fallbacks

`MediaView` now has the robust fallback behavior of `FixedImageView`. Depending on which `MediaView.Fallback` is specified, if no image loads, a fallback will be displayed.

### `AspectRatioBounds`

`MediaView` now has the ability to bound aspect ratios vertically and horizontally (i.e., specify a min and max height) or to a fixed aspect ratio.

### Navigation

`MediaView` expects a `NavigationLayer` to be present in the environment, but this isn't always the case. Passing a `NavigationLayer?` in proves incredibly cumbersome, so I have instead defined the environment key `navigationContext` which offers a `NavigationLayer?`. All instances of `.environment(layer)` have been replaced with `.navigationEnvironment(layer)`, a simple view modifier which populates both the implicit `NavigationLayer.self` environment key and the new `navigationContext` key.

### Refactoring

All the key `MediaView`-related stuff is now all in `Core`, and I've pulled a couple types into `MediaView`. There's still a lot of cleanup that could be done, but everything should at least be in a sane location now.


Mlem/Localizable.xcstrings
```diff
@@ -4646,6 +4646,9 @@
           }
         }
       }
+    },
+    "Navigation present" : {
```
Sjmarf: I presume this is from a previous version? Can't find where this is from. Building again should remove it, if so

Sjmarf (approved):
Nice work! iirc it's possible to make the regular environment object optional, in which case it'll be `nil` if no environment object is present

```swift
@Environment(NavigationLayer.self) var navigationLayer: NavigationLayer?
```

Then, we wouldn't need the extra environment value entry.

Sjmarf (changes requested):
(oops, meant to rc)

Sjmarf:
Also the image fallback is larger and more prominent, now. I think it looked better when it was smaller and lighter, like before

Mlem/Localizable.xcstrings
EricBAndrews: Yeah, this was part of a debug overlay.

Sjmarf (approved):
Nice 🚀 🚀

Sjmarf:
Getting some slightly strange behavior when dismissing an image with an error using the image viewer 🤔 

https://github.com/user-attachments/assets/dc11b79d-973e-47c4-adda-fd1a3e314506

EricBAndrews:
I've updated it so that the fallback isn't shown if the error overlay is shown--that fixes the above issue, but isn't the cleanest possible solution. I'll make it nicer in [a later PR](https://github.com/mlemgroup/mlem/issues/1861).

PR #1864, opened 2025-03-14
Add mlemUrlRequest logic to MediaLoader

EricBAndrews:
Follow-up to #1857. Images loaded on-the-fly (without prefetching; e.g., avatars in search results) weren't getting the `User-Agent` header, so weren't loading from lemmy.ml.

`MediaLoader` now uses the `urlRequest` form of `ImageRequest` and adds the `User-Agent` header.


Sjmarf (approved):
LGTM 👍

PR #1866, opened 2025-03-14
Animated Avatars

### LINKED ISSUE #1511: Animated profile pictures
Some people have animated media as their profile picture. These should properly play rather than rendering as a still image with a play button. 

Users may find animated avatars distracting or annoying, so there should also be a toggle to render them as still images instead.

EricBAndrews:
## Issues

- closes #1511 

## Description

Adds support for animated avatars. They can be toggled on/off Settings -> General with the other avatar settings.

## Implementation Notes

Includes a minor change to the `MediaView` API:
- If `MediaControlState` is initialized with `animating: true`, that media will autoplay; otherwise it will not.
- `MediaView.largeImage` now provides `Settings.main.autoplayMedia` as the `animating` parameter


Mlem/App/Views/Root/Tabs/Settings/GeneralSettingsView.swift
```diff
@@ -87,6 +88,9 @@ struct GeneralSettingsView: View {
             Section {
                 Toggle("User Avatar", systemImage: Icons.personCircle, isOn: $showPersonAvatar)
                 Toggle("Community Avatar", systemImage: Icons.communityCircle, isOn: $showCommunityAvatar)
+                if #available(iOS 18, *) {
+                    Toggle("Animated Avatars", systemImage: Icons.playCircle, isOn: $animatedAvatars)
+                }
```
Sjmarf: I wonder whether this would be better in Accessibility 🤔 

Do we want this to be a three-way setting?
```
Animate avatar...
- Always
- In profile only
- Never
```

And do we want to consider the system "reduce motion" option?

Sjmarf (changes requested):
Ooh, fancy! 

Looks like aspect ratios aren't preserved here:

Screenshot 2025-03-14 at 21 46 04

Mlem/App/Views/Root/Tabs/Settings/GeneralSettingsView.swift
EricBAndrews: > Do we want this to be a three-way setting?
👍 

Re: a11y: that's a good point. I've updated it to default to `.never` if system-wide reduce motion is enabled and added it to the a11y settings page.

I do think we should also keep it in general, since that's where all the other avatar settings live.

Sjmarf:
The aspect ratio is fixed now, but it seems to be cropped too far in.



There's also this visual bug (I reckon this is a result of the aforementioned issue, though)

Mlem/App/Views/Root/Tabs/Settings/GeneralSettingsView.swift
Sjmarf: Nice, looks good 👍 

There's also an argument that it could go in "Media & Links", where "autoplay" is. I'm not sure what the best option is. Happy to stick with what we've got now if you're happy.

Mlem/App/Views/Root/Tabs/Settings/AccessibilitySettingsView.swift
```diff
@@ -43,6 +44,20 @@ struct AccessibilitySettingsView: View {
...
+                        "Animated Avatars",
+                        value: .init(localized: animatedAvatars.label),
+                        fallbackValue: "",
+                        systemImage: Icons.playCircle,
+                        destination: .settings(.animatedAvatars)
+                    )
+                }
+            } header: {
+                Text("Reduce Motion")
+            }
```
Sjmarf: This hides the setting on iOS 17, but the section header is still shown. Fix:

```suggestion
            if #available(iOS 18, *) {
                Section {
                    NavigationLink(
                        "Animated Avatars",
                        value: .init(localized: animatedAvatars.label),
                        fallbackValue: "",
                        systemImage: Icons.playCircle,
                        destination: .settings(.animatedAvatars)
                    )
                } header: {
                    Text("Reduce Motion")
                }
            }
```

Sjmarf (approved):
LGTM! 🚀

PR #1886, opened 2025-03-17
Fix emojis sometimes being rendered as full-width images

### LINKED ISSUE #1855: Emojis from hexbear appear at incorrect scale
### Requirements

### Description

Hexbear uses the lemmy embedded image feature (afaik) to create emojis. On mlem they often don’t load or if they do they load and incorrect scales. Apollo has managed to get them to load at a consistent scale though I’m not sure how. 

I have attached some comparisons between mlem, Apollo, and the original website below:

Image
Image
Image

———

Image
Image
Image
Image

While Apollo renders them slightly differently than the website but I actually think it’s less obstructive, but that’s personal opinion.

### Mlem Version

_No response_

Sjmarf:
If a comment contained only an emoji, and nothing else, it would incorrectly be rendered as a full-size image.

Fixes #1855

([LemmyMarkdownUI changes](https://github.com/mlemgroup/LemmyMarkdownUI/pull/60/files))


EricBAndrews (approved):
LGTM! 💯

PR #1887, opened 2025-03-17
Fix comment not always being inserted into tree

### LINKED ISSUE #1740: Newly created comment not always inserted into comment tree
Just had this happen. Not sure what caused it. It seems to work correctly when I try in the mlemtesting comm

Sjmarf:
Closes #1740


EricBAndrews (approved):
Nice 👍

Issue #1883, opened 2025-03-17
Better webp, gif rendering


EricBAndrews:
### Requirements

- [x] There are no existing issues for this feature.
- [x] This is a request for a **single** feature (create multiple issues for multiple feature requests).

### Description

Webp and gif rendering is currently handled by `SDWebImageSwiftUI`'s `AnimatedImage`, which works OK out of the box but needs some janky hacks to play nice with scrubbing and suffers from performance issues when scrubbing backwards. This view should be rewritten as a custom `UIViewRepresentable` around the core `SDAnimatedImageView` to offer better performance and more elegant scrubbing.

This should also enable displaying runtime and current play time on the playback bar.

PR #1890, opened 2025-03-17
Improved Scrubbing

### LINKED ISSUE #1834: Improved GIF/Video Scrubbing
### Requirements

### Description

Follow-up to #173 focused on enhancing the scrubbing UI:
- Adjustable scrubbing speed
- Swipe-to-scrub

EricBAndrews:
## Issues

- closes #1834 

## Description

Refines the scrubbing UI:
- Scrubbing is now done with a horizontal swipe gesture
- Sliding your finger up while scrubbing will increase the scrub rate, down will decrease it. Minimum scrub rate is 1/8, max is 8.
- Made the playback bar prettier

## Implementation Notes
- Also adds a "No Debug" product scheme that launches the app without attaching a debugger


Mlem/App/Views/Pages/ImageViewer.swift
```diff
@@ -153,6 +151,87 @@ struct ImageViewer: View {
...
+        }
+        
+        if dragIsScrub == nil {
+            dragIsScrub = abs(value.velocity.height) < abs(value.velocity.width)
+        }
+        
+        guard let dragIsScrub else {
+            assertionFailure("DragGesture updated but dragIsScrub is nil")
+            return
+        }
```
Sjmarf: Could we do something like:

```suggestion
        let dragIsScrub = self.dragIsScrub ?? abs(value.velocity.height) < abs(value.velocity.width)
        self.dragIsScrub = dragIsScrub
```

Mlem/App/Views/Pages/ImageViewer+Views.swift
```diff
@@ -139,25 +141,15 @@ extension ImageViewer {
...
-                        .highPriorityGesture(DragGesture()
-                            .onChanged { value in
-                                if value.location.x >= 0, value.location.x <= draggableWidth {
-                                    controlState.scrubTarget = value.location.x / draggableWidth
-                                }
-                            }
-                            .onEnded { _ in
-                                controlState.scrubTarget = nil
-                            }
-                        )
```
Sjmarf, line 160: IMO we should retain a separate gesture handler for the playback bar. When I drag the circle to scrub, it feels like the circle isn't sticking to my finger as perfectly as before.

Sjmarf (changes requested):
Nice work! I think it would be worth showing timestamps on either end of the playback bar, though feel free to leave that as OOS.

EricBAndrews:
Swipe-to-dismiss freezes if the swipe gesture starts sort of sideways (exponential curve shaped gesture). Commenting so I remember to fix this.

Mlem/App/Views/Pages/ImageViewer.swift
EricBAndrews: That is nicer 👍

Mlem/App/Views/Pages/ImageViewer+Views.swift
EricBAndrews: That's a good point. The `simultaneousGesture` on the existing drag handler makes adding a separate handler nasty, so I've just piggybacked on the existing handler and adjust the behavior according to whether it's a bar scrub or a "free" scrub. Scrubbing from the bar now tracks the gesture precisely, disables variable scrub rate, and snaps the scrub to your finger at the start of the gesture.

I'm leaving time OOS because it doesn't look like there's a straightfoward way to get that information out of `SDWebImage`, at least not without #1883

Sjmarf:
Nice, works perfectly 👍 iirc `simultaneousGesture` has a gesture-masking parameter, which might be a solution to the simultaneous gesture issue? I've never used it though

Sjmarf (approved):

PR #1892, opened 2025-03-18
Add navigation titles to all settings pages

### LINKED ISSUE #1870: Add navigation title to all settings pages
At the moment, we omit the `.navigationTitle()` for settings pages where we don't want the title to be shown. This has the unfortunate side effect of making the back button just say "Back" rather than the title of the previous page. I'm guessing that VoiceOver also uses the navigation title value.

I think the `.toolbar(removing: .title)` modifier can be used to remove just the title at the top of the page, while still allowing the title value to be used in other places.

Sjmarf:
Closes #1870


EricBAndrews (approved):
Nice 👍

PR #1895, opened 2025-03-18
Mixed feed comments design

### LINKED ISSUE #1837: Make posts and comments more visually distinct in combined feeds
From here: https://lemmy.ml/post/26903523/17141628

Possibly solvable by adding the comment indent bar in combined feeds too? But make it gray or something? I'm not sure, would have to prototype some ideas. Another option would be to rethink the way that we display the post context for the comment - if we showed it at the top rather than the bottom, it might be more obvious that it's a comment replying to something.

Sjmarf:
Closes #1837

Moved the post context for comments to the top rather than the bottom.

I think having the comment bar always shown for comments is a good distinguishing mark to separate them from posts. The comment bar is now shown for top-level comments on a post, and for comments in a combined feed. 

If this change is well-received I'd like to expand this design to the Inbox and Mod Mail, too.

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Feeds/Feed Comments/FeedCommentView.swift
+++ b/Mlem/App/Views/Root/Tabs/Feeds/Feed Comments/FeedCommentView.swift
@@ -16,7 +16,8 @@ struct FeedCommentView<EmbeddedContent: View>: View {
     @Environment(\.reportContext) var reportContext: Report?
     
     @Setting(\.postSize) var settingsPostSize
-    
+    @Setting(\.blurNsfw) var blurNsfw
+
     let comment: any Comment
     var overriddenSize: PostSize?
     @ViewBuilder var embeddedContent: () -> EmbeddedContent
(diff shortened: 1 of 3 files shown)


Mlem/App/Views/Root/Tabs/Feeds/Feed Comments/FeedCommentView.swift
```diff
@@ -53,6 +60,43 @@ struct FeedCommentView<EmbeddedContent: View>: View {
...
+        }
+    }
+    
+    @ViewBuilder
+    func headerView(post: any Post) -> some View {
+        HStack {
+            MediaView(
+                url: headerUrl(post: post),
+                size: .init(width: 40, height: 40),
+                aspectRatioBounds: .absoluteSquare,
```
EricBAndrews:
```suggestion
                size: .init(width: 40, height: 40),
                controlState: .constant(.init(
                    blurred: post.nsfw && blurNsfw != .never,
                    animating: false,
                    overlays: []
                )),
                aspectRatioBounds: .absoluteSquare,
```

EricBAndrews (changes requested):
Looks great!

One minor suggestion re: media control state, plus design refinements per Slack.

EricBAndrews (approved):
LGTM! 🔥

PR #1901, opened 2025-03-19
Improved Gifs

### LINKED ISSUE #1883: Better webp, gif rendering

EricBAndrews:
## Issues

- closes #1883 

## Description

- Improves scrubbing performance for gifs and webps. Scrubbing backwards on large gifs is now nice and smooth (https://sh.itjust.works/post/33262204 being a nice testing image)
- Adds time markers to playback bar

## Implementation Notes

- Rewrites `AnimatedImageView` as a `UIViewRepresentable` around `SDAnimatedImageView`. This gives us access to `SDAnimatedImage.preloadAllFrames()`, loads the entire gif into memory and thus enables smooth backwards scrubbing. It also simplifies a lot of the logic around interacting with the `AnimatedImageView`.
- Updates how `AnimatedImageView` populates `controlState.playbackPosition`; there is now an observer on the `SDAnimatedImagePlayer` that listens to frame changes.
- Adds `scrubbingAvailable` to `MediaControlState`. This lets us only perform `preloadAllFrames` when we actually need to.
- Adds `duration` to `MediaControlState`


Mlem/App/Utility/Extensions/TimeInterval+Extensions.swift
```diff
@@ -0,0 +1,22 @@
...
+//  TimeInterval+Extensions.swift
+//  Mlem
+//
+//  Created by Eric Andrews on 2025-03-19.
+//
+// Adapted from https://stackoverflow.com/a/30772571
+
+import Foundation
+
+extension TimeInterval {
```
Sjmarf: `TimeInterval` is a typealias of `Double`, which means that these computed properties will also be available on all `Double` values. Do we want that? I wonder whether there's a better way to do this

Sjmarf (approved):
Nice, this is super smooth! One minor implementation question

Mlem/App/Utility/Extensions/TimeInterval+Extensions.swift
EricBAndrews: That's a good point--I've moved this logic into `MediaControlState`, since we don't need it anywhere else. If we end up needing it elsewhere we can look for an elegant general solution.

Mlem/App/Views/Shared/Images/Core/MediaControlState.swift
```diff
@@ -84,4 +89,11 @@ class MediaControlState {
         self.muted = muted ?? Settings.main.muteVideos
         self.scrubbingAvailable = scrubbingAvailable
     }
+    
+    private func minuteSecondString(from timeInterval: TimeInterval) -> String {
```
Sjmarf, line 93: Possibly doable using `Duration` https://goshdarnformatstyle.com/duration-styles/#time-style

EricBAndrews: Ooh, nice find--implemented

Mlem/App/Utility/Extensions/TimeInterval+Extensions.swift
Sjmarf: LGTM 👍

PR #1905, opened 2025-03-20
Improve hitboxes

### LINKED ISSUE #1891: Image hitbox extends beyond bounds
Tall images can be tapped outside of their clipped box. This is especially noticeable in tile mode. We had a similar issue in v1--easily fixed by disabling hit registration on the image itself and adding a clear overlay to the view that registers gestures

EricBAndrews:
## Issues

- closes #1891 (good repro: https://lemmy.ml/post/27416939 in tile post mode)

## Description

Also makes the playback bar hitbox larger and extend a little bit below the visible bar.


Sjmarf (approved):
Nice 👍

PR #1914, opened 2025-03-22
"Jump to last position" button in expanded post view

### LINKED ISSUE #1912: "Scroll to last position" button next to jump button in the post page
This button would be shown if it's not your first time seeing the post. Tapping it would scroll to the last comment you were looking at. This would make it easier to find your place again if you accidentally tap to go back when looking at the post.

This could act as a less-than-ideal alternative to #174.

Sjmarf:
Added a button that takes you to the comment you were looking at last time you were at the page. This only has a short-term memory of 10 posts at the moment, which doesn't persist across restarts. This is intended to be a fix for the scenario where a user accidentally taps back from the post and loses their place.

Also changed the design of the jump button a bit - what do you think? Is this better or worse?

Closes #1912

### DIFF
--- a/Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView.swift
+++ b/Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView.swift
@@ -175,7 +175,7 @@ struct ExpandedPostView<Content: View>: View {
                 .overlay(alignment: jumpButton.alignment) {
                     let shouldShow = topVisibleCommentAtLastVisit == nil && (post.commentCount_ ?? 0) > 10
                     JumpButtonsView(
-                        showJumpButton: true,
+                        showJumpButton: (tracker?.nodes.count ?? 0) > 1,
                         topVisibleItem: topVisibleItem,
                         scrollToLastVisitedPosition: shouldShow ? nil : scrollToLastVisitedPosition,
                         scrollToNextComment: scrollToNextComment,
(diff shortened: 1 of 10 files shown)


Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView.swift
```diff
@@ -256,3 +267,34 @@ struct ExpandedPostView<Content: View>: View {
...
+    var body: some View {
+        VStack(spacing: 0) {
+            if let scrollToLastVisitedPosition, topVisibleItem.isAtPost {
+                JumpButtonView(
+                    systemImage: "chevron.down.2",
+                    onShortPress: scrollToLastVisitedPosition,
+                    onLongPress: nil
+                )
+            }
+            if jumpButton != .none, showJumpButton { // (tracker?.nodes.count ?? 0) > 1
```
EricBAndrews: *nit*

Dead code?

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView.swift
```diff
@@ -256,3 +267,34 @@ struct ExpandedPostView<Content: View>: View {
...
+    
+    var scrollToLastVisitedPosition: (() -> Void)?
+    var scrollToNextComment: () -> Void
+    var scrollToPreviousComment: () -> Void
+    
+    var body: some View {
+        VStack(spacing: 0) {
+            if let scrollToLastVisitedPosition, topVisibleItem.isAtPost {
+                JumpButtonView(
+                    systemImage: "chevron.down.2",
```
EricBAndrews: *nit*

`Icons`

EricBAndrews (approved):
LGTM, works perfectly! 🔥 

A couple minor nits, plus design considerations in Slack

PR #1915, opened 2025-03-22
Local MlemMiddleware

Sjmarf:
Moved MlemMiddleware to a local package. MlemApiTypes is still a submodule. I'll archive the MlemMiddleware repo when this merges.

I've disabled SwiftLint for the MlemMiddleware folder for now - eventually we should turn this back on, but at present there are a lot of errors (including dozens in MlemApiTypes), so I think it's easier to leave it out of this PR and fix it later

### DIFF
--- a/.github/workflows/ci_build_16.yml
+++ b/.github/workflows/ci_build_16.yml
@@ -19,7 +19,7 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v3
       with:
-        submodules: true
+        submodules: 'true'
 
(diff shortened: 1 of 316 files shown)


EricBAndrews (changes requested):
It looks like the CI isn't picking up the submodule. I think adding `submodules: true` to the checkout action in `workflows/ci_build_16.yml` will fix this.

EricBAndrews (approved):
LGTM! 💯

PR #1918, opened 2025-03-28
Customizable swipe actions

### LINKED ISSUE #1409: Customizable swipe actions
Ability to rearrange the swipe actions, similar to how interaction bar widgets can be rearranged.

Sjmarf:
Closes #1409

I think it makes sense to group the interaction bar and swipe action configurations in the same struct. Our current system is a little annoying to use, though... in future I think we should reconsider how these settings are stored. Having to pass `isReport` everywhere is annoying, but I can't think of a simple way around it


EricBAndrews (approved):
LGTM! 🔥 

The icons in the config page look great!

I think it would be nice to have a delete button at the top level--the swipe is a little clunky. Fine to leave OOS, though, since I suspect this UI will be reworked somewhere down the line.

Sjmarf:
> I think it would be nice to have a delete button at the top level--the swipe is a little clunky. 

How's this?



> I suspect this UI will be reworked somewhere down the line.

Yeah, in theory it would be nice to have a fancy UI like we do for the interaction bar. That said, I'm wary of sinking too much time into Settings UIs, lol. The interaction bar editor has been rewritten from the ground up 4 times (I think) and in each case it's taken quite a lot of time. We've probably put dozens of hours into a screen that the average person will only use once, if at all. We got there in the end, but doing it all again for swipe actions sounds like it might not be worth the time-value proposition (at least for now)

EricBAndrews (approved):
LGTM! 🔥

PR #1920, opened 2025-03-29
TestFlight update banner

### LINKED ISSUE #1913: For TestFlight users, show a "TestFlight updated!" banner at the top of the feed when TF updates
I'm imagining a small banner that says "TestFlight updated!" with a "What's changed?" button that links to the post.

TF has a built-in changelog popup system, which we don't use at the moment. I think that's the right decision, because it's rather obtrusive imo. Having a banner at the top of the feed is ignorable, which I think is better.

Sjmarf:
Closes #1913

When the version number changes, a banner will be shown at the top of the feed (but only if you're on the TestFlight). Swipe or long press to dismiss the banner. If you tap "What's New?", Mlem shows the most recent post in c/mlemapp from an Mlem developer that includes "testflight" in the title. There's a "reset feed testflight banner" button in developer settings for testing convenience.

Due to the way that I'm checking whether the app is on the TestFlight, the banner won't show up on a simulator


Mlem/App/Views/Root/Tabs/Feeds/Components/UpdateBannerView.swift
```diff
@@ -0,0 +1,87 @@
...
+    
+    func submit() {
+        isLoading = true
+        Task { @MainActor in
+            let community: Community2 = try await appState.firstApi.getCommunity(url: URL(string: "https://lemmy.ml/c/mlemapp")!)
+            // Assuming the update announcement is pinned, it'll probably be one of these 5
+            let posts = try await community.getPosts(sort: .new, limit: 5).posts
+            let announcementPost = posts.first { post in
+                post.creator.isMlemDeveloper && post.title.lowercased().contains("testflight")
+            }
```
EricBAndrews: *nit*

This work doesn't need to happen on the main actor--it's marginally better for performance to remove the `@MainActor` annotation from the task as a whole and just perform the view updates on the main actor

EricBAndrews (changes requested):
Nice work!

The fetch logic breaks if the most recent TestFlight announcement is not pinned but there is another pinned post with TestFlight in the name due to pinned posts _always_ coming first regardless of actual sort order. I've unpinned the current TestFlight announcement so you can repro this consistently--the banner now goes to the first community update. Two changes would make this logic far more robust:
- Manually sort the returned posts by creation date before grabbing the first
- Check for the full `[ TestFlight Update ]` tag, since we always include that in the title

Sjmarf:
> The fetch logic breaks if the most recent TestFlight announcement is not pinned but there is another pinned post with TestFlight in the name

That's a good point. Even if we do pin the post, there's also a good chance it doesn't federate properly to all instances. I'll implement the fix you suggested 👍

Mlem/App/Views/Root/Tabs/Feeds/Components/UpdateBannerView.swift
Sjmarf: This task will run on the main actor anyways, even without the `@MainActor` annotation, because the view itself is isolated to the main actor. I can unisolate it completely using `Task.detached`?

Admittedly I don't fully know how actor isolation works. I watched an Apple Developer video a while back where they were running some pretty intensive stuff on the main actor. I suspect that I need to do some reading up on what it actually is

EricBAndrews (approved):
LGTM! 🚀

PR #1926, opened 2025-03-30
Fix wrong color background being visible when keyboard opens

Sjmarf:
Fixed a white background being visible briefly above the keyboard whilst the keyboard is opening:

https://github.com/user-attachments/assets/eceb36e4-9be1-4079-8bb4-7ac84490b501


EricBAndrews (approved):
LGTM! 🔥

PR #1928, opened 2025-03-31
Fix Animated Media Shaking

### LINKED ISSUE #1906: Animated media shakes when zooming out

EricBAndrews:
## Issues

- closes #1906 

## Description

Replaces `ZoomableContainer` with a custom UIKit-based zoom and pan solution. This has a few advantages:
- Stops the internal zoomable scroll view from interfering with scroll handlers (i.e., pausing the video sometimes when scrolling around)
- Fixes the "shaking" behavior identified in #1906 
- Grants us full control over the behavior of the zoomable container

## Implementation Notes

The core architecture is very simple: `ZoomableImageView` uses a combination of `.scaleEffect` and `.offset` to zoom and pan the image. It overlays a `ZoomRecognizer`, which uses UIKit to recognize gestures and update the bound scale and offset.

Everything really interesting happens in `ZoomRecognizerCoordinator` and its extensions. There are four gestures that need to be accounted for:
- Pinch: this should scale the image in and out and also pan it if the pinch gesture moves
- Pan: this gesture's behavior depends on context:
  - Zoomed in: pan around the image. When the gesture ends, it should either continue to pan with momentum or halt; if it halts beyond the maximum allowable pan, the image should reset to within bounds.
  - Zoomed out: pass gesture information back up to handle viewer-level drag info (scrub and interactive dismiss)
- Double tap: zoom in/out
- Single tap: if the image is currently panning with momentum, kill the momentum; otherwise activate higher-level behavior (show/hide controls).

This requires some custom gesture coordination, which is handled in `ZoomRecognizerCoordinator+GestureRecognition`.

Zoom and pan are not particularly complicated, but do require some rather precise math to anchor correctly. `computeOffsetDeltas` is responsible for that math. Panning _while zooming_ is also a bit tricky, since the pan increment is dependent on the zoom scale at that particular tick. This is handled by `PanningPinchRecognizer`, which is a `UIPinchGestureRecognizer` with some additional logic to track the panning offset.

Most of the complexity exists around momentum. When momentum starts, a `CADisplayLink` is created to simulate the position. The momentum simulation uses a set of three custom curves defined in `ZoomCurves`:
- `SinusoidalFriction` (red): Decelerates using a cosine-based velocity curve. This is used when the view is in bounds to gradually slow it down (effectively `easeOut`).
- `PolynomialBoundBounce` (green): Decelerates to a local maximum, then returns gently to 0. Used when the momentum carries the view out of bounds to gently return it to bounds, similar to using `.easeOut` to the local maximum and `.easeInOut` on the way back to bounds.
- `PolynomialBoundReset` (black): Identical to the return component of `PolynomialBoundBounce`. Used to reset the view to bounds when momentum _begins_ with the view out of bounds.

Screenshot 2025-03-31 at 5 16 36 PM

Note that the momentum is computed independently for each axis, so if the user drags out of bounds on one axis but is within bounds on the other when momentum begins, the out-of-bound axis will return to bounds using `PolynomialBoundReset` while the other obeys the standard `SinusoidalFriction` + `PolynomialBoundBounce` behavior.


Mlem/App/Views/Shared/Images/Helpers/ZoomRecognizer/MomentumStatus.swift
```diff
@@ -0,0 +1,78 @@
...
+    private let boundResetDuration: Double = 0.3
+
+    var xt0: CFTimeInterval?
+    private var xv0: CGFloat
+    private(set) var xOob: Bool = false
+    private var xUnitCurve: any ZoomCurve
+    
+    var yt0: CFTimeInterval?
+    private var yv0: CGFloat
+    private(set) var yOob: Bool = false
```
Sjmarf, line 33: What does `yOob` mean?

Mlem/App/Views/Shared/Images/Helpers/ZoomRecognizer/ZoomRecognizerCoordinator+Logic.swift
```diff
@@ -0,0 +1,307 @@
...
+            
+            scale = newScale
+            offset = .init(
+                width: newOffset.width.bounded(lower: -maxOffsets.width, upper: maxOffsets.width),
+                height: newOffset.height.bounded(lower: -maxOffsets.height, upper: maxOffsets.height)
+            )
+        case .ended, .cancelled:
+            panType = .none
+        case .failed:
+            panType = .none
```
Sjmarf: Maybe:

```suggestion
        case .ended, .cancelled, .failed:
            panType = .none
```

(Unless there's a readability reason for it to be as it is)

Sjmarf (approved):
Nice, feels great! Fantastic job getting this working, looks like it took a lot of work 🚀

Mlem/Preview Content/PreviewLocalizable.xcstrings
```diff
@@ -1,292 +1,7 @@
 {
   "sourceLanguage" : "en",
   "strings" : {
-    "community.1.displayName" : {
```
Sjmarf: What's going on here? This shouldn't be getting removed 🤔

Mlem/App/Views/Shared/Images/Helpers/ZoomRecognizer/MomentumStatus.swift
EricBAndrews: "Y out of bounds." I've added some doc comments to all these properties to make that clearer.

Mlem/Preview Content/PreviewLocalizable.xcstrings
EricBAndrews: Git artifacts--I had to push my local to get the commit suggestion. It _should_ come back once I build and push again

EricBAndrews: Yep, 71929617c3aa393fe56e710b6bc00047ca705e13 brought this back to normal

PR #1925, opened 2025-03-30
Performance improvements when scrolling to top of a view

Sjmarf:
Eliminated a slight hitch that would occur when you scroll to the top of a `PersonView`, `CommunityView` or `InstanceView`.

PR #1933, opened 2025-04-01
Fix toolbar title not hiding/showing

Sjmarf:
Navigation bars that are only shown at the top of the view were broken in #1926. It seems like `.ignoresSafeArea()` was messing with the `isAtTop` detection somehow.

I've switched to using `containerBackground` rather than ignoring the safe area. Unfortunately this solution doesn't fix the keyboard background issue on iOS 17.

This issue is unrelated to #1925. Will merge this PR before that one.


EricBAndrews (approved):
LGTM! 🔥

PR #1934, opened 2025-04-01
Fade Markdown toolbar edges

### LINKED ISSUE #1842: Add fade to markdown toolbar to imply that it's scrollable

EricBAndrews:
## Issues

- closes #1842 

## Description

Adds a fade to the edges of the markdown toolbar when scrolling is available; also makes the toolbar background scale with system font size.

This feature makes use of `onScrollVisibilityChange` to determine whether to show the fade or not, and so does not work on iOS 17. The same logic could theoretically be implemented using `GeometryReader` and some preference keys to detect scrolling, but that seemed like overkill for such a small feature, especially considering that iOS 18 has already achieved 66% market share.


Sjmarf (approved):
Nice 💯

PR #1942, opened 2025-04-04
Inbox Design Updates

EricBAndrews:
Per Slack:
- Removed `CommentContextHeaderView`
- Moved post link to between poster and body in feed comments and inbox


Sjmarf:
It looks like there's more padding on the right of the post context than on the left. Is that intentional? It looks strange to me. Maybe it would be better without?

EricBAndrews:
Nice catch, definitely not intentional

Sjmarf (approved):
LGTM! 🚀

PR #1943, opened 2025-04-04
Better handling of errors in post and comment pages

### LINKED ISSUE #1850: Error message should be shown when comment loading fails
At the moment the `ProgressView` spins indefinitely when there's a `couldnt_find_post` error. There is no error toast because the `handleError` is silenced at `ContentLoader:112`

Sjmarf:
Closes #1850. Easiest way to repro this is to go to a removed post in the instance modlog


EricBAndrews (approved):
LGTM! 🔥

PR #1946, opened 2025-04-05
Minor localization fixes

Sjmarf:
Fixed a string not being included in `Localizable` and added a comment explaining what the `{{` `}}` are for in some localized strings


EricBAndrews (approved):
Nice! 🚀

PR #1947, opened 2025-04-05
`ComponentViews` package

Sjmarf:
Added a local package which organizes views that:
- Are not related to any one specific page of the app
- Do not depend on MlemMiddleware

In future I'd like to move `BubblePicker` here too, but it currently relies on the haptic system so that will need to be extracted out to a package first


Mlem/Packages/ComponentViews/Sources/ComponentViews/CloseButtonView.swift
```diff
@@ -9,21 +9,26 @@ import Foundation
...
+    public var body: some View {
         Button {
             if let callback {
                 callback()
             } else {
                 dismiss()
             }
         } label: {
-            Image(systemName: Icons.closeCircleFill)
+            Image(systemName: "xmark.circle.fill")
```
EricBAndrews, line 31: This might be a problem down the line--in theory we can move `Icons` into its own package so that it's available in other packages. Happy to leave OOS here, but we should solve that before the number of hardcoded strings gets too high.

EricBAndrews (approved):
LGTM! 🔥

Mlem/Packages/ComponentViews/Sources/ComponentViews/CloseButtonView.swift
Sjmarf: Agreed, in future `Icons` should be in a package too. Then we could use `Icons` here 👍

PR #1949, opened 2025-04-05
Contextual infostack coloring

### LINKED ISSUE #1929: Upvote/downvote icon in info stack should be colored if upvote/downvote buttons not present
https://lemmy.ml/comment/17274145

EricBAndrews:
## Issues

- closes #1929 

## Description

Reworks the info stack coloring logic to color readouts displaying information that is not captured in a widget. This involves a couple minor architecture changes:
- `showColor` is now a parameter to the `.readout` methods provided on `Post`/`Comment`/`Reply`, rather than a blanket parameter to the entire `InfoStack`
  - `showColor` is no longer a parameter on `InfoStack`
  - Convenience initializers for `InfoStackView` now take `coloredReadouts` instead; the default initializer assumes that `Readout` is correctly configured for the desired coloring behavior.
- `ActionTypeProviding` and `CounterTypeProviding` now expose `.associatedReadouts`, which returns the set of readouts that are redundant with the particular action or counter.
  - `InteractionConfigurationItem` also exposes this as a passthrough function to the underlying type
  - Because the score readout's redundancy can depend on upvote/downvote state (e.g., if the user has an upvote widget but no downvote widget, score should be colored when the item is downvoted but not upvoted), `.associatedReadouts` takes in the relevant `Interactable1Providing` as a parameter


Sjmarf (approved):
LGTM, nice solution 💯

PR #1951, opened 2025-04-06
Improve "jump back to position" button logic

Sjmarf:
In addition to showing the button when you return to a thread you've visited previously, the jump back to position button will now also be shown if you scroll back to the top of the post you're visiting for the first time.

If you're reading a comment thread and want to quickly check something in the post, you can scroll to the top (e.g. by tapping the status bar). Then, the jump button will appear to take you back to where you were in the comment thread.


EricBAndrews (approved):
LGTM! 🔥 

Super convenient feature

PR #1959, opened 2025-04-07
Fix inbox swipe action configuration not persisting

### LINKED ISSUE #1957: Inbox Swipe Actions Reset After App Relaunch
I have tested Posts, Comments, and Inbox, and so far Inbox is the only one that is not saving the customized swipe actions after closing and reopening the app. 

I am attempting to switch the order of the swipe actions, so I click the delete icon for Select Text and then add a t back so it comes after Mark Read. I have also tried to remove both items and then add them back.

Thank you!

Sjmarf:
Fixes #1957


EricBAndrews (approved):
Nice 👍

Issue #1323, opened 2024-09-13
Interaction Bar Widget Persistence Improvements


EricBAndrews:
Interaction bar widgets should be brought into line with the new settings. This involves:
- Move layout widgets into Settings
- Add logic to decode and translate v1 widget layout

Depends on removing settings from AppStorage.

Sjmarf:
> Add logic to decode and translate v1 widget layout

Ideally we'd do this before 2.0

EricBAndrews:
Moving this out of 2.0 since the key 2.0 change (decoding v1 layouts) is implemented and moving this into settings introduces a lot of technical challenge with no user-facing benefit and the tech debt here doesn't stand to grow if left until post-2.0

Sjmarf:
Closed?

Sjmarf:
Ah, not yet 👍

PR #1961, opened 2025-04-08
Migrate Settings off of AppStorage

### LINKED ISSUE #1820: Non-AppStorage Settings Persistence
### Requirements

### Description

This was discussed in Slack a while back but an issue was never opened. 

This would also enable per-account settings, which has been requested with some regularity.

EricBAndrews:
## Issues

- closes #1820 
- progress towards #1323 

## Description

This PR migrates settings to a file-based storage solution instead of AppStorage. This has several benefits:
- Allows storing more complex settings types (filters, #1323)
- Enables explicit settings versioning using the system settings path
- Reduces redundant logic between save/restore and normal settings persistence

As a proof of concept for more complex settings types, filters have been moved entirely into settings.

## Implementation Notes

The new settings stack is made of four components:
- Persistent file: stores the settings state on disk. Read from at app launch and written to every time a setting is modified.
- `SettingsValues`: this is the engine of the settings system. It stores every setting as a member variable, and conforms to:
  - `@Observable`: publishes settings updates. Unfortunately, this also necessitates explicitly specifying coding keys.
  - `Codable`: enables it to be read from and written to disk
- `Settings`: provides a static interface for interacting with `SettingsValues`. To ensure that `Settings.values`, the source of truth for settings in the app, is never accessed, both `values` and the static `main` singleton are private. Interaction happens through the static functions of `Settings`, which in turn act on the private static `main`. This guarantees that settings changes are always saved.
- `Setting`: property wrapper for access to settings in the rest of the app. Works identically as before, just with slightly different logic to plug into `Settings`.

Most of the line count is just renames. The important changes are in:
- `Settings`
- `SettingsValues`
- `SettingPropertyWrapper`
- `FiltersTracker`


Sjmarf:
Nice work 😄 Am about to go to bed so will give this a proper look over tomorrow

> @Observable: publishes settings updates. Unfortunately, this also necessitates explicitly specifying coding keys.

We might be able to avoid defining this by setting the `keyEncodingStrategy` of the `JSONEncoder` to `.custom`, passing a function that strips leading underscores 🤔 Maybe more trouble than it's worth though if we also need to do camel case -> snake case conversion

That would also allow us to replace the underscores with dots, while we're at it. E.g. `"accounts.sort"` rather than `accounts_sort`

Mlem/App/Configuration/User Settings/Settings.swift
```diff
@@ -1,212 +1,117 @@
...
     
-    var codable: CodableSettings { .init(from: self, filteredKeywords: FiltersTracker.main.rawKeywords) }
+    private func _save(to systemSetting: SystemSetting) async {
+        do {
+            try await persistenceRepository.saveSystemSettings(values, setting: systemSetting)
+            ToastModel.main.add(.success("Saved Settings"))
+        } catch {
+            handleError(error)
+        }
+    }
```
Sjmarf, line 72: Dead code?

```suggestion
```

Sjmarf (approved):
Awesome work! This is so much easier to work with; `@AppStorage` only supporting certain types was super annoying. 🚀

Mlem/App/Configuration/User Settings/Settings.swift
EricBAndrews: That one is used--it's called at L39, and backs the save/restore functionality

PR #1968, opened 2025-04-11
Better Interaction Bar Persistence

### LINKED ISSUE #1323: Interaction Bar Widget Persistence Improvements

EricBAndrews:
## Issues

- closes #1323 

## Description

Move interaction bar persistence into Settings; also removes v1 interaction bar decoding.

**Testing Note:** migration from the previous persistence solution is bundled with the other 2.0 -> 2.1 settings migrations, so it will only pick up the 2.0 configuration if the `.v2_system` file is not present. Migration testing should therefore be done from the 2.0 release tag rather than the latest state of dev.


Sjmarf (approved):
LGTM! Migration works perfectly 👍

PR #1973, opened 2025-04-11
Fix MainActor warnings in MediaLoader

EricBAndrews:
Makes `MediaLoader` state updates `MainActor`. Silences a big warning stack that started appearing with #1968, though I'm not sure what that PR did to cause the warnings.


Sjmarf (approved):
Nice 👍

PR #1975, opened 2025-04-11
Fix default post size

### LINKED ISSUE #1971: Default post size is `.compact`; it should be `.large`

EricBAndrews:
Sets default post size to `.large`

This has to happen in `LegacySettings` for now because that gets used for initialization any time the `.v2_system` settings aren't found. 

Closes #1971


Sjmarf (approved):
LGTM 👍

PR #1985, opened 2025-04-13
Fix single-frame "animated" images

### LINKED ISSUE #1911: Assertion failure in `AnimatedImageView`
I have [this comment](https://sh.itjust.works/comment/17357597) in my inbox. I consistently hit the following assertionFailure at `AnimatedImageView:46` when the inbox comment view loads.

```swift
guard let player = imageView.player else {
    assertionFailure("ImageView had nil player")
    return
}
```

URL of the avatar that is causing it: https://feddit.org/pictrs/image/bf9867ec-fdc9-4684-bcab-d4dea471648f.gif?thumbnail=32

EricBAndrews:
## Issues

- closes #1911 

## Description

Adds logic to handle single-frame animated images. `AnimatedImageView` now updates `MediaControlState.animationAvailable` to `false` after the image is decoded if the image only has 1 frame.

### DIFF
--- a/Mlem/App/Views/Shared/Images/Core/Animated/AnimatedImageView.swift
+++ b/Mlem/App/Views/Shared/Images/Core/Animated/AnimatedImageView.swift
@@ -25,6 +25,12 @@ struct AnimatedImageView: UIViewRepresentable {
             return imageView
         }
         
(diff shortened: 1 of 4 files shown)


Sjmarf (approved):
Nice! 💯

PR #1987, opened 2025-04-16
Icon Package

### LINKED ISSUE #1952: Move icons to package

Sjmarf:
> [!note]
> Drafting until 2.1 is out

Moved the icons system to an `Icons` package, and rewrote the system to be more convenient to use. I've moved all of the code over to using the new package, **except for** the action system. This is because I want to do some major refactoring to the action system sometime soon, and don't think it's worth moving those over now. The action system continues to use the old `Icons` object, which will be removed once I've redone the action system

# The New System

Previously, we stored icon names statically as strings in the `Icons` enum. In this PR, icons are now represented using an `Icon` struct. Instances of `Icon` are defined statically on the `Icon` type. E.g.

```swift
Icons.lemmy.upvote
```

 I've overloaded the SwiftUI initializers that accept `systemImage: String` to alternatively accept `icon: Icon`:
 
 ```swift
Image(icon: .lemmy.upvote)
 ```
 
 Icons are organised into namespaces. Lemmy-related concepts are under the `.lemmy` namespace, as seen above. Icons that are more general are under a `.general` namespace l (e.g. `.general.success`), and settings icons are under a `.settings` namespace.
 
 In future, namespaces could be segmented into packages. For example, down the line we could move the Fediseer-related views into a `FediseerViews` package. When we do that, the Fediseer-related icons could be contained within an internal `.fediseer` icon namespace of that package.

## Symbol Variants

Previously, we defined `-fill` and `-circle` variants as separate icons. In the new system, this is no longer the case. Instead, icon variants should be applied using SwiftUI's `.symbolVariant` modifier in most cases.

In cases where an icon's variant is used to indicate an on/off state, you can use the `representingState(active: )` method that I've added to `Icon`:

```swift
Image(icon: .lemmy.saved.representingState(active: post.saved))
```

For most icons, using `representingState` just fills the icon when `active` is true. However, some icons do not have a fill state. Those icons are instead wrapped in a shape, so that fill status can be used to convey on/off state. A good example of this is the `.lemmy.upvote` icon:


Mlem/App/Views/Root/Onboarding/SignUpView+Views.swift
```diff
@@ -69,17 +69,21 @@ extension SignUpView {
...
-                    case .checking:
-                        ProgressView()
-                            .tint(.themedSecondary)
-                    case .valid:
-                        Image(systemName: Icons.successCircleFill)
-                            .foregroundStyle(.themedPositive)
-                    case .taken, .tooShort, .invalidCharacters:
-                        Image(systemName: Icons.failureCircleFill)
-                            .foregroundStyle(.themedNegative)
+                Group {
```
EricBAndrews, line 72: What's the motivation for putting these in `Group`?

Mlem/Localizable.xcstrings
```diff
@@ -2978,30 +2981,51 @@
...
     },
     "end.of.feed.icon.1" : {
+      "extractionState" : "extracted_with_value",
       "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "new",
+            "value" : "figure.climbing"
+          }
+        },
```
EricBAndrews, line 2989: Not sure how these ended up in here 🤔

Mlem/Packages/Icons/Sources/Icons/Icon+Fediseer.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  File.swift
```
EricBAndrews:
```suggestion
//  Icon+Fediseer.swift
```

Mlem/Packages/Icons/Sources/Icons/Icon+Settings.swift
```diff
@@ -0,0 +1,81 @@
...
+        public let blurNsfw: Icon = .init("eye.trianglebadge.exclamationmark")
+        public let upvoteOnSave: Icon = .init("arrow.up.heart")
+        public let readIndicatorSetting: Icon = .init("book")
+        public let readIndicatorBarSetting: Icon = .init("rectangle.leftthird.inset.filled")
+        public let profileTabSettings: Icon = .init("person.text.rectangle")
+        public let nicknameField: Icon = .init("rectangle.and.pencil.and.ellipsis")
+        public let label: Icon = .init("tag")
+        public let unreadBadge: Icon = .init("envelope.badge")
+        public let showAvatar: Icon = .init("person.fill.questionmark")
+        public let widgetWizard: Icon = .init("wand.and.stars")
```
EricBAndrews: Do we still use this one anywhere?

Mlem/Packages/Icons/Sources/Icons/Icon+Markdown.swift
```diff
@@ -0,0 +1,29 @@
+//
+//  File.swift
```
EricBAndrews:
```suggestion
//  Icon+Markdown.swift
```

Mlem/Packages/Icons/Sources/Icons/Icon+Settings.swift
```diff
@@ -0,0 +1,81 @@
+//
+//  File.swift
```
EricBAndrews:
```suggestion
//  Icon+Settings.swift
```

Mlem/Packages/Icons/Sources/Icons/Icon+Uptime.swift
```diff
@@ -0,0 +1,18 @@
+//
+//  File.swift
```
EricBAndrews:
```suggestion
//  Icon+Uptime.swift
```

Mlem/Packages/Icons/Sources/Icons/Icon.swift
```diff
@@ -0,0 +1,100 @@
+//
+//  Icons.swift
```
EricBAndrews:
```suggestion
//  Icon.swift
```

EricBAndrews (commented):
Fantastic work! This is much nicer than the old system. I've left a couple comments, I'll give it a proper review once 2.1 ships

Mlem/Packages/Icons/Sources/Icons/Icon+Settings.swift
Sjmarf: Nope, removed 👍

Mlem/App/Views/Root/Onboarding/SignUpView+Views.swift
Sjmarf: I had meant to apply the `.symbolVariant` to the `Group`, but it seems I forgot to. Fixed, thanks 👍

Mlem/Localizable.xcstrings
Sjmarf: Fixed 👍

EricBAndrews (approved):
LGTM! 🔥

PR #1990, opened 2025-04-19
Scrub Bugfix

EricBAndrews:
Fixes some minor scrubbing issues:
- Playback bar hitbox coordinate space was offset from intended location; fixed by using `.global` coordinate space, which matches what the `ZoomRecognizer` uses
- Dragging exactly on the playback bar didn't do anything; fixed by disabling hit detection on that view
- Scrub speed could still be changed when dragging the playback head; fixed by correctly tracking start location in `ZoomRecognizerCoordinator`


Sjmarf (approved):
Nice 👍

PR #1994, opened 2025-04-20
Don't clear posts when pull-to-refresh triggered; fix inbox refresh delay

### LINKED ISSUE #1984: Pull-to-refresh in the inbox tab is shown for a few seconds after the inbox refresh finishes
My guess is that pull-to-refresh is reloading both the inbox and mod mail, and mod mail takes longer than the inbox which gives the appearance of a delay. If so, this could be fixed by only reloading the page the user is looking at

### LINKED ISSUE #1993: Do not clear feed when refreshing
Apple doesn't clear the feed when you pull-to-refresh, so I don't think we should either. At the moment we're showing two loading symbols (the pull to refresh symbol, and the feed loader ProgressView) which is redundant

Sjmarf:
This is for 2.2

Closes #1984, closes #1993


Mlem/App/Views/Root/Tabs/Inbox/InboxView.swift
```diff
@@ -182,8 +182,12 @@ struct InboxView: View {
             if selectedFeed == .modMail, !appState.isModOrAdmin {
                 selectedFeed = .inbox
             }
-            try await inboxFeedLoader.refresh(clearBeforeRefresh: true)
-            try await modMailFeedLoader.refresh(clearBeforeRefresh: true)
+            switch selectedFeed {
+            case .inbox:
+                try await inboxFeedLoader.refresh(clearBeforeRefresh: false)
+            case .modMail:
+                try await modMailFeedLoader.refresh(clearBeforeRefresh: false)
```
EricBAndrews: I think we should still refresh the other loader, just not hang the `.refreshable`--wrapping the other refresh in a `Task` ought to do the trick.

EricBAndrews (changes requested):

Mlem/App/Views/Root/Tabs/Inbox/InboxView.swift
Sjmarf: Good point, done 👍

EricBAndrews (approved):
LGTM! 🔥

Issue #1996, opened 2025-04-21
Migrate media to Swift 6


EricBAndrews:
### Requirements

- [x] There are no existing issues for this feature.
- [x] This is a request for a **single** feature (create multiple issues for multiple feature requests).

### Description

Media views should be refactored to compile against Swift 6 strict concurrency

PR #1997, opened 2025-04-21
Package Media

### LINKED ISSUE #1941: Move media views into package

EricBAndrews:
## Issues

- closes #1941 

## Description

Moves media views into a package. The only real change is moving overlays out of `MediaControlState`.

There are a lot of Swift 6 concurrency issues, so the package uses Swift 5; the scope of this PR is just packaging without additional code changes, so I've opened #1996 to track this.

In several places I have commented rather than removed calls to `handleError`; this is another component we may be able to package along with the toast system.


Sjmarf (approved):
LGTM 🔥

PR #1999, opened 2025-04-22
Minor zoom performance

EricBAndrews:
Minor tweaks to improve the zoom performance--at 120fps, the first few frames are a bit choppy. This PR implements two tweaks:
- Turn down the `CADisplayLink` max fps. 100 was a bit optimistic, 90 feels much more reliable
- Use static singletons for zoom curves to eliminate redundant initialization


Sjmarf (approved):
Nice! 💯

PR #2005, opened 2025-04-25
Improve Mark Read on Scroll Reliability

### LINKED ISSUE #1998: Mark read on scroll (possibly) state-fakes posts as read, but doesn't actually mark them as such?
Reported here:

https://lemmy.ml/comment/18097495

Haven't done any testing yet.

EricBAndrews:
## Issues

- closes #1998 

## Description

This behavior was caused by the `flushPostReadQueue` not being called in certain cases. Previously, that method was called:
- When moving the app into the background
- When switching accounts
- When marking a post read

However, if the user scrolled a `PostGridView`, left that page without interacting, and returned, none of the flush triggers were hit and so the posts were reset to unread.

This PR fixes that by adding a call to `flushPostReadQueue` whenever a `PostGridView` disappears.

### DIFF
--- a/Mlem/App/Views/Shared/PostGridView.swift
+++ b/Mlem/App/Views/Shared/PostGridView.swift
@@ -56,10 +56,10 @@ struct PostGridView: View {
                 }
             }
             .onDisappear {
-                if let firstAccount = AppState.main.firstAccount as? UserAccount {
+                if let api = postFeedLoader.items.first?.api {
                     Task {
                         do {
-                            try await firstAccount.api.flushPostReadQueue()
+                            try await api.flushPostReadQueue()
                         } catch {
                             handleError(error)
                         }
--- a/Mlem/App/Views/Shared/PostGridView.swift
+++ b/Mlem/App/Views/Shared/PostGridView.swift
@@ -55,6 +55,17 @@ struct PostGridView: View {
                     handleError(error)
                 }
             }
+            .onDisappear {
+                if let firstAccount = AppState.main.firstAccount as? UserAccount {
+                    Task {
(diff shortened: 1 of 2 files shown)


Mlem/App/Views/Shared/PostGridView.swift
```diff
@@ -55,6 +55,17 @@ struct PostGridView: View {
                     handleError(error)
                 }
             }
+            .onDisappear {
```
Sjmarf, line 58: Might we also need this for `PersonContentGridView`?

Mlem/App/Views/Shared/PostGridView.swift
```diff
@@ -55,6 +55,17 @@ struct PostGridView: View {
...
+                if let firstAccount = AppState.main.firstAccount as? UserAccount {
```
Sjmarf: It might be better to use the `api` of the first post in the feed loader instead, in-case the user switches accounts and has an out-of-date feed

Mlem/App/Views/Shared/PostGridView.swift
EricBAndrews: Mark read on scroll is only enabled on `PostGridView`. It's obviously pointless on the saved feed, and IIRC we decided to leave it off of profile because that view isn't consumed the same way a feed is

Mlem/App/Views/Shared/PostGridView.swift
EricBAndrews: Nice catch, updated 👍

Sjmarf (approved):
LGTM! 💯

PR #2008, opened 2025-04-25
Fix crash when opening reply editor

### LINKED ISSUE #2007: Crash when opening reply editor
### Requirements

### Description

_No response_

### Mlem Version

_No response_

EricBAndrews:
## Issues

- closes #2007 

## Description

Fixes a crash when opening the reply editor. This was due to `person.crop` failing to resolve to a SFSymbol, which in turn crashed the force unwrap in `SimpleAvatarView:31`. This PR fixes the issue by updating `Icon.personAvatar` to `person.crop.circle` and adding a guard to prevent the force unwrap crashing the app, plus making it easier to debug. Also fixes `.select` not resolving.

### DIFF
--- a/Mlem/Packages/Icons/Sources/Icons/Icon+Lemmy.swift
+++ b/Mlem/Packages/Icons/Sources/Icons/Icon+Lemmy.swift
@@ -187,7 +187,7 @@ public extension Icon {
         public let subscriptionList: Icon = .init("list.bullet")
         
         @inlinable public var communityAvatar: Icon { community }
-        public let instanceAvatar: Icon = .init("building.2.crop")
+        public let instanceAvatar: Icon = .init("building.2.crop.circle")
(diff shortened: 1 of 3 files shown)


Mlem/Packages/Icons/Sources/Icons/Icon+Lemmy.swift
```diff
@@ -188,7 +188,7 @@ public extension Icon {
         
         @inlinable public var communityAvatar: Icon { community }
         public let instanceAvatar: Icon = .init("building.2.crop")
```
Sjmarf: We may want to apply the same logic here?

```suggestion
        public let instanceAvatar: Icon = .init("building.2.crop.circle")
```

Sjmarf (approved):
Nice 👍

PR #2011, opened 2025-04-28
Interaction Bar Hitboxes

EricBAndrews:
## Issues

- community feedback

## Description

This PR increases the interaction bar widget hitboxes to the intended full 44x44px. This necessitated some light reworking of how the interaction bar is attached to its parent view: the interaction bar now has no padding applied by the parent view and no spacing between itself and the top of the parent view. The overall appearance of the bar is unchanged; the only difference is that the padding is now an implicit result of the bar constants rather than explicitly defined on the parent view. This approach eliminates the need for negative padding tricks and thus minimizes the risk of hitbox overlap down the line.


Sjmarf (changes requested):
Nice! This is much cleaner than using negative padding. The interaction bar has extra padding in the inbox; probably just need to remove some spacing from `ReplyView`

Mlem/App/Views/Shared/CommentView.swift
```diff
@@ -64,39 +64,41 @@ struct CommentView<EmbeddedContent: View>: View {
...
-                                configuration: interactionBarConfiguration,
-                                commentTreeTracker: commentTreeTracker,
-                                communityContext: communityContext,
-                                reportContext: reportContext
-                            )
-                            .padding(.horizontal, 2)
-                            .padding(.bottom, 3)
-                            .padding(.top, 1)
-                        }
+                    if !compact {
```
Sjmarf: Need a `!collapsed` somewhere here; the interaction bar doesn't currently get hidden when collapsing a comment

Sjmarf (changes requested):
There's some horizontal padding issues if the user puts one of the counters in the rightmost slot of the interaction bar:

(Before)


(After)

Sjmarf:
Also the "comment was deleted" text is incorrectly aligned in a profile feed

Sjmarf (approved):
LGTM 🚀

Issue #1137, opened 2024-07-09
Local Community List in `InstanceView`


Sjmarf:
https://lemmy.ml/post/17766734

Sjmarf:
Such a list is now available from search by adding a filter for a specific instance. Though I think it would still be worthwhile to include it in `InstanceView` anyway.

Sjmarf:
Ideally we'd fix the issue (that we currently have in the Search implementation of this) where the subscribed status aren't shown. I really don't like having to display that warning; it exposes the technical underside of the app. I've got some ideas for how this could be resolved. This could potentially be a good testing ground to implement some of the architecture we'll eventually need for multi-accounts.

Sjmarf:
Either we'd add a "Communities" tab to `InstanceView` or make the "Communities" section of the `Details` tab tappable, and push a new page to the stack when it gets tapped.

The benefit of adding a new tab is that it's consistent with the other pages we've got (e.g. the "Communities" tab in `PersonView`). 

However, the tab picker is already getting quite full; this might be one too many. One way to resolve that might be to move "Uptime" and "Trust & Safety" into "Details". "Trust & Safety" could alternatively go into "Administration", as that tab contains details on how well the instance is run.

Could even put the instance config list into a new page with a link from the "Details" tab; most of those config options aren't particularly interesting anyway.

PR #2024, opened 2025-05-05
Show subscription and blocked status in instance-filtered search

EricBAndrews:
## Issues

- progress towards #1137 

## Description

Adds subscription and blocked status to search results when filtered by instance.

More generally, this is a proof-of-concept for synthetic state management, which will be integral to multi-accounts.

### General Changes
- Added `SyntheticStateManager` (see below)
- `StateManager` now has both `wrappedValue` and `displayedValue`. This allows `StateManager` and `SyntheticStateManager` to expose a consistent API. All references to `.wrappedValue` have been updated to `displayedValue` (this is most of the line count)

### `SyntheticStateManager`
`SyntheticStateManager` is a subclass of `StateManager` that wraps any `MergeableValue`.

Types can conform to the `MergeableValue` protocol to define how they should merge with each other to produce a single synthetic value. Merging can be done either disjunctively or conjunctively depending on the use case (e.g., votes should show as true if any value is true, but saved should only show as true if all values are true).

`SyntheticStateManager` defines a map of sibling `SyntheticStateManagers`. Its `displayedValue` is computed by reducing the sibling values using `MergeableValue.merge` according to the merge type of the state manager.

### Architecture

To keep this code flexible, the search simply takes in a `hostApi` parameter that determines which ApiClient to use for enriching the synthetic state with account-specific information. Right now that's passed in from `SearchView`; any multi-account `ApiClient` structure should have this information available, so updating the code to use that architecture should require only updating where and how `hostApi` is passed in.


Sjmarf (approved):
Nice work 😄

PR #2030, opened 2025-05-14
Unread count in account switcher

### LINKED ISSUE #1599: Fetch unread counts for inactive accounts and display badges in account switcher
Title. Do we also want to display a combined notification count for other accounts in the tab bar somehow? Maybe on the profile tab, and then inside of the profile tab we put a "switch account" button that also has the badge? In v1 we had a "switch account" button in the top-left toolbar position of the profile tab. Maybe it's worth bringing that back?

Sjmarf:
Closes #1599

Also added a "show response time" setting under Settings -> Accounts that shows the response times of each instance in the account switcher.

### DIFF
--- a/Mlem/App/Views/Shared/ListRow/AccountListRowBody.swift
+++ b/Mlem/App/Views/Shared/ListRow/AccountListRowBody.swift
@@ -12,11 +12,13 @@ struct AccountListRowBody: View {
     @Environment(AppState.self) private var appState
     
     enum Complication: CaseIterable {
-        case instance, lastUsed, isActive
+        case instance, lastUsed, responseTime, isActive, unreadCount
     }
     
     let account: any Account
-    var complications: Set<Complication> = .withTime
+    var unreadCount: Int?
+    var responseTime: TimeInterval?
+    var complications: Set<Complication> = .instanceAndTime
     
     var body: some View {
         HStack(alignment: .center, spacing: 10) {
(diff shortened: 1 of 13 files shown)


Mlem/App/Views/Shared/ListRow/AccountListRowBody.swift
```diff
@@ -73,12 +82,43 @@ struct AccountListRowBody: View {
...
     }
 }
 
+private struct AccountListRowBodyReadoutView: View {
+    let isActive: Bool
+    let unreadCount: Int?
+    let complications: Set<AccountListRowBody.Complication>
+    
+    var body: some View {
+        ZStack {
```
EricBAndrews: Why `ZStack`?

EricBAndrews (changes requested):
Nice work!

One minor code note--also I'm seeing this warning once(ish) per account when opening the switcher:

`ApiClient.myPerson or ApiClient.myInstance is nil at UnreadCount refresh - this may lead to unneeded API calls`

Mlem/App/Views/Shared/ListRow/AccountListRowBody.swift
Sjmarf: Good point, this was leftover from me trying to figure out why the animations weren't working properly. Have removed it 👍

Sjmarf:
> One minor code note--also I'm seeing this warning once(ish) per account when opening the switcher:

Thanks, have fixed 👍

EricBAndrews (approved):
LGTM! 💯

PR #2031, opened 2025-05-15
Fix success toast

### LINKED ISSUE #2020: Success toast is missing circle background
Introduced in #1987 

Also need to fix the profile tab bar icon

EricBAndrews:
## Issues

- closes #2020 

## Description

Apologies for issue poaching

### DIFF
--- a/Mlem/App/Views/Shared/Toast/ToastView.swift
+++ b/Mlem/App/Views/Shared/Toast/ToastView.swift
@@ -21,7 +21,7 @@ struct ToastView: View {
     static let dimmedSymbols: Set<Icon> = [.lemmy.block]
     
     // These symbols need `.symbolVariant(.circle.fill)` applied to render properly
-    static let circledSymbols: Set<Icon> = [.general.success, .general.error, .general.failure]
+    static let circledSymbols: Set<Icon> = [.general.success, .general.error, .general.failure, .general.undo]
     
     var body: some View {
         HStack {
--- a/Mlem/App/Views/Shared/Toast/ToastView.swift
+++ b/Mlem/App/Views/Shared/Toast/ToastView.swift
@@ -20,6 +20,9 @@ struct ToastView: View {
     // These symbols only have a single hierarchical layer, so we render it as `.secondary`
     static let dimmedSymbols: Set<Icon> = [.lemmy.block]
     
+    // These symbols need `.symbolVariant(.circle.fill)` applied to render properly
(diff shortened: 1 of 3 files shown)


Mlem/App/Views/Shared/Toast/ToastView.swift
```diff
@@ -92,6 +92,7 @@ struct ToastView: View {
         HStack(spacing: Constants.main.doubleSpacing) {
             if let icon {
                 image(icon, color: imageColor)
+                    .symbolVariant(.circle.fill)
```
Sjmarf: Regrettably we can't do this because there are cases where we *dont* want the circle, e.g. when blocking someone. 

One solution would be to add a `symbolVariants` value to `ToastType` that lets the caller specify whether they want a circle or not. Or maybe a `withCircle` Boolean, to prevent use of the other variants (callers shouldn't really be able to set `.slash`, for example)

Sjmarf (changes requested):

Mlem/App/Views/Shared/Toast/ToastView.swift
EricBAndrews: Good point--I've defined a set of `circledSymbols` containing the symbols that need `.circle.fill`, and conditionally applied the `.symbolVariant`. This should make the rendering work correctly without ever exposing that nuance to the caller.

Mlem/App/Views/Shared/Toast/ToastView.swift
```diff
@@ -20,6 +20,9 @@ struct ToastView: View {
     // These symbols only have a single hierarchical layer, so we render it as `.secondary`
     static let dimmedSymbols: Set<Icon> = [.lemmy.block]
     
+    // These symbols need `.symbolVariant(.circle.fill)` applied to render properly
+    static let circledSymbols: Set<Icon> = [.general.success, .general.error, .general.failure]
```
Sjmarf: I think `.general.undo` needs to be in here too. It's the fallback icon (line 58) though I don't believe we're using it anywhere yet

```suggestion
    static let circledSymbols: Set<Icon> = [.general.success, .general.error, .general.failure, .general.undo]
```

Sjmarf (approved):

Mlem/App/Views/Shared/Toast/ToastView.swift
Sjmarf: Nice! That's a really clean solution 👍

PR #2001, opened 2025-04-23
"Always Show Account Age" Setting

Sjmarf:
General -> Show Account Age

Closes #1818. Also fixes #1988

Issue #1988, opened 2025-04-16
Comment timestamp can sometimes be shown as older than author's account


Sjmarf:
I presume this is just a difference in the way that we're calculating the relative timestamp.

<img src="https://github.com/user-attachments/assets/b6468816-c734-4a5e-944c-c8ecc20ef740" width=50%>

PR #2033, opened 2025-05-15
[#2032] Fix localized elapsed time on profile

### LINKED ISSUE #2032: Elapsed time on profile since creation is not the expected time
### Requirements

# Description

Given a Lemmy account created at 04/11/2023
When I display my profile with Mlem 2.1 (322)
Then I see in my profile it was created less than one year ago

Given a Lemmy account created at 04/11/2023
When I display my profile with Mlem 2.1 (322)
Then I see in my profile it was created "-1 a" which can be hardly understandable

# Screenshots

Lemmy web profile

Mlem 2.1 profile with bug

### Mlem Version

Mlem 2.1 (322)

pylapp:
## Issues

- closes #2032

## Description

The elapsed time since the creation of the Lemmy account displayed on the profile now has the expected value.
The value is computed and does not display for example "-1" if the account is more than 2 years.
The string is also displayed in full format so as to help understandability and *Voice Over* vocalization.

## Implementation Notes

- Use the dates in the expected format so as to allow the `RelativeDateTimeFormatter` to process them (i.e. otherwise fail if hours are in dates)
- Change units style to have a more understandable string representation of the elapsed time
- Add unit tests to ensure the date calculations remain the expected ones
- Make some properties and function public to allow unit tests

### DIFF
--- a/Mlem/App/Utility/Extensions/Date+Extensions.swift
+++ b/Mlem/App/Utility/Extensions/Date+Extensions.swift
@@ -8,16 +8,33 @@
 import SwiftUI
 
 extension Date {
-    // Returns strings like "3 seconds ago" and "10 days ago"
-    func getRelativeTime(date: Date = .now, unitsStyle: RelativeDateTimeFormatter.UnitsStyle = .full) -> String {
+    /// Forges a localized `String` with inside the computed elapsed time between `self` and another `Date`.
+    /// Uses a `RelativeDateTimeFormatter` and a given units style.
+    ///
+    /// For example, if the `self` is date 04/11/2023 and `date` is 15/05/2025, will return "2 years ago" (localized).
+    ///
+    /// - Parameters:
+    ///    - date: The date to comapre with `self`, by default `Date.now`
+    ///    - unitsStyle: The style of the string to forge, by default `RelativeDateTimeFormatter.UnitsStyle.full`
+    /// - Returns String: The localized string based.
+    public func getRelativeTime(date: Date = .now, unitsStyle: RelativeDateTimeFormatter.UnitsStyle = .full) -> String {
         let formatter = RelativeDateTimeFormatter()
         formatter.unitsStyle = unitsStyle
-
-        return formatter.localizedString(for: self, relativeTo: date)
+        // Need to get rid of hours and keep only a simple pure date to have the relative date time formatter working
+        return formatter.localizedString(for: shortered, relativeTo: date)
     }
-    
-    // Returns strings like "5/10/2023"
-    var dateString: String {
+
+    /// Returns `self` but with only the day, the month and the year, i.e. the current date without hours.
+    /// If the conversion fails, returns `self` as is.
+    public var shortered: Date {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "dd/mm/yyyy"
+        return dateFormatter.date(from: dateString) ?? self
+    }
+
+    /// Returns the current `Date` as a shorter version in `String`.
+    /// For example if the date in the 5th of October 2023, returns "5/10/2023"
+    public var dateString: String {
         let dateFormatter = DateFormatter()
         dateFormatter.dateStyle = .short
         dateFormatter.timeStyle = .none
--- a/Mlem/App/Utility/Extensions/Date+Extensions.swift
+++ b/Mlem/App/Utility/Extensions/Date+Extensions.swift
@@ -20,23 +20,12 @@ extension Date {
     public func getRelativeTime(date: Date = .now, unitsStyle: RelativeDateTimeFormatter.UnitsStyle = .full) -> String {
         let formatter = RelativeDateTimeFormatter()
         formatter.unitsStyle = unitsStyle
-        // Need to get rid of hours and keep only a simple pure date to have the relative date time formatter working
-        // Using date components to truncate useless data is not relevant as relative date time formatter
-        // cannot process sunch down rounded values
-        return formatter.localizedString(for: shortered, relativeTo: date)
-    }
-
-    /// Returns `self` but with only the day, the month and the year, i.e. the current date without hours.
-    /// If the conversion fails, returns `self` as is.
-    public var shortered: Date {
-        let dateFormatter = DateFormatter()
-        dateFormatter.locale = Locale.current
-//        dateFormatter.dateFormat = "dd/MM/yyyy"
-        return dateFormatter.date(from: dateString) ?? self
+        return formatter.localizedString(for: self, relativeTo: date)
     }
 
     /// Returns the current `Date` as a shorter version in `String`.
-    /// For example if the date in the 5th of October 2023, returns "5/10/2023"
+    /// For example if the date in the 5th of October 2023, returns "5/10/2023".
+    /// Uses the current locale to let the `DateFormatter` apply the suitable date format depending to the user needs.
     public var dateString: String {
         let dateFormatter = DateFormatter()
         dateFormatter.dateStyle = .short
--- a/Mlem/App/Utility/Extensions/Date+Extensions.swift
+++ b/Mlem/App/Utility/Extensions/Date+Extensions.swift
@@ -11,16 +11,18 @@ extension Date {
     /// Forges a localized `String` with inside the computed elapsed time between `self` and another `Date`.
     /// Uses a `RelativeDateTimeFormatter` and a given units style.
     ///
-    /// For example, if the `self` is date 04/11/2023 and `date` is 15/05/2025, will return "2 years ago" (localized).
+    /// For example, if the `self` is date 04/11/2023 and `date` is 15/05/2025, will return "1 year ago" (localized).
     ///
     /// - Parameters:
-    ///    - date: The date to comapre with `self`, by default `Date.now`
+    ///    - date: The date to compare with `self`, by default `Date.now`
     ///    - unitsStyle: The style of the string to forge, by default `RelativeDateTimeFormatter.UnitsStyle.full`
     /// - Returns String: The localized string based.
     public func getRelativeTime(date: Date = .now, unitsStyle: RelativeDateTimeFormatter.UnitsStyle = .full) -> String {
         let formatter = RelativeDateTimeFormatter()
         formatter.unitsStyle = unitsStyle
         // Need to get rid of hours and keep only a simple pure date to have the relative date time formatter working
+        // Using date components to truncate useless data is not relevant as relative date time formatter
+        // cannot process sunch down rounded values
         return formatter.localizedString(for: shortered, relativeTo: date)
     }
 
@@ -28,7 +30,7 @@ extension Date {
     /// If the conversion fails, returns `self` as is.
     public var shortered: Date {
         let dateFormatter = DateFormatter()
-        dateFormatter.dateFormat = "dd/mm/yyyy"
+        dateFormatter.dateFormat = "dd/MM/yyyy"
         return dateFormatter.date(from: dateString) ?? self
     }
 
--- a/Mlem/App/Utility/Extensions/Date+Extensions.swift
+++ b/Mlem/App/Utility/Extensions/Date+Extensions.swift
@@ -30,7 +30,8 @@ extension Date {
     /// If the conversion fails, returns `self` as is.
     public var shortered: Date {
         let dateFormatter = DateFormatter()
-        dateFormatter.dateFormat = "dd/MM/yyyy"
+        dateFormatter.locale = Locale.current
+//        dateFormatter.dateFormat = "dd/MM/yyyy"
         return dateFormatter.date(from: dateString) ?? self
     }
 
@@ -40,6 +41,7 @@ extension Date {
         let dateFormatter = DateFormatter()
         dateFormatter.dateStyle = .short
         dateFormatter.timeStyle = .none
+        dateFormatter.locale = Locale.current
         return dateFormatter.string(from: self)
     }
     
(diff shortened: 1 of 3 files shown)


pylapp:
@EricBAndrews @Sjmarf FYI the pull request is ready for review :)

pylapp:
FYI as usual I only submitted a pull request with cherry-picked commits.
My working branch is [that one](https://github.com/pylapp/mlem/tree/fix/2032-elapsed-time-on-profile), based on my own [dev branch](https://github.com/pylapp/mlem/tree/fix/dev) aligned to [your](https://github.com/mlemgroup/mlem/dev).
It should work on your side :)

Mlem/App/Utility/Extensions/Date+Extensions.swift
```diff
@@ -8,16 +8,33 @@
...
+        return formatter.localizedString(for: shortered, relativeTo: date)
+    }
 
-        return formatter.localizedString(for: self, relativeTo: date)
+    /// Returns `self` but with only the day, the month and the year, i.e. the current date without hours.
+    /// If the conversion fails, returns `self` as is.
+    public var shortered: Date {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "dd/mm/yyyy"
+        return dateFormatter.date(from: dateString) ?? self
```
Sjmarf: We should use the user's preferred date format rather than hard-coding it. `dd/mm/yyyy` is the preferred format in areas of Europe, but in the US it's `mm/dd/yyyy`, and in other places it's `yyyy-mm-dd`. The user can configure this in System Settings under General -> Language & Region -> Date Format

Sjmarf (changes requested):
Thanks. I am at work right now, will give this a more thorough review later. Is this maybe related to [this issue](https://github.com/mlemgroup/mlem/issues/1988)?
 
 If an account was created less than two years ago, I don't think we should be displaying "2 years ago" in the English version of the app. Rather, I think it would be less confusing to round down and display "1 year ago".  

You mention that it doesn't make sense to display "-1 a" for a 1.5-year-old account in French. I'm curious as to whether we might need to handle this differently for French, in that case? The minus isn't present in the English version - does it have any special meaning?

Mlem/App/Utility/Extensions/Date+Extensions.swift
pylapp: Nope, it does not matter that much.
In fact, this is just a format to force the date to get rid of hours, timezones etc.
The returned type is still a `Date`, the `dateFormat` is just to eliminate useless data the `RelativeDateTimeFormatter` cannot parse.
Indeed if the *shortered* value is used as a string it will be more interesting to use user preferences of course, but that's not the aim of this property :D

pylapp:
> Thanks. I am at work right now, will give this a more thorough review later. Is this maybe related to [this issue](https://github.com/mlemgroup/mlem/issues/1988)?
> 
> If an account was created less than two years ago, I don't think we should be displaying "2 years ago" in the English version of the app. Rather, I think it would be less confusing to round down and display "1 year ago".
> 
> You mention that it doesn't make sense to display "-1 a" for a 1.5-year-old account in French. I'm curious as to whether we might need to handle this differently for French, in that case? The minus isn't present in the English version - does it have any special meaning?

I cannot say of this is related to issue #1988, but when I compile locally the code with your current *dev* branch (last commit 2712ed77b8da165c37f626fab89694e3cc95c5b6) the issue I spotted is still here :D 

About the "-1 a" case, I would have said that "-1 a" or '-1 y" is not helpful for everybody, and should be extanded to "- 1 an" or "- 1 year", ie. use plain words :) 

I did not change the logic of the formatter ; just helped it to parse the date and return the suitable value.

My case is that one : 
- my account was created at 04/11/2023
- current code base displays "- 1 a", and I suppose it means "less than 1 year", which is not correct
- my fix provides "2 years agos" ; indeed it its better than current solution but not perfect as rounded higher, but this is the behaviour of the formatter (and basically 2023 was 2 years ago)

In a nutshell the current code base provides a too short value which is not correct ; my fix brings longer string with more accurate value.  You can run the code sample I shared in the issue (I used it in a playground) ; with the same date (i.e. 04/11/2023) the formatter returns 2 years, not one, and it seems it fails with dates having hours ([see this thread for exammple](https://stackoverflow.com/questions/71069385/relativedatetimeformatter-says-a-date-from-2-days-ago-is-yesterday))

I use te outputs of the formatter you use ; if you want for example "1.5 years" for a less than two years olds account, the relative parser you call should not be used. In fact, I do not change the logic how the elapsed time is computed ; I fix a misbehavior because of dates the parser cannot process :-D

Mlem/App/Utility/Extensions/Date+Extensions.swift
Sjmarf: Ah, I see! That makes sense. My apologies, I was just skim reading the code on my phone earlier and missed what this was being used for.

This is a working way to remove the hours from the `Date`, but I suspect that there might be a better way of doing this that avoids converting to a string. A better way to do this would be to create a `DateComponents` object from the `Date` - the `DateComponents` initializer lets you choose which components to include (so here, we'd be including only the year, month and date). Then, either convert the DateComponents back into a `Date` object. 

A while ago [I implemented a `roundingDownToMostSignificantComponent` function](https://github.com/mlemgroup/mlem/pull/2001/files#diff-dc7c81bdfd0ef0db51ab05c51c3f660065f2b2de1d8820c7600913a95dbd5c29) that may be useful here. If you call that function on a `DateComponents`, it removes all of the components except for the largest one.

> my fix provides "2 years agos" ; indeed it its better than current solution but not perfect as rounded higher, but this is the behaviour of the formatter (and basically 2023 was 2 years ago)

In the context of accounts, I think it's better to always round down rather than to the nearest year. For human beings, you are only "2 years old" when you have been alive for *at least* two years - if you are less than two years old, you are considered to be one year old. I think we should use the same logic here. Using `roundingDownToMostSignificantComponents` would fix this, I think.

Sjmarf: There's a `getShortRelativeTime` function in `Date+Extensions` that does a lot of this already, actually. Maybe we could reuse that function

pylapp: Ok I will give a try tomorrow. However keep in mind the relative date time formatter you used is quite "high level" so for an event occurred two years ago (i.e. in 2023), it will return "2 years ago".

I will make some trials, and if relevant share with you in this PR the results.

Let's keep in touch!

Sjmarf: Sounds good 👍 

> Ok I will give a try tomorrow. However keep in mind the relative date time formatter you used is quite "high level" so for an event occurred two years ago (i.e. in 2023), it will return "2 years ago".

It returns "1 year ago" for me. I just noticed in your test case that you're using a format of `dd/mm/yyyy` - `mm` is for minutes, not for months. For months, use `MM`. Maybe that's the reason?

pylapp: The returned value depends to the logic of the relative date time formatter, I don't have details in mind 😅

Here are the outputs I have for my account with the current implementation.

I keep you in touch with further tests tomorrow 😄

image
image

Sjmarf: Sure 👍 

The date 04/11/2023 is less than two years ago, so we should display "1 year ago". `getShortRelativeTime` returns "1 year ago" for that date :)

pylapp: Ok! FMPOV this is not the best solution but it's your project before all, so I will make the updates tomorrow with these conditions 😄

pylapp: @Sjmarf Just in case of, are you aware that if we take the rounded lower value the display is not isofunctional with some "legacy" web frontends?

I share you the screenshot for the web portal of the [instance I am in](https://programming.dev/u/pylapp). 

Just wanted to share ; I will round lower the date if you don't change your POV 😄

image

Sjmarf: Yes, I'd still prefer to round down to "1 year". What do you think @EricBAndrews?

There are plenty of places where we intentionally differ from the web frontend for design reasons, rather than being consistent - e.g by renaming "feature post" to "pin post". A large portion of our users only use Lemmy through Mlem, so if there's something we can do better than Lemmy-ui, I don't think consistency with Lemmy-ui should be holding us back.

pylapp: Ok, sounds clear! It's nice to have such details to understand the story and background of the project, In understand!
So let's take your proposals, I should take that tomorrow 😊

pylapp: And I am one of these users using Lemmy only with Mlem ☺️

pylapp: @Sjmarf Just found a more elegant fix based on your comments, will be uploaded today ;)

pylapp:
@Sjmarf I updated the branch to display the "lower" rounded time as required ;)

I also made some tests about how the string can be computed, and here are my conclusions:
- using `DateComponents` is not enough as if we round them we lose details and computation fails
- the use of `getShortRelativeTime` is interesting however it does not produce the suitable wording (missing "ago" words)
- compute date from rounded date components give a bad object the formater cannot use properly
- so, using the string formatter as I provided (and with the mm/MM fix you spotted) does the job

You can find attached the plaground with my test cases ;)

Feel free to review / amend or merge :D 

[#2032.playground.zip](https://github.com/user-attachments/files/20243965/2032.playground.zip)

Sjmarf:
Thanks, that's a very thorough investigation :)

Mlem/App/Utility/Extensions/Date+Extensions.swift
```diff
@@ -8,16 +8,35 @@
...
+        dateFormatter.dateFormat = "dd/MM/yyyy"
+        return dateFormatter.date(from: dateString) ?? self
     }
-    
-    // Returns strings like "5/10/2023"
-    var dateString: String {
+
+    /// Returns the current `Date` as a shorter version in `String`.
+    /// For example if the date in the 5th of October 2023, returns "5/10/2023"
+    public var dateString: String {
```
Sjmarf: This computed property returns a localized version of the `Date`. For my device, it returns `2023-10-05`, not `5/10/2023`. This breaks the `shortered` computed property. We need specify the date format we need (`dd/MM/yyyy`) explicitly in `dateString` in order for it to work properly in all locales.

The StackOverflow post you linked mentions that removing the hours is necessary for "yesterday" to display correctly. Is that the only case where removing the hours is needed? If so, we could stop the formatter from showing `yesterday` like this, and then we wouldn't need to remove the hours.

```swift
        let formatter = RelativeDateTimeFormatter()
        formatter.dateTimeStyle = .numeric
```

Another consideration is that for very new accounts, we currently display the minutes since creation. I suppose that removing the hours from the date would break this?

pylapp: I'll have a look on these cases :)

pylapp: @Sjmarf Branch has been updated.
So, current changes are:
- updated unit tests to ensure the `getRelativeTime()` returns the suitable values (years, days, seconds, etc.)
- used `Local.current` to not force the date format for the formatter and let it define the format depending to user preferences
- removed the now useless proeprty `shortened`

I made also quick-and-dirty screenshots to display the elapsed time (so don't deal with the non-green color or the lead, just wanted to see the values after the commas).

The several tests cases (account with age of seconds, less than one day, some days, more than one year etc.) are covered.

In few words: the bug is fixed and no issues with format of date and minutes / seconds are kept :D 

hour
hours
seconds
year

pylapp: @Sjmarf As a user I have a question: what does the green leaf mean?
(Don't mind with my previous screenshots if it was suppoed to have on leaf or not ; I juste tinkered the view in a dirty way for tests)

Sjmarf: Nice, works perfectly :)

The green leaf means that the account is less than 30 days old. It's also shown next to the user's name in posts and comments:

MlemTests/DateTests.swift
```diff
@@ -0,0 +1,153 @@
...
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd HH:mm:ss Z"
+        let someDate = dateFormatter.date(from: "2023-11-04 10:05:00 +0000")!
+        
+        // When
+        let dateString = someDate.dateString
+        
+        // Then
+        #expect(dateString == "04/11/2023")
+    }
```
Sjmarf: This should be removed, right?

```suggestion
```

Sjmarf (changes requested):

MlemTests/DateTests.swift
pylapp: I will have a look soon, but yes indeed

pylapp:
@Sjmarf I removed the useless unit test and rebased the branch ; now your turn to review again and again (^_^)

Sjmarf (approved):
Nice job! 😄

pylapp:
> Nice job! 😄

Thank you a lot for your comments, patience and details!

Just a dumb question for myself as a user: I see in some profiles a kind of green leaf aside green text: what does it mean?

Thank you for this awesome app, I am glad to help 😄

Sjmarf:
You're welcome, thank _you_ for contributing! I answered the leaf question [here](https://github.com/mlemgroup/mlem/pull/2033#discussion_r2093326265) :)

PR #2035, opened 2025-05-15
[#2034] Add missing french translations, fix basic toast instanciation for `urlCopyError`case

### LINKED ISSUE #2034: Missing wording for french, Toast for URL share is not translated
### Requirements

### Description

- Some wording in french are missing
- The toast use for hyperlink share it not translated in case of missing URL

# Note

Pull request will come in next minutes :)

Toast always in english

### Mlem Version

2.2 (local build)

pylapp:
## Issues

- closes #2034

## Description

I spotted in *dev* rbanch some missing wording in french, mainly for toasts.
The toast displayed in case of error during URL copy is also never translated.

## Implementation Notes

- Added missing translations
- Fix generation of basic toast (using the static helper and not directly the enum case as explained in comment in the code)

### DIFF
--- a/Mlem/Localizable.xcstrings
+++ b/Mlem/Localizable.xcstrings
@@ -9203,9 +9203,6 @@
       }
     },
     "Would you like to open this email address in your mail app?" : {
-<<<<<<< HEAD
-
-=======
       "localizations" : {
         "fr" : {
           "stringUnit" : {
@@ -9214,7 +9211,6 @@
           }
         }
       }
->>>>>>> 80b76dfa (fix: missing french translations, basic toast for URL copy error never translated (#2034))
     },
     "Wrap Code Block Lines" : {
       "localizations" : {
--- a/Mlem/Localizable.xcstrings
+++ b/Mlem/Localizable.xcstrings
@@ -604,7 +604,14 @@
       }
     },
     "Account Created %@" : {
-
+      "localizations" : {
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Compte crée %@"
+          }
+        }
+      }
     },
     "Accounts" : {
       "localizations" : {
@@ -1830,7 +1837,14 @@
       }
     },
     "Choose whether to show a user's account age next to their username." : {
-
+      "localizations" : {
+        "fr" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Choisir d'afficher ou non l'âge du compte d'un utilisateur à côté de son nom d'utilisateur."
+          }
+        }
+      }
     },
     "Choose whether to show a warning when opening a page that is likely to contain sensitive content." : {
       "localizations" : {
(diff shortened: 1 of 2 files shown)


pylapp:
@Sjmarf @EricBAndrews Spotted little issues with localization, brought fixes :-D
PR ready for review!

Mlem/Localizable.xcstrings
```diff
@@ -2339,6 +2353,17 @@
         }
       }
     },
+    "Copy a URL to the clipboard, then try again." : {
+      "extractionState" : "manual",
```
Sjmarf: This line indicates that you entered the string into the `Localizable` file manually, rather than letting Xcode add the string to the file. We should never need to add strings manually - after you made the above change to the code, the string would have appeared in the `Localizable` file when you pressed build.

Because this string has the `"manual"` extraction state, Xcode won't automatically mark it as "stale" if we later remove this string, so we need to remove this line :)

```suggestion
```

Sjmarf (changes requested):

Mlem/Localizable.xcstrings
pylapp: I agree however Xcode never added those strings in the localizables ; that is the reason why I added them manually.

Do you suggest to have a try and defining them as "automatic"? I let you confirm that and then I'll update the PR ;)

Sjmarf: If you go in the `Localizable` file and delete the row for this string, Xcode will add it back in again next time you build the project (⌘B). Xcode only updates the `Localizable` file after you build the project. You will see it say "Syncing localizations" in the top-right when it is doing this.

Screenshot 2025-05-15 at 20 01 23

pylapp: Not convinced, did not see this behaviou yet in the project but I will check again tomorrow 😄

Anyway, I keep you in touch here 😃

Sjmarf: I've just checked and Xcode does add it back in if I delete it :) Xcode won't have done this automatically before you made the change from `basic(title: ...` to just `basic(...,` but it does now that you've made that change.

pylapp: Ah ok! Didn't make tests since I changed the call. Indeed in that case Xcode spots the strings and add them, nice 😃

pylapp:
@Sjmarf I applied your suggestions and rebased the branch to solve conflcits ; you can now review this PR again :)

Mlem/Localizable.xcstrings
```diff
@@ -9126,7 +9203,18 @@
       }
     },
     "Would you like to open this email address in your mail app?" : {
+<<<<<<< HEAD
```
Sjmarf: Conflict needs resolving :)

Sjmarf (changes requested):

Mlem/Localizable.xcstrings
pylapp: Grumph, missed that, fix will be uploaded this afternoon :)

pylapp:
@Sjmarf Ready for review again (^_^)

Sjmarf (approved):
Nice, looks good! 😄

PR #2039, opened 2025-05-18
Update MlemMiddleware to support latest 1.0.0 alpha

Sjmarf:
> [!note]
> See also [this MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/19) and [this codegen PR](https://github.com/mlemgroup/lemmy-swift-codegen/pull/14).

This PR updates MlemApiTypes to use the latest state of Lemmy 1.0. Since the last time we updated to support 1.0.0, Lemmy has made some changes that necessitates some refactoring on our side. This is **not** an attempt to actually support 1.0.0 - just to support those new api type schemas.

Before this PR, the middleware system was composed of two layers: The front-facing models (e.g. `Post1`), and the auto-generated api types (e.g. `ApiPost`).

This worked great up until now, but there was a little complexity involved in some cases where there was differences between versions. In cases where properties of api types were directly renamed, codegen was configured to merge those two properties into one. Codegen would generate a custom `init(from decoder: Decoder)` for the struct that combines them back into one.

Another example is when there was multiple api types that we wanted to create a middleware model from. Before this PR, we did this by creating a "backer" protocol (e.g. `ApiReplyBacker`) that each middleware type would conform to. This worked, but had some disadvantages:
- Sometimes we had to use less-than-ideal names for computed properties of the protocol to avoid shadowing the property names of implementing types, e.g. `resolvedSaved`.
- The caches for such models weren't able to inherit from `ApiTypeBackedCache`, which meant duplicating quite a bit of code for the convenience methods that `ApiTypeBackedCache` would otherwise provide (`getModels` and `getOptionalModel`)

In 1.0.0, Lemmy has now implemented some further changes to the schema that can't be solved as cleanly with the same approaches - specifically, they've moved some properties up and down model hierarchies.

Rather than making codegen much more complex in order to support this, I've opted to instead introduce a new translation layer between our two existing layers. This layer consists of a number of "snapshot" structs. There is a snapshot struct for each middleware model (`Post1Snapshot`, `Post2Snapshot`, etc). Each snapshot uses the same property names and types as their respective middleware models. You can think of a snapshot type as a "snapshot" of the state of that model at a given time.

E.g.

```swift
public struct Post1Snapshot: CacheIdentifiable {
    public let actorId: ActorIdentifier
    public let id: Int
    public let creatorId: Int
    public let communityId: Int
    // ...

    init(from post: ApiPost) throws(ApiClientError) {
        // ...
    }
}
```

Each model cache now accepts a snapshot rather than an api type. It is the responsibility of the ApiClient to convert whichever api type it receives into a snapshot before passing it to the cache.

Some other advantages of this system:
1) It reduces some of the code duplication that we would previously do in the cache and in a model's `update` method, if we needed to convert any of the values from an api type into another format.
2) In future, if we chose to support another platform, we could reasonably easily make a second set of API types (e.g. `PiefedPost`, `PiefedCommunity`), and those could also be converted into `Snapshot` types.
3) For the "recent searches" functionality, we need to be able to save a model to the disk. Currently, we do this by converting to an api type. This isn't ideal. With this system, we can instead convert the model to a snapshot and serialise that (I haven't actually implemented this yet).


EricBAndrews (approved):
LGTM! I think the snapshot system is a nice way to handle versioning discrepancies.

PR #2046, opened 2025-05-21
Clean up middleware

Sjmarf:
Cleaned up some things I missed in #2039. 

Some of these were infinite recursions, which could cause hangs in theory (though I haven't encountered any in practice, yet). This needs to be merged before the next TF release just in case. Don't include this in the patch notes - #2039 hasn't been released to the TF yet.


EricBAndrews (approved):
LGTM! 💯

PR #2059, opened 2025-05-24
New Sign Up System (part 1)

Sjmarf:
Added *part* of the new sign-up system - in the interest of not submitting an enormous PR, I'm going to implement it in bits. This PR contains the "join Lemmy.world" and "choose a username" screens. These pages are currently only accessible via a button in Developer settings - they aren't actually used for onboarding just yet.

The steps of the sign-up process will be as follows:
- "Join Lemmy.world or choose another instance" (implemented)
- Choose a username (implemented)
- Enter email
- Enter password
- Enter captcha & agree to TOS
- Verify email
- Wait for automated Lemmy.world approval
- Automatically signed into new account


Mlem/App/Views/Root/Login/Onboarding/SignUpUsernameView.swift
```diff
@@ -0,0 +1,135 @@
...
+    @State var usernameValidity: UsernameValidity?
+    
+    var body: some View {
+        ScrollView {
+            VStack {
+                Text("Choose a Username")
+                    .font(.title)
+                    .fontWeight(.bold)
+                    .padding(.top, 50)
+                Text("Be careful - you can't change this later.")
```
EricBAndrews:
```suggestion
                Text("This cannot be changed later.")
```

Mlem/App/Views/Root/Login/Onboarding/SignUpUsernameView.swift
```diff
@@ -0,0 +1,135 @@
...
+    @ViewBuilder
+    var nextButtonView: some View {
+        Button(action: submit) {
+            Text("Next")
+                .frame(maxWidth: .infinity)
+                .padding(.vertical, 10)
+        }
+        .buttonStyle(.borderedProminent)
+        .buttonBorderShape(.roundedRectangle(radius: 16))
+        // IMO it's less jarring to only disable the button once they try to tap on it - otherwise the user would see the button flicker as they type out their username
```
EricBAndrews: The best way to handle this IMO would be to perform a debounced validity check--start disabled, then once the user hasn't typed for a certain amount of time check availability and validity. This should also be accompanied by a `ProgressView`.

Happy to leave OOS/continue working this problem in future PRs, since this is just part 1

EricBAndrews (changes requested):
Nice work!

The dark mode version is _way_ better than light mode. I think we should:
- Make the onboarding a `.fullScreenCover`
- Use the dark mode as the start screen for the onboarding flow regardless of system color scheme

We could even think about making this flow a single-page operation--"Let's Go" slides the first "page" out of view and brings in the next couple fields, etc. as the user progresses through the flow.

Mlem/App/Views/Root/Login/Onboarding/SignUpUsernameView.swift
Sjmarf: I agree, that's a much better way to do it 👍

Sjmarf:
> We could even think about making this flow a single-page operation--"Let's Go" slides the first "page" out of view and brings in the next couple fields, etc. as the user progresses through the flow.

That's a cool idea. I'll have a go at it 👍 

> The dark mode version is way better than light mode. I think we should [...] Use the dark mode as the start screen for the onboarding flow regardless of system color scheme

Hmm, I'm not so sure. Ideally, I think we should find a way to respect the user's system theme. I do also prefer the dark mode screen, but I think the light mode version looks good enough. Idk, maybe I've just been looking at it for too long. The problem with only using the dark mode one is that we'd have to either:
- Switch to light mode in the transition to the username field, which would be jarring, or
- Put the entire onboarding experience in dark mode

Sjmarf:
What about something like this?

EricBAndrews:
> What about something like this?
> 
> 
> 
> 
> 
> 

That looks great!

Sjmarf:
Moved both views to a single page, with a transition between them 👍

EricBAndrews (approved):
LGTM! 🔥

PR #2060, opened 2025-05-25
fix post editor search error

### LINKED ISSUE #2058: Searching for community when creating post from profile fails
### Requirements

### Description

Profile -> New Post -> attempt to search for community. Throws 400 error. 

Reported [here](https://lemmy.ml/post/30253274/18738281)

### Mlem Version

_No response_

EricBAndrews:
## Issues

- closes #2058 

## Description

Searching within your own instance sometimes causes resolution failures. I'm not sure at this point whether that's an issue when searching on foreign hosts--I haven't been able to repro in that context.

In either event, there's no reason to perform the resolution step if the search is being made on `hostApi`, so I've added a check to disable it in that case.

I've also added error handling on the resolution step to prevent the entire search from failing if the resolution step fails--it's better IMO to display the search results sans subscription status than fail the whole search.


Sjmarf (approved):
Nice! 💯

PR #2064, opened 2025-05-29
Move haptics to package

### LINKED ISSUE #1106: [2.0] Remove `.sentinel` haptic?
https://github.com/mlemgroup/mlem/pull/1101#discussion_r1645017762

### LINKED ISSUE #2062: Move haptic system to package

Sjmarf:
Moved haptics to a separate package. 

Also made the `HapticManager` available as an environment object, and deprecated accessing the singleton using `.main`. I think we should generally avoid globally available singletons in future because it makes SwiftUI previews a pain to write. 

The only place where this change is slightly annoying is the action system, which I've got some ideas on how to fix later down the road. For now, the actions continue to access the `HapticManager` via the `main` property (which is marked as deprecated).

Closes #2062
Closes #1106


Mlem/App/Configuration/User Settings/SettingsValues.swift
```diff
@@ -130,7 +131,14 @@ class SettingsValues: Codable { // swiftlint:disable:this type_body_length
         self.behavior_biometricUnlock = try container.decodeIfPresent(Bool.self, forKey: ._behavior_biometricUnlock) ?? false
         self.behavior_confirmImageUploads = try container.decodeIfPresent(Bool.self, forKey: ._behavior_confirmImageUploads) ?? true
         self.behavior_enableQuickSwipes = try container.decodeIfPresent(Bool.self, forKey: ._behavior_enableQuickSwipes) ?? true
-        self.behavior_hapticLevel = try container.decodeIfPresent(HapticPriority.self, forKey: ._behavior_hapticLevel) ?? .high
+        
+        do {
+            self.behavior_hapticLevel = try container.decodeIfPresent(HapticLevel.self, forKey: ._behavior_hapticLevel) ?? .high
+        } catch DecodingError.dataCorrupted { // Decodes the 'sentinel' value, which was replaced with `nil` in Mlem 2.2
+            print("DECODE SENTINEL")
```
EricBAndrews: Debug print?
```suggestion
```

EricBAndrews (changes requested):
Nice work!

I'm not sure I agree with the "priority" -> "level" rename--"level" in the context of a haptic implies a degree of haptic intensity rather than a binary play/don't play filter.

Sjmarf:
That's a good point 🤔 I'm not sure I like "priority" either, though--to me, that implies there's a queue of something and a higher priority puts it further to the front of the queue. 

Maybe "importance"?

EricBAndrews:
I like "importance"--"tier" may capture the same idea in fewer characters

Sjmarf:
I like "tier" 👍

EricBAndrews (approved):
LGTM! 🔥

PR #2066, opened 2025-05-30
Remove duplicate active user count from community details

Sjmarf:
Git merge went wrong at some stage, I think


EricBAndrews (approved):
Nice 👍

Issue #1550, opened 2024-12-27
Allow users with "Mlem Developer" flair to toggle it on/off


Sjmarf:
Requires #1549

PR #2068, opened 2025-06-02
Backend

### LINKED ISSUE #1313: Mlem HTTP Server
We need this for a number of features ([list here](https://github.com/mlemgroup/mlem/issues?q=is:open+is:issue+label:%22Needs+Mlem+Server%22)).

Doing this will involve storing a user's auth token on the server. We should probably have it be a separate auth token from the Mlem app. Do we need to update the privacy policy if we want to do this?

I haven't looked that deeply into what languages or frameworks we could use for this. Using Swift might be a good idea, because we're already using it. Some brief researching suggests https://vapor.codes could be good? [They've got built-in Apple Push Notification support](https://docs.vapor.codes/advanced/apns/).

[Memmy's server is open source](https://github.com/Memmy-App/memmy-api) and written in TypeScript.

### LINKED ISSUE #1549: Fetch list of users who have "Mlem Developer" flair from external source rather than hard-coding
This would allow us to update the list remotely. 

Could be done without #1313, but doing this with #1313 would enable #1550.

### LINKED ISSUE #1979: Move MlemStats responsibilities to Mlem server
The logic we currently do in MlemStats should be done on the Mlem server instead. At the moment we rely on lemmyverse.net; it would be nice to not be dependent on them if possible, though that may be too difficult to implement.

EricBAndrews:
## Issues

- closes #1313 
- closes #1979 
- closes #1549
- progress towards #1550
- progress towards #1513

## Description

Replaces the following functionality with calls to the Mlem backend:
- Testflight update: latest update post URL is now fetched from the backend
- Developer flairs: are now fetched from the backend instead of hardcoded
- Instance search: instances are now fetched from the backend instead of mlem-stats

## Implementation Notes

This PR is meant to be unobtrusive; I have mostly tried to just plug in the backend for whatever data source each functionality was using before (e.g., instance search still fetches the list of instances and performs searching locally).

Calls to the backend happen through the `BackendClient`. I have chosen to place this _in_ `MlemMiddleware` but _not_ bundle the calls into `ApiClient`; `BackendClient` and `ApiClient` may wish to refer to each other, and share constructs such as `SiteVersion`, but the two entities have entirely different scopes.


Sjmarf:
Awesome work!

> progress towards https://github.com/mlemgroup/mlem/issues/1550

Did you mean #1513? The `isActive` boolean seems to be referring to that issue to me - correct me if I'm wrong.

Edit: Ah, I read `isActive` as "is an active contributor", but I now realise you probably mean "the flair itself is active". If the user has disabled their developer flair, maybe it's better to just not list their name in the response at all?

> Calls to the backend happen through the BackendClient. I have chosen to place this in MlemMiddleware but not bundle the calls into ApiClient; BackendClient and ApiClient may wish to refer to each other, and share constructs such as SiteVersion, but the two entities have entirely different scopes.

Makes sense to me 👍

EricBAndrews:
> The `isActive` boolean seems to be referring to that issue to me - correct me if I'm wrong.

Yeah, that is ambiguous. I've reworked how the backend flairing system works:
- The backend table now has three columns: `ap_id`, `flair_type`, and `flair_enabled`
- Currently there are two `flair_type`s: `activeDev` and `inactiveDev`
- The endpoint has been updated from `/mlem/developers` to `/mlem/flairs`. It takes an optional query parameter `enabledOnly`, which defaults to false; if true, only enabled flairs will be returned. This should in the future allow us to give users with a disabled flair the ability to enable it, though that's an authenticated call so will need to wait until we have Mlem accounts set up.
- `BackendClient` now stores `flairs: MlemFlairs`, which contains sets for each flair type. Currently `activeDev` and `inactiveDev` both just get mapped into `developers`, since the scope of this PR is just plugging in the backend without making design changes.

Sjmarf:
Nice, that makes sense to me 👍

Mlem/App/Views/Root/Tabs/Settings/DeveloperSettingsView.swift
```diff
@@ -28,6 +34,28 @@ struct DeveloperSettingsView: View {
...
+                        Image(systemName: Icons.present)
+                            .foregroundStyle(backendStatus ? .themedPositive : .themedNegative)
+                    } else {
+                        ProgressView()
+                    }
+                }
+                
+                Button("Refresh") { checkBackendStatus() }
+            } header: {
+                Text("Backend")
```
Sjmarf:
```suggestion
                Text(verbatim: "Backend")
```

Mlem/App/Views/Root/Tabs/Settings/DeveloperSettingsView.swift
```diff
@@ -28,6 +34,28 @@ struct DeveloperSettingsView: View {
...
+            } footer: {
+                if let lastBackendStatusCheck {
+                    Text("Refreshed \(lastBackendStatusCheck.formatted(date: .abbreviated, time: .standard))")
```
Sjmarf:
```suggestion
                    Text(verbatim: "Refreshed \(lastBackendStatusCheck.formatted(date: .abbreviated, time: .standard))")
```

Mlem/App/Views/Root/Tabs/Settings/DeveloperSettingsView.swift
```diff
@@ -28,6 +34,28 @@ struct DeveloperSettingsView: View {
...
+                }
+            }
+            .task { checkBackendStatus() }
```
Sjmarf:
```suggestion
            .onAppear { checkBackendStatus() }
```

Or, we could make `checkBackendStatus` an `async` method

Mlem/App/Views/Root/Tabs/Settings/DeveloperSettingsView.swift
```diff
@@ -28,6 +34,28 @@ struct DeveloperSettingsView: View {
...
+                    }
+                }
+                
+                Button("Refresh") { checkBackendStatus() }
+            } header: {
+                Text("Backend")
+            } footer: {
+                if let lastBackendStatusCheck {
+                    Text("Refreshed \(lastBackendStatusCheck.formatted(date: .abbreviated, time: .standard))")
+                }
```
Sjmarf: *nit*

When you open the page this text takes a second to load, which is a little jarring - this could be fixed with something like

```suggestion
                if let lastBackendStatusCheck {
                    Text(verbatim: "Refreshed \(lastBackendStatusCheck.formatted(date: .abbreviated, time: .standard))")
                } else {
                    Text(verbatim: "Refreshing...")
                }
```

Then again this is the developer settings page so it doesn't really matter lol

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Backend Client/BackendClient.swift
```diff
@@ -0,0 +1,90 @@
...
+            do {
+                try await fetchFlairs()
+                try await fetchTestflightUpdate()
+            } catch {
+                print(error)
+            }
+        }
+    }
+    
+    public func healthcheck() async throws -> Bool {
```
Sjmarf:
```suggestion
    public func healthCheck() async throws -> Bool {
```

Mlem/App/Views/Root/Tabs/Settings/DeveloperSettingsView.swift
```diff
@@ -28,6 +34,28 @@ struct DeveloperSettingsView: View {
                 NavigationLink(String("Error Log"), destination: .settings(.errorLog))
             }
             
+            Section {
+                HStack {
+                    Text("Status")
```
Sjmarf:
```suggestion
                    Text(verbatim: "Status")
```

Mlem/App/Views/Root/Tabs/Settings/DeveloperSettingsView.swift
```diff
@@ -8,18 +8,24 @@
...
+    @Environment(BackendClient.self) var backendClient
+    
     @Environment(NavigationLayer.self) var navigation
     @Dependency(\.persistenceRepository) var persistenceRepository
     
     @Setting(\.tip_feedWelcomePrompt) var showFeedWelcomePrompt
     @Setting(\.dev_developerMode) var developerMode
     
-    @AppStorage("lastBuildNumber") var lastBuildNumber: String?
+    @AppStorage("lastTestflightUpdate") var lastTestflightUpdate: URL?
```
Sjmarf: Should `testflight` be camel-cased (`testFlight`)? This is one of those cases where it could go either way - we just need to be consistent. I'm inclined to go with `testFlight` because that's how Apple spells it. What do you think? I think it's worth deciding this now, because we're using it in an `@AppStorage` key which makes it cumbersome to change later.

Sjmarf (changes requested):
Awesome work! It's great to get our own server working; we've been talking about doing this for ages 🚀 🚀 A few minor points:

Mlem/App/Views/Root/Tabs/Settings/DeveloperSettingsView.swift
EricBAndrews: `testFlight` works for me, updated 👍

Sjmarf (approved):
LGTM! 🔥

Issue #2070, opened 2025-06-04
Use a custom Nuke ImagePipelineDelegate to apply the User-Agent header to image requests


Sjmarf:
This would save us calling `mlemUrlRequest` everywhere. The delegate should be defined in the media package. We can set it up along with the rest of our Nuke configuration in `MlemApp`. (Maybe that Nuke configuration can also move to the Media package)?

I haven't looked into the specifics too deeply yet but I expect it should be possible through some combination of a custom `ImagePipelineDelegate`, a `DataLoading` (both from Nuke) and/or a `URLSessionConfiguration` with `httpAdditionalHeaders`.

PR #2071, opened 2025-06-04
Move some ApiClient code into a RestClient class

Sjmarf:
> [!note]
> See also [this MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/21) and [this lemmy-swift-codegen PR](https://github.com/mlemgroup/lemmy-swift-codegen/pull/16).

Moved some of the `ApiClient` code into a `RestClient` class, which `ApiClient` uses internally. 

`RestClient` is in its own package, `Rest`. This means that `Rest` needs to be imported at the top of all of the Api type files, hence the MlemApiTypes PR.

At the moment `Rest` needs to be imported into Mlem, too, but this is only loosely required and we can remove that dependency once #2070 is done.

Separating out `RestClient` in this way will make it easier to implement PieFed support, and also makes it easy to use `RestClient` in other places (e.g. `BackendClient`) if we wanted to.


Sjmarf:
Oops, forgot to push

EricBAndrews (approved):
Nice work! This should be very helpful 💯

PR #2073, opened 2025-06-06
Fix mod mail showing the reporter as the reported content creator

Sjmarf:
This bug was introduced when I added the snapshot structs (so it's not in the App Store version of Mlem)


EricBAndrews (approved):
Nice 👍

PR #2074, opened 2025-06-06
Fix haptics not playing

Sjmarf:
Missed this in #2064. This should be merged before any further TF releases, as it fixes a bug introduced since the last TF release


EricBAndrews (approved):
Nice 👍

PR #2075, opened 2025-06-07
Add collapse, collapse parent, and collapse to top comment actions

### LINKED ISSUE #210: Swipe to collapse top-level comment
## Introduction
In Apollo the swipe-gestures on comments could be configured, e.g. to collapse the topmost comment of a given thread, without having to scroll back up.

## Requirements
It would be cool if Mlem could be configured similarly, e.g. if there could be a setting for customizing the actions revealed upon swiping left or right on a comment and if a "collapse top-level comment" action could be included.

## Discussion
There could be other ways to represent this in the UI, a swipe is probably one of the most convenient ones that doesn't overlap with the existing "tap-to-collapse". Making it customizable would let users pick whichever action they'd like to reveal on swipe. (The default could stay exactly as-is currently)

un1970ix:
## Issues

- closes #210

## Description
Adds three new comment collapse swipe actions: Collapse, Collapse Parent, and Collapse to Top. These actions allow users to quickly collapse comments and comment threads.

## Implementation Notes
- Uses existing CommentTreeTracker with minimal new code
- Added 6 new SF Symbols icons for filled and unfilled states of each action
- Animated scrolling automatically handles focus after collapse using `onChange(of: node.collapsed)`
- Actions are available for user configuration <ins>but not included in defaults</ins>

### DIFF
--- a/Mlem/App/Models/Action/ActionAppearance+StaticValues.swift
+++ b/Mlem/App/Models/Action/ActionAppearance+StaticValues.swift
@@ -285,7 +285,8 @@ extension ActionAppearance {
             label: "Collapse",
             color: .themedColorfulAccent(4),
             icon: Icons.collapse,
-            swipeIcon2: Icons.collapseFill
+            swipeIcon1: Icons.collapseSquare,
+            swipeIcon2: Icons.collapseSquareFill
         )
     }
     
@@ -294,7 +295,8 @@ extension ActionAppearance {
             label: "Collapse Parent",
             color: .themedColorfulAccent(4),
             icon: Icons.collapseParent,
-            swipeIcon2: Icons.collapseParentFill
+            swipeIcon1: Icons.collapseParentSquare,
+            swipeIcon2: Icons.collapseParentSquareFill
         )
     }
     
@@ -303,7 +305,8 @@ extension ActionAppearance {
             label: "Collapse to Top",
             color: .themedColorfulAccent(4),
             icon: Icons.collapseToTop,
-            swipeIcon2: Icons.collapseToTopFill
+            swipeIcon1: Icons.collapseToTopSquare,
+            swipeIcon2: Icons.collapseToTopSquareFill
         )
     }
 }
--- a/Mlem/App/Models/Action/ActionAppearance+StaticValues.swift
+++ b/Mlem/App/Models/Action/ActionAppearance+StaticValues.swift
@@ -279,4 +279,31 @@ extension ActionAppearance {
     static func viewVotes() -> Self {
         .init(label: "View Votes", color: .themedAccent, icon: Icons.votes)
     }
+        
+    static func collapse() -> Self {
+        .init(
+            label: "Collapse",
+            color: .themedColorfulAccent(4),
+            icon: Icons.collapse,
+            swipeIcon2: Icons.collapseFill
+        )
+    }
+    
+    static func collapseParent() -> Self {
+        .init(
+            label: "Collapse Parent",
+            color: .themedColorfulAccent(4),
+            icon: Icons.collapseParent,
+            swipeIcon2: Icons.collapseParentFill
+        )
+    }
+    
+    static func collapseToTop() -> Self {
+        .init(
+            label: "Collapse to Top",
+            color: .themedColorfulAccent(4),
+            icon: Icons.collapseToTop,
+            swipeIcon2: Icons.collapseToTopFill
+        )
+    }
 }
--- a/Mlem/App/Utility/Extensions/Content Models/Comment1Providing+Extensions.swift
+++ b/Mlem/App/Utility/Extensions/Content Models/Comment1Providing+Extensions.swift
@@ -115,6 +115,7 @@ extension Comment1Providing {
         false
     }
     
+    // swiftlint:disable:next cyclomatic_complexity
     func action(
         appState: AppState,
         type: CommentBarConfiguration.ActionType,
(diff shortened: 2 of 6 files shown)


Mlem/App/Utility/Extensions/Content Models/Comment1Providing+Extensions.swift
```diff
@@ -188,4 +192,34 @@ extension Comment1Providing {
             callback: callback
         )
     }
+    
+    func collapseAction(commentTreeTracker: CommentTreeTracker? = nil) -> BasicAction {
+        .init(
+            id: "collapse\(uid)",
+            appearance: .collapse(),
+            callback: { @MainActor in
+                commentTreeTracker?.nodesKeyedByActorId[self.actorId]?.collapsed = true
```
Sjmarf: We can use `withAnimation` here to give swipe-to-collapse the same animation as when you tap-to-collapse:

```suggestion
                withAnimation(UIAccessibility.isReduceMotionEnabled ? nil : .default) {
                    commentTreeTracker?.nodesKeyedByActorId[self.actorId]?.collapsed = true
                }
```

(This change would be repeated for the other collapse swipe actions too)

Mlem/App/Models/Action/ActionAppearance+StaticValues.swift
```diff
@@ -279,4 +279,31 @@ extension ActionAppearance {
     static func viewVotes() -> Self {
         .init(label: "View Votes", color: .themedAccent, icon: Icons.votes)
     }
+        
+    static func collapse() -> Self {
+        .init(
+            label: "Collapse",
+            color: .themedColorfulAccent(4),
+            icon: Icons.collapse,
+            swipeIcon2: Icons.collapseFill
```
Sjmarf: In the swipe action settings, the icons still have squares around them:



I think we shouldn't have squared icons in the settings page - this can be done by writing something like:

```suggestion
            icon: Icons.collapse,
            swipeAction1: Icons.collapseSquare,
            swipeIcon2: Icons.collapseSquareFill
```

```swift
  // Icons
  static let collapse: String = "minus"
  static let collapse: String = "minus.rectangle"
  static let collapseFill: String = "minus.rectangle.fill"
```

Mlem/App/Utility/Extensions/Content Models/Comment1Providing+Extensions.swift
```diff
@@ -188,4 +192,34 @@ extension Comment1Providing {
...
+            }
+        )
+    }
+    
+    func collapseParentAction(commentTreeTracker: CommentTreeTracker? = nil) -> BasicAction {
+        .init(
+            id: "collapseParent\(uid)",
+            appearance: .collapseParent(),
+            callback: { @MainActor in
+                commentTreeTracker?.nodesKeyedByActorId[self.actorId]?.parent?.collapsed = true
```
Sjmarf: This should be `.toggle()` - the user should be able to swipe to un-collapse when the comment is collapsed.
```suggestion
                commentTreeTracker?.nodesKeyedByActorId[self.actorId]?.parent?.collapsed.toggle()
```

(Ditto for the other collapse actions)

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView+Views.swift
```diff
@@ -42,9 +43,19 @@ extension ExpandedPostView {
...
+                        }
+                    }
+                    .onChange(of: node.collapsed) { _, isCollapsed in
+                        guard isCollapsed else { return }
+                        
+                        withAnimation(UIAccessibility.isReduceMotionEnabled ? nil : .default) {
+                            if node == node.topParent,
+                               let topParentIndex = tracker.nodes.firstIndex(of: node),
+                               topParentIndex + 1 < tracker.nodes.count {
+                                scrollProxy.scrollTo(tracker.nodes[topParentIndex + 1].actorId, anchor: .top)
```
Sjmarf: This is a cool idea! It works well when the comment that the user is reading is at the top of the page. However, it produces some undesirable behavior when the comment the user is reading is not already at the top:

https://github.com/user-attachments/assets/8bd1dbaa-8f4b-440a-a5b5-c9c97d59b288

I don't think we should be scrolling in the case shown in the video. Unfortunately, I'm not sure whether there's an easy way around this. It might be best to keep the behaviour we have right now (just scroll to the comment when it's collapsed using `scrollProxy.scrollTo(comment.actorId)`).

Sjmarf (changes requested):
Nice, works well! 😄 A few comments

Mlem/App/Utility/Extensions/Content Models/Comment1Providing+Extensions.swift
```diff
@@ -188,4 +193,40 @@ extension Comment1Providing {
...
+        )
+    }
+    
+    func collapseParentAction(commentTreeTracker: CommentTreeTracker? = nil) -> BasicAction {
+        .init(
+            id: "collapseParent\(uid)",
+            appearance: .collapseParent(),
+            callback: { @MainActor in
+                withAnimation(UIAccessibility.isReduceMotionEnabled ? nil : .default) {
+                    commentTreeTracker?.nodesKeyedByActorId[self.actorId]?.parent?.collapsed.toggle()
```
Sjmarf: I wonder if this should coalesce to the comment itself if there's no parent 🤔 Something like:

```suggestion
                    guard let comment = commentTreeTracker?.nodesKeyedByActorId[self.actorId] else { return }
                    (comment?.parent ?? comment).collapsed.toggle()
```

This already works for the "collapse to top" option, so no changes are needed there.

Sjmarf (approved):
Looks good to me 🚀 I think you're got merge permissions?

un1970ix:
Yeah, first time using it. 🎊

PR #2077, opened 2025-06-08
LemmyConnector

Sjmarf:
Added a new `LemmyConnector` class. It provides a very similar interface to `ApiClient` but, instead of returning content models (e.g. `Post2`) each method returns snapshots (e.g. `Post2Snapshot`).

`LemmyConnector` is basically an `ApiClient`, but stateless (no caching). `ApiClient` has been modified to use `LemmyConnector` internally. Each `ApiClient` method calls a method of `LemmyConnector`, and applies the caching layer ontop of it.

`LemmyConnector` conforms to an `InstanceConnector` protocol. In future, there will be a `PiefedConnector` that also conforms to `InstanceConnector`. `ApiClient` will be responsible for deciding which connector it should use.

There's still some work to do, which I'll implement in future PRs:
- Image uploads are still handled by `ApiClient` directly - in future I will also move this into `LemmyConnector`.
- A few of the `InstanceConnector` methods still return raw Api types - all of these will need to be converted to software-agnostic types (e.g. `Post1Snapshot`) before we add `PiefedConnector`.


EricBAndrews (changes requested):
Nice work! This is a huge step towards PieFed support.

Happy to leave OOS, but I think we should take this opportunity to think about how `MlemMiddleware` should be organized--with both the `BackendClient` and soon two `InstanceConnection`s, it's getting busy in there.

Currently our top level folders are:
- API Client
- Backend Client
- Constants
- Content Models
- Custom API Models
- Extensions
- FeedLoaders
- InstanceConnection
- MlemApiTypes
- Protocols
- Snapshot

Some basic groupings we could apply:
- Internal
  - FeedLoaders
  - Content Models
- Mlem Backend
  - Backend Client
  - Subfolders organizing Backend Client stuff
- Instances
  - API Client
  - InstanceConnection
  - Snapshot
  - Custom API Models
  - MlemApiTypes
- Constants
- Extensions
- Protocols

There are a couple other considerations, too:
- Are PieFed API types included in MlemApiTypes, or will they be separate?
- Folder names should be standardized to either all have spaces between words or none. Personally I'm inclined towards spaces, since that's how the main app tree behaves.

Sjmarf:
Maybe we could do some more package separation. We could add the following three packages:

## (Can't think of a name)
- All of the `Snapshot` structs
- `InstanceConnector` protocol

## Lemmy
- Imports the above package
- `LemmyConnector`
- All of the Lemmy api types (which would be `internal` to this package)

## Piefed
- Also imports the `InstanceConnector` package
- `PiefedConnector`
- The Piefed API types

-----------

It's not quite possible to do this yet because some raw API types are still used in `ApiClient`. I agree it would be a good idea to organise the folders in MlemMiddleware a bit better, too. That folder structure looks good to me. I'm gonna leave it OOS though until I've had a chance to extract into packages, if you agree with that plan

> Are PieFed API types included in MlemApiTypes, or will they be separate?

They'll be generated by the same script—with the package architecture described above, the generated types would need to go in a new repo. Or, we just put them in the main repo (and move the Lemmy API types there too).

> Folder names should be standardized to either all have spaces between words or none. Personally I'm inclined towards spaces, since that's how the main app tree behaves.

Personally I'm not a fan of spaces in folder names in general because it means you have to put quotation marks around paths in the terminal

EricBAndrews:
I'm not sure breaking Lemmy and PieFed connectors into separate packages is worth it--it seems cleaner to me for all the connectors to live in the same package, since they're so similar in scope. A single `InstanceConnector` package may be a worthwhile layer of abstraction, though; we could feasibly implement the same for `BackendConnector`, and have all the `Client`s serve as clean app-facing APIs on top of their respective `Connector`s

> Personally I'm not a fan of spaces in folder names in general because it means you have to put quotation marks around paths in the terminal

Works for me

Sjmarf:
> I'm not sure breaking Lemmy and PieFed connectors into separate packages is worth it--it seems cleaner to me for all the connectors to live in the same package, since they're so similar in scope

Yeah, that would avoid the sub-repo issue too. I'm happy to go with that.

> A single InstanceConnector package may be a worthwhile layer of abstraction, though; we could feasibly implement the same for BackendConnector, and have all the Clients serve as clean app-facing APIs on top of their respective Connectors

Sounds good 👍

EricBAndrews (approved):
LGTM! 🔥

PR #2083, opened 2025-06-13
Progress towards decoupling InstanceConnection from Lemmy-specific Api Types

Sjmarf:
Some of the `InstanceConnection` methods returned Lemmy Api types. I've replaced some of these with new instance-agnostic types. There's still some work to do before it's completely decoupled - will finish those in a later PR.


Mlem/App/Utility/Extensions/Content Models/ModlogEntryContent+Extensions.swift
```diff
@@ -0,0 +1,165 @@
+//
+//  ModlogEntryContent+Extensions.swift
+//  Mlem
```
Sjmarf, line 3: This file hasn't changed at all - I renamed `ModlogEntryType` to `ModlogEntryContent`, and made a new type called `ModlogEntryType`. Git got confused over the rename.

EricBAndrews (approved):
LGTM! 🚀

PR #2052, opened 2025-05-21
Remove 0.18.0 support

Sjmarf:
> [!note]
> See also [this MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/20) and [this lemmy-swift-codegen PR](https://github.com/mlemgroup/lemmy-swift-codegen/pull/15)

Also bumped up to the latest 1.0 changes

Closes #2048

PR #2084, opened 2025-06-13
Fix login failing

Sjmarf:
As of #2052, logging in fails most of the time.

When you choose an instance in the login flow, Mlem makes an unauthenticated `api/v3/site` request, so that it can display the instance avatar. When you enter your details and login, Mlem makes a second `api/v3/site` request - this time, authenticated. Mlem does this so that it can get the username of the account, which it won't yet have if the user has put their email in the "username" field.

Swift's URLRequest caching system was caching the response of the first (unauthenticated) call, and returning it for the second (authenticated) call. It seems that Swift doesn't recognise the two reqeusts as being different despite having different headers. I've fixed this by disabling the URLRequest cache.

Before #2052, we attached the auth token as a field in the body of the request in addition to the header. Because of this, Swift was able to recognize the request as being different.

I've disabled the cache for all requests, not just the one related to login. Because the cache ignores headers, it might in theory return responses from the wrong account after you switch account.


EricBAndrews (approved):
LGTM! 👍

PR #2085, opened 2025-06-13
Finish decoupling LemmyConnection

Sjmarf:
Built ontop of #2083 -drafting until that is in

I've removed a few of the content model properties that we weren't using. All of those properties had schemas that will change significantly in Lemmy 1.0, so I don't think it's worth implementing proper support for them now when we aren't actually using them. If we need them later they can be re-implemented


EricBAndrews (approved):
Nice work! 💯

PR #2086, opened 2025-06-13
Move ApiClient context into LemmyConnection

Sjmarf:
Relies on #2085 - drafting until that's in

ApiClient has a cached SiteVersion:

```swift
try await api.version
```

In this PR, I've moved that caching layer down into `LemmyConnection` instead.


EricBAndrews (approved):
LGTM! 💯

PR #2087, opened 2025-06-13
Add `ApiClient.supports(_ feature:)` method

Sjmarf:
Built ontop of #2086, drafting until that's in.

Added a `supports(_ feature: Feature)` method to `ApiClient`, and removed `try await api.version`.

PR #2088, opened 2025-06-14
Keep track of instance software type

Sjmarf:
Built ontop of #2087, drafting until that's in

Added a new `SiteSoftwareType` enum:

```swift
enum SiteSoftwareType {
    case lemmy
}
```

Wherever we currently store an instance's `SiteVersion`, we're now also storing its `SiteSoftwareType`.

---------

The search tab now shows an instance's software name:

------

`SearchSheetView` has a feature that allows you to set a minimum required version for the results. This is used when filtering search results by instance in the "users" tab:

In this PR, you now specify a `Feature` that all instances in the results list must support, rather than a minimum version. This unfortunately makes the warning less specific, but that's something we could improve on in future.


Mlem/App/Views/Shared/Navigation/NavigationPage+View.swift
```diff
@@ -132,10 +132,10 @@ extension NavigationPage {
                         .padding(.vertical, 6)
                         .background(.themedSecondaryGroupedBackground, in: .rect(cornerRadius: Constants.main.standardSpacing))
                 }
-                .disabled(instance.version < (minimumVersion ?? .zero))
+                .disabled(requiredFeature.map { !instance.software.supports($0) } ?? false)
             } header: {
-                if let minimumVersion {
-                    Text("This feature is only supported for instances running version \(String(describing: minimumVersion)) or later.")
+                if let requiredFeature {
+                    Text("This feature is only supported on instances running certain software versions.")
```
EricBAndrews: Since we're losing precision, we might as well cut as many words as possible. I think the best way to handle this would be to have a (?) help tappable that opens a more detailed explanation, though that may be OOS here.

```suggestion
                    Text("This feature is not available on all instances.")
```

EricBAndrews (changes requested):

Mlem/App/Views/Shared/Navigation/NavigationPage+View.swift
Sjmarf: Makes sense 👍 Yeah, I'm gonna leave that OOS and implement something like that down the line

EricBAndrews (approved):
LGTM! 🚀

PR #2089, opened 2025-06-14
PieFedConnection boilerplate

Sjmarf:
Built ontop of #2088, drafting until that's in

Added boilerplate code for `PieFedConnection`. For now, every method just throws an error.

The first time a call is made to `ApiClient`, it now makes that call on both `LemmyConnection` and `PieFedConnection`. Only one will succeed (at the moment, `PieFedConnection` will never succeed because all methods throw an error). ApiClient remembers which connection succeeded, and uses _only_ that connection for all subsequent calls.


Mlem/App/Views/Shared/Navigation/NavigationPage+View.swift
```diff
@@ -135,7 +135,7 @@ extension NavigationPage {
                 .disabled(requiredFeature.map { !instance.software.supports($0) } ?? false)
             } header: {
                 if let requiredFeature {
-                    Text("This feature is not available on all instances.")
+                    Text("This feature is only supported on instances running certain software versions.")
```
EricBAndrews: Looks like update from the previous PR got missed here

```suggestion
                    Text("This feature is not available on all instances.")
```

Mlem/Localizable.xcstrings
```diff
@@ -8232,6 +8229,9 @@
           }
         }
       }
+    },
+    "This feature is only supported on instances running certain software versions." : {
```
EricBAndrews:
```suggestion
    "This feature is not available on all instances." : {
```

EricBAndrews (approved):
Nice work! Connection discovery is a nice way to handle this 👍

PR #2090, opened 2025-06-14
Person feed loader loading state fix

### LINKED ISSUE #1261: `PersonContentLoader` erroneously in loading state after loading first page if first page contained less items than the page limit

EricBAndrews:
## Issues

- closes #1261 

## Description

The `PersonContentStream` was not correctly detecting end-of-feed if the profile had a mismatched number of pages of posts and comments. On the last page of the less numerous type, the combined fetch would correctly return `.success` instead of `.done` because the more numerous type still had items to load. The less numerous type's stream, however, would never receive the 0-item input to `addItems` that signified end-of-feed, which caused it to remain stuck in the `.idle` state.

This fixes that problem by setting the `PersonContentStream` to consider itself done if the fetch returns fewer than the infinite load offset. This has a fringe case that fails loading too, but a much rarer one--I've left a comment in the code detailing that, though the fix will require 1.0's API.

This PR also updates `PersonContentGridView` to always infinite scroll. Currently, since no feed loader is provided to `EndOfFeedView`, it will simply spin infinitely at the bottom of the first page. The theory behind this change is that the option to disable  infinite scroll is a "doomscroll protection," which is not necessary in the contexts where `PersonContentGridView` is used.


Sjmarf (approved):
Nice, great work! 😄

PR #2092, opened 2025-06-15
Generalize `ShareInstancePickerView` sizing logic

Sjmarf:
The instance picker sheet that appears when you have your share preferences set to "ask every time" has some custom sizing logic that makes the sheet match the size of its content. In this PR I've generalized that logic into a view extension that we could reuse elsewhere if we needed to.


EricBAndrews (approved):
LGTM! 💯

PR #2093, opened 2025-06-18
Fix account avatar not immediately updating in the accounts list

### LINKED ISSUE #2079: Account Switcher Icon Remained Cached After Uploading New Photo
It took clearing the cache under the Developer menu, and restarting the app, but when uploading a new profile photo, the account switcher didn't update immediately. 

Also I'm not sure if it's related, but I got the following error when reloading the app after clearing the cache:

> Optional(A valid token was assigned to an ApiClient for the wrong account.) (Mlem/UserSession.swift, init(account:):55)

And 20 seconds later:

> Optional(Swift.CancellationError()) (Mlem/ContentView+Logic.swift, loadAvatar(url:):48)

Mlem 2.2 (385)

image

Sjmarf:
Drafting until #2088 is in

Closes #2079


EricBAndrews (approved):
LGTM! 🔥

PR #2095, opened 2025-06-19
Don't require space in spoiler opener

### LINKED ISSUE #2069: Multiple Spoilers in Community About Not Rendering
It seems that the use of multiple spoilers, at least in the About section of a community, does not render. 

image

Using Mlem 2.2 (365) and one example community is !youshouldknow@lemmy.world.

Sjmarf:
Before this PR, only `::: spoiler` was recognized as valid - `:::spoiler` was not. This has been corrected in this PR. 

Closes #2069


EricBAndrews (approved):
Nice 👍

PR #2096, opened 2025-06-20
PieFed support for `InstanceView`

Sjmarf:
Built on-top of #2089, drafting until that's in

> [!note]
> Relies on [this MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/22) 

`InstanceView` now renders correctly for PieFed instances, whereas previously it would give an error. Parts of `InstanceView` that are yet to be supported by PieFed are hidden.

PieFed instances don't appear in instance search - the easiest way to get to a PieFed `InstanceView` is to search for a community, then tap on the instance link in the ellipsis menu. Two test communities:
- !piefed_meta@piefed.social
- !feddit_online@feddit.online

The avatar for piefed.social doesn't load - this is because the icon is an SVG, which we don't support


EricBAndrews (approved):
LGTM! 🚀

PR #2097, opened 2025-06-21
Better rules list parsing

Sjmarf:
Improved the parsing logic that we use to show a rules list in the reporting sheet. This covers the case where the rules list is enclosed within a spoiler, but there is also a second node within the spoiler aside from a bulleted/numbered list. This happens on lemy.lol.


EricBAndrews (approved):
Nice! 👍

PR #2098, opened 2025-06-21
Expose removed content in `CommentPage` if navigated to from modlog

Sjmarf:
If you navigate to a `CommentPage` from the modlog, it'll now show the original comment content rather than "comment was removed". This applies to all removed comments in the thread. Removed comments are shown in red.


EricBAndrews (approved):
LGTM, super cool feature! 🔥

PR #2110, opened 2025-06-27
Fix always blurring images in post body

EricBAndrews:
Fixes a bug introduced in #2098 that caused all images inside a post to be blurred.


Sjmarf (approved):
Whoops, nice catch. LGTM 😄

Issue #2104, opened 2025-06-25
PieFed: Inbox


Sjmarf:
There's *some* API support for this, but implementing it would be a pain. Issues to improve the API:

https://codeberg.org/rimu/pyfedi/issues/905
https://codeberg.org/rimu/pyfedi/issues/922

PR #2112, opened 2025-06-30
Log in to PieFed

Sjmarf:
Added the ability to log in to PieFed, and implemented most of the logged-in actions (vote, save, etc).

Still missing:
- Inbox (#2104)
- Moderation tools (#2107)
- Image uploading (#2105)
- Communities show the local subscriber count, not the total subscriber count. This is due to an API limitation; it looks like it may be fixed shortly


Mlem/App/Models/Session/UserSession.swift
```diff
@@ -23,6 +23,8 @@ class UserSession: Session {
     /// This **only** includes requests made by calling `toggleInstanceBlock` on this `UserSession`.
     private(set) var ongoingInstanceBlockRequests: Set<ActorIdentifier> = []
     private(set) var visitHistory: VisitHistory?
+    
+    private(set) var subscriptionListErrorDetails: ErrorDetails?
```
Sjmarf, line 27: I added an `ErrorView` to the subscription list so that we could avoid showing the error toast in this context. I implemented this because the PieFed subscription list endpoint was broken at the time. They've since fixed that endpoint so this change isn't really related to PieFed anymore, but it's nice to have anyway

Mlem/App/Views/Shared/Bubble Picker/ChildSizeReader.swift
```diff
@@ -33,7 +33,7 @@ struct ChildSizeReader<Content: View>: View {
             if index < sizes.count {
                 sizes[index] = preferences
             } else {
-                assertionFailure()
+//                assertionFailure()
```
EricBAndrews:
```suggestion
                assertionFailure()
```

Mlem/Packages/Rest/Sources/RestClient/RestClient.swift
```diff
@@ -22,7 +22,7 @@ public class RestClient<ErrorType: Decodable & CustomStringConvertible> {
     ) async throws(RestError) -> Request.Response {
         let urlRequest = try urlRequest(baseUrl: baseUrl, request: request, token: token)
         // this line intentionally left commented for convenient future debugging
-//        urlRequest.debug()
+        urlRequest.debug()
```
EricBAndrews:
```suggestion
//        urlRequest.debug()
```

EricBAndrews (approved):
Nice work! 🔥 

Saved posts don't load correctly--this appears to be due to an API discrepancy between PieFed and Lemmy. I think this sort of thing would best be addressed at the `ApiRepository` level after the architecture discussions in Slack.

Issue #2111, opened 2025-06-27
StateManager race condition


EricBAndrews:
### Requirements

- [x] There are no existing issues for this bug report.

### Description

The `StateManager` does not always behave as expected. The easiest way to reproduce this is to scroll the feed, open each post, and navigate back to feed; sometimes the post will not be shown as read in the feed.

I _believe_ this is caused by the upgrade operation's snapshot being fetched before the mark read operation completes, but the post itself not being updated with that snapshot until after; this causes a stale value to appear in the `StateManager`.

Sometimes upvote on save also triggers similar behavior--the upvote indicator will flash on, but be reset when the saved indicator goes on.

### Mlem Version

Latest dev

EricBAndrews:
Unified model architecture resolves this and is now implemented on all the places this behavior showed up

PR #2114, opened 2025-07-03
Add `ApiRepository` Layer

EricBAndrews:
## Issues

- progress towards #2111 

## Description

This PR extracts all direct `Connection` and `RestClient` calls out of the `ApiClient` and into the `ApiRepository`. The `ApiRepository` is internal to `MlemMiddleware` and is responsible for handling the `Connection` logic; its methods return `Snapshot`s, which the `ApiClient` then consumes and processes into frontend-ready models.

This PR does not alter the `ApiClient`'s app-facing API (except for removing `getReportCount`, which was returning a `Snapshot`; the call to `ApiClient.getReportCount` has been replaced with a direct repository call). This means that there is currently a lot of boilerplate passthrough code, which will be removed as responsibility for state tracking is migrated to the new system.

### DIFF
--- a/Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/ApiClient+Inbox.swift
+++ b/Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/ApiClient+Inbox.swift
@@ -14,13 +14,13 @@ public extension ApiClient {
         limit: Int,
         unreadOnly: Bool = false
     ) async throws -> [Reply2] {
-        let snapshot = try await repository.getReplies(
-                sort: sort,
-                page: page,
-                limit: limit,
-                unreadOnly: unreadOnly
-            )
-        return await caches.reply2.getModels(api: self, from: snapshot)
+        let snapshots = try await repository.getReplies(
+            sort: sort,
+            page: page,
+            limit: limit,
+            unreadOnly: unreadOnly
+        )
+        return await caches.reply2.getModels(api: self, from: snapshots)
     }
     
     func getMentions(
@@ -29,13 +29,13 @@ public extension ApiClient {
         limit: Int,
         unreadOnly: Bool = false
     ) async throws -> [Reply2] {
-        let snapshot = try await repository.getMentions(
-                sort: sort,
-                page: page,
-                limit: limit,
-                unreadOnly: unreadOnly
-            )
-        return await caches.reply2.getModels(api: self, from: snapshot)
+        let snapshots = try await repository.getMentions(
+            sort: sort,
+            page: page,
+            limit: limit,
+            unreadOnly: unreadOnly
+        )
+        return await caches.reply2.getModels(api: self, from: snapshots)
     }
     
     func getMessages(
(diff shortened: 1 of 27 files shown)


Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Repository/ApiRepository.swift
```diff
@@ -0,0 +1,148 @@
...
+    public let username: String?
+    
+    private var ongoingConnectionDiscoveryTask: Task<Any, Error>?
+    var connection: (any InstanceConnection)?
+    
+    var restClient: RestClient<ApiErrorResponse> = .init()
+    
+    public internal(set) var token: String?
+    
+    public init(baseUrl: URL, username: String? = nil) {
```
Sjmarf: These should not be `public`, as the parent class is `internal` (not sure why Xcode doesn't complain about this)

```suggestion
    internal var token: String?
    
    init(baseUrl: URL, username: String? = nil) {
```

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Repository/ApiRepository.swift
```diff
@@ -0,0 +1,148 @@
...
+
+import Rest
+import Foundation
+
+/// This class represents an abstract interface on top of the underlying Connection; it is responsible for managing the Connection and
+/// serving consistent Snapshot models to higher layers.
+///
+/// The data access methods in here are all intentionally dumb--they just make calls and return snapshots. Validation, enrichment,
+/// and other business logic that must occur before serving the final model to the app should be performed by the consumer of the repository.
+class ApiRepository {
```
Sjmarf: (Optional)

Personally I prefer to explicitly mark internal stuff as `internal`, to distinguish between "I intend for this to be internal" and "I forgot to set an access level"

```suggestion
internal class ApiRepository {
```

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/ApiClient+Inbox.swift
```diff
@@ -104,19 +92,14 @@ public extension ApiClient {
...
             message.readManager.updateWithReceivedValue(read, semaphore: semaphore)
         }
     }
     
     func getPersonalUnreadCount() async throws -> PersonalUnreadCountSnapshot {
-        let response = try await performingForConnection { connection in
-            try await connection.getPersonalUnreadCount()
-        }
-        return response
+        try await repository.getPersonalUnreadCount()
```
Sjmarf: This function should be removed, for the same reason that we're removing `getReportCount`

Sjmarf (changes requested):
Nice work! A few small comments

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/ApiClient+Inbox.swift
EricBAndrews: Nice catch!

Sjmarf (approved):
Looks good! 🚀

PR #2116, opened 2025-07-03
Fix `assertionFailure` in `BubblePicker`

Sjmarf:
The `assertionFailure` would happen if the bubble picker appeared with `n` tabs, and was later updated to have more than `n` tabs. 

This is because the value of `sizes` was initialized when the `BubblePicker` appeared, but was never changed if the value of `tabs` was changed.

`BubblePicker` now only stores the size of the currently selected tab, rather than all tabs. This fixes the issue


EricBAndrews (approved):
LGTM! 🔥

PR #2126, opened 2025-07-05
Remap PieFed "active" sort to "New Comments"

Sjmarf:
The "Active" sort mode on PieFed isn't the same as the "Active" sort on Lemmy. Instead, it just sorts by newest comment date. I think it's best to rename it to "New Comments" in Mlem, which is what I've done here.


EricBAndrews (approved):
Nice 👍

PR #2140, opened 2025-07-12
Support threadiverse.link

### LINKED ISSUE #2136: Support threadiverse.link links
We already support Lemmyverse.link links. This is presumably a clone that also supports PieFed, or something like that

Sjmarf:
Closes #2136


EricBAndrews (approved):
Nice 💯

PR #2153, opened 2025-07-17
PieFed: Mark posts as read

### LINKED ISSUE #2117: PieFed: mark posts as read
When I wrote #2112 the mark-read endpoint wasn't implemented yet. It's not been implemented, so we can add it in

Sjmarf:
Closes #2117. Only available on version 1.0.1 and later, but doesn't throw any user-visible errors on 1.0.0 so I haven't implemented a version check


EricBAndrews (approved):
LGTM! 🔥

PR #2155, opened 2025-07-18
Post UpdateQueue

EricBAndrews:
## Issues

- progress towards #2111 

## Description

This PR replaces the existing `StateManager` state management solution with a new queue-based snapshot architecture.

## Implementation Notes

This architecture ensures that network race conditions do not occur and that model state remains responsive by satisfying the following properties:
1. For a given model, only one mutating request (i.e., a request returning an updated server state that the model should be updated to reflect) is sent at a time
2. The model is only updated with the server-side value when no further mutating requests are queued

Property 1 ensures that each request returns a server state guaranteed to be newer than the previously received state. This avoids network race conditions (e.g., when sending both upvote and save in quick succession without queueing, the server may process the upvote first but return it more slowly; in this scenario, the upvote response would not reflect the saved status but would be treated as newer).

Property 2 ensures that local state faking is never overridden before the associated server call is executed. 

The core work takes place in `PostUpdateQueue`. Requests are added to the queue and executed serially. The `PostUpdateQueue` tracks a `lastVerifiedSnapshot`, which reflects the most up-to-date server state; when each request returns, if it succeeded, `lastVerifiedSnapshot` is updated with its returned value. Once the last request in the queue is executed, the associated post model is updated with `lastVerifiedSnapshot`. The basic operation of the queue is straightforward, but there are a few technical details worth noting:

### Snapshot tiers and merging

Not all tasks are guaranteed to return snapshots of the same tier. When a task returns a higher tier snapshot than `lastVerifiedSnapshot`, the correct behavior is obviously to upgrade `lastVerifiedSnapshot` to that higher tier by means of a simple assignment. When a task returns a snapshot of a lower tier, however, the correct behavior is subtler: we should retain the higher-tier information, but update the lower. This is achieved by `PostSnapshotProviding.merge`, which takes in another `PostSnapshotProviding` and returns a synthetic snapshot in the highest available tier containing the most recent data at each tier. This function assumes that the snapshot it is called on is more recent than the snapshot being merged in.

### Snapshot-creating vs. snapshot-modifying tasks

Not all `ApiRepository` calls return snapshots. For operations such as mark read, a success response indicates that the server state has updated but does not offer a convenient snapshot with which to update `lastVerifiedSnapshot`. To handle these cases, the `PostUpdateQueue` can take in a task that accepts a `PostSnapshotProviding` parameter which is then mutated to reflect the implicitly updated server state and returned. When these tasks are called, they are provided `lastVerifiedSnapshot`.

To ensure that there is always a snapshot to update, post models now offer a `takeSnapshot` method that creates a snapshot. When the queue's parent is assigned for the first time, it sets `lastVerifiedSnapshot` to `parent.takeSnapshot()`; when the queue's parent is upgraded to a higher tier, `lastVerifiedSnapshot` is merged with `parent.takeSnapshot()`.


Sjmarf:
I wonder whether there would be any merit to each model having a single stored property: the most recently updated snapshot. Then, each computed property in Post1Providing could access the corresponding property of the stored snapshot

This would remove a lot of repetitive code, but might have performance implications. What do you think?

EricBAndrews:
That would make state faking awkward--we'd have to either store every fakable property independently or have a "clean" and a "dirty" snapshot. I don't love the number of field copying functions we have right now, but I think it's on balance a pretty small price to pay for an otherwise clean architecture.

Sjmarf:
Ok, makes sense 👍

Sjmarf:
> Not all ApiRepository calls return snapshots. For operations such as mark read, a success response indicates that the server state has updated but does not offer a convenient snapshot with which to update lastVerifiedSnapshot.

Conveniently for us, this request could return a Snapshot on Lemmy 1.0. They seem to be phasing out SuccessResponse entirely, so once we drop support for earlier versions things should get simpler

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/ApiClient+Post.swift
```diff
@@ -197,49 +176,7 @@ public extension ApiClient {
...
-    
-    @discardableResult
-    func voteOnPost(id: Int, score: ScoringOperation, semaphore: UInt? = nil) async throws -> Post2 {
-        let snapshot = try await repository.voteOnPost(id: id, score: score)
-        return await caches.post2.getModel(
-            api: self,
-            from: snapshot,
-            semaphore: semaphore
-        )
-    }
```
Sjmarf, line 209: Removing these cuts off our ability to do certain operations from Mlem's side. We would no longer be able to operate directly on a post using only its ID - we'd have to fetch the full model first. How big of a deal is that?

Of course, we could reach back into the ApiRepository layer if we needed to, but I'd really rather avoid doing that on Mlem's side, at least.

I'm okay with removing these if you'd prefer it that way - the situation I mention above hasn't happened yet, so arguably keeping these methods around is just incurring unnecessary maintenance cost.

Mlem/App/Utility/Extensions/Content Models/DeletableProviding+Extensions.swift
```diff
@@ -10,23 +10,18 @@ import MlemMiddleware
...
-                            callback: { self.updateDeleted(false) },
+                            callback: { self.updateDeleted(false, callback: nil) },
                             color: .themedNegative
                         )
                     )
-                case .failed:
+                } else {
                     ToastModel.main.add(.failure("Failed to delete post!"))
-                default:
-                    handleError(MlemError.unexpectedValue, silent: true)
```
Sjmarf: I don't like the wording of `success` here - it doesn't read naturally as a boolean. Something like `didSucceed` would be better in that regard, but that feels weird to me... maybe going back to an enum would be better?

```swift
switch result {
    case .success:
        // ...
    case .failure:
       // ...
}
```

That would also let us pass through the error, if we ever wanted that:

```swift
switch result {
    case .success:
        // ...
    case let .failure(error):
       // ...
}
```

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Comment/Comment2Providing.swift
```diff
@@ -48,16 +48,16 @@ public extension Comment2Providing {
...
-    func updateVote(_ newValue: ScoringOperation) -> Task<StateUpdateResult, Never> {
-        votesManager.performRequest(expectedResult: votes.applyScoringOperation(operation: newValue)) { semaphore in
+    func updateVote(_ newValue: ScoringOperation) throws {
+        // TODO: UpdateQueue use queued state management
+        _ = votesManager.performRequest(expectedResult: votes.applyScoringOperation(operation: newValue)) { semaphore in
             try await self.api.voteOnComment(id: self.id, score: newValue, semaphore: semaphore)
         }
     }
     
-    @discardableResult
```
Sjmarf, line 58: Does this need to be throwing? Ditto for the other functions

Mlem/App/Utility/Extensions/Content Models/Post1Providing+Extensions.swift
```diff
@@ -44,45 +48,67 @@ extension Post1Providing {
...
+                        feedback: feedback
+                    )
+                }
+            }
+        } catch {
+            handleError(error)
         }
     }
     
     func togglePinnedCommunity(feedback: Set<FeedbackType>) {
```
Sjmarf, line 72: I tried pinning a post on PieFed. Doing so is expected to produce an error, because #2166 isn't merged yet. Instead, I see the loading symbol spin indefinitely, and it spams the haptic feedback. Logs:

```
DEBUG found next task
This instance doesn't support that operation.
DEBUG found next task
This instance doesn't support that operation.
DEBUG found next task
This instance doesn't support that operation.
DEBUG found next task
This instance doesn't support that operation.
DEBUG found next task
This instance doesn't support that operation.
DEBUG found next task
```

Sjmarf (changes requested):
Nice! This is a super clean architecture 😄 A few points:

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/ApiClient+Post.swift
EricBAndrews: I'm not opposed to this style of headless model-mutating call, though it would need to be rewritten to first check the cache for the model, then either call the model's vote method or submit a repository call directly. It's pretty trivial to re-implement, so let's leave them out for now and circle back if we need them down the line.

Mlem/App/Utility/Extensions/Content Models/DeletableProviding+Extensions.swift
EricBAndrews: Enum with associated error works for me 👍

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Comment/Comment2Providing.swift
EricBAndrews: Nope, nice catch!

Mlem/App/Utility/Extensions/Content Models/Post1Providing+Extensions.swift
EricBAndrews: Whoops, never made failed items dequeue 🙃

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/ApiClient+Post.swift
Sjmarf: Sounds good 👍

Sjmarf (approved):
LGTM 🚀

PR #2174, opened 2025-07-27
Improve mark read on scroll reliability

### LINKED ISSUE #2169: Mark read on scroll is super inconsistent

EricBAndrews:
## Issues

- closes #2169 

## Description

Root cause was the incredibly tight `.onGeometryChange` threshold. There was so little margin between the post crossing the threshold and it being derendered altogether that posts would sometimes drop even when scrolling very slowly.

This fixes that for all sane use cases by increasing the threshold to 90px. This causes the post to be marked read as it's sliding away under the header bar--the state change is still hidden from the user but the added space significantly improves detection reliability. Posts can still drop when scrolling extremely fast, but I believe that behavior is far enough outside of the expected usage profile to not warrant a more sophisticated fix.


Sjmarf (approved):
Nice, works perfectly! 😄

PR #2175, opened 2025-07-28
Warnings Cleanup

EricBAndrews:
Cleans up simple warnings--things like non-mutated `var`s, unused `let` bindings, redundant optional unwrappings. Thread safety warnings and the deprecated `HapticManager.main` calls are omitted from the scope of this PR.


Sjmarf (approved):
Nice 👍

Issue #2016, opened 2025-05-01
Menu icons are incorrectly colored


Sjmarf:
![image](https://github.com/user-attachments/assets/b465cdac-59ed-404d-8f2b-07c49fcde799)

Sjmarf:
Possibly caused by a minor iOS update? I reckon this is due to the `.tint` we applied at the top level, but I don't remember menu items being tintable before

Sjmarf:
Looks like it happens in all context menus

![image](https://github.com/user-attachments/assets/e3595717-4448-42cd-a4cc-9ca77597baca)

Sjmarf:
Huh, I just made a new build and it no longer has this issue 🤔 Still present on the TF for me

EricBAndrews:
Yeah, I can't reproduce this locally either. I've tried building release product as well, but the tint works as expected. Not entirely sure how to go about fixing this issue.

Sjmarf:
Do you also see this issue on the weekly TF? Still on the TF for me, but not locally. I wonder whether we need to change the version of Xcode used to build the TFs or something

EricBAndrews:
I can reproduce this locally with XCode 16.4. Root cause does appear to be that `.tint`--using the deprecated `.accentColor` produces the intended behavior.

EricBAndrews:
Removing this from 2.2 because we have a band-aid in place, but leaving open until we have a long term solution.

PR #2185, opened 2025-08-03
Short term tint fix

EricBAndrews:
## Issues

- progress towards #2016 

## Description

Not sure if this is a bug in SwiftUI or the intended behavior, but `.tint` colors the menu buttons when building with XCode 16.4. I don't love using the deprecated API here but it does work.

### DIFF
--- a/Mlem/App/Views/Root/ContentView.swift
+++ b/Mlem/App/Views/Root/ContentView.swift
@@ -65,7 +65,7 @@ struct ContentView: View {
                     shareInfo: .init(get: { navigationModel.shareInfo }, set: { navigationModel.shareInfo = $0 }),
                     contentPickerTracker: navigationModel.contentPickerTracker
                 )
-                .tint(.themedAccent)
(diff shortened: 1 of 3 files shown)


Sjmarf:
There are a handful of places where this doesn't apply properly, particularly inside sheets.




Maybe the `.accentColor` doesn't propagate to sheets properly or something - we could try applying `.accentColor` at the top level of each sheet, in addition to the `ContentView`. Failing that, we could individually fix those cases by applying a `.tint`.

Alternatively, we could continue to use `.tint` at the top level, and apply `.tint(.primary)` to all uses of `.contextMenu`. It might be possible to do this in an overload of `.contextMenu`.

Also worth noting that we've got another `.tint(.themedAccent)` inside our custom `Form` definition. Might be worth checking whether that causes any issues if we end up using `.accentColor`.

Sjmarf (approved):
LGTM 💯

PR #2188, opened 2025-08-04
Fix tapping on notification on PieFed

### LINKED ISSUE #2187: Tapping on notification doesn't work on PieFed
Reported here https://lemmy.ml/post/33702199/20219056

There's a chance this is PieFed's problem. Either way, we should never show an infinite loading screen - if we get an error, we should show an `ErrorView`.

Sjmarf:
Closes #2187


EricBAndrews (approved):
LGTM! 💯

PR #2193, opened 2025-08-10
Basic 1.0 Compat

Sjmarf:
Drafting until 2.2 is out

I've got a local Lemmy 1.0 instance set up. This is the initial compatibility work - you can now log in to Lemmy 1.0, but there's still plenty of errors that will need fixing.

The main change here is that `LemmyConnection` will now attempt to make requests on both the `v3` and `v4` endpoints. It remembers which one succeeds, and uses that endpoint for all subsequent calls. 

This is very similar behavior to the logic we use in `ApiRepository` to decide whether to use Lemmy- or Piefed-format requests. I've moved that logic into a `ConnectionMultiplexer` class, which is reused for the new endpoint multiplexing inside of `LemmyConnection`.


Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Sort Types/SortTimeRange.swift
```diff
@@ -30,7 +30,7 @@ public enum SortTimeRange: Hashable, Sendable {
     // This should only be used internally within ApiClient
     var timeRangeSeconds: Int {
         switch self {
-        case .allTime: .max
+        case .allTime: 2_147_483_647 // This is the upper bound of an i32 in Rust; going higher than this causes an error
```
EricBAndrews: *nit*

We might want to define this reusably as something like `.rustI32Max`, since this behavior might crop up elsewhere

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/InstanceConnection/LemmyConnection/LemmyConnection+General.swift
```diff
@@ -22,6 +22,8 @@ public extension LemmyConnection {
             )
         }
         
+        print(response)
```
EricBAndrews: *nit*

Dead code?

Mlem/App/Views/Shared/CustomTabBarController.swift
```diff
@@ -72,6 +72,7 @@ class CustomTabBarController: UITabBarController, UITabBarControllerDelegate {
     }
     
     @objc func swipeGestureTriggered(_ recognizer: UISwipeGestureRecognizer) {
+        print("DEBUG swipe gesture triggered")
```
EricBAndrews: *nit* Dead code?

EricBAndrews (approved):
LGTM! 🔥

PR #2200, opened 2025-08-11
Show "short description" field in `InstanceView`

Sjmarf:
Instances have a "description" and a "short description". Previously we only showed the description; now we show both.

An example of one that has both is lemmy.world. An example of an instance that only has a short description is piefed.zip. Previously, nothing would be shown in piefed.zip's about page.


EricBAndrews (approved):
LGTM! 🔥

PR #2208, opened 2025-08-15
New Search Tab

### LINKED ISSUE #2082: Make "Saved" easier to get to
Currently there's a dedicated feed for Saved. This makes it impossible to check your "Saved" list without losing your position in the feed. There are different ways we could fix this:

1) Remember the state of the feed when you switch feed using the drop-down in the navigation bar title.
2) Move the Saved feed to the profile, to match other clients (I personally don't like this solution)
3) Move it to the Search tab, which is what Apple News does

Another consideration is that at some point we'll need to add:
- Upvoted and downvoted lists
- History

Those should probably go in the same place as saved

Sjmarf:
Closes #2082

There's some dead code in here for the grid UI. That'll get used once topics are implemented

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Feeds/SavedFeedView.swift
+++ b/Mlem/App/Views/Root/Tabs/Feeds/SavedFeedView.swift
@@ -1,5 +1,5 @@
 //
-//  FeedsView.swift
+//  SavedFeedView.swift
 //  Mlem
 //
 //  Created by Eric Andrews on 2024-01-07.
@@ -32,10 +32,7 @@ struct SavedFeedView: View {
     init(feedSelection: FeedSelection? = nil) {
         // need to grab some stuff from app storage to initialize with
         @Setting(\.behavior_internetSpeed) var internetSpeed
-        @Setting(\.feed_showRead) var showReadPosts
-        @Setting(\.post_defaultSort) var defaultSort
         @Setting(\.post_size) var postSize
-        @Setting(\.feed_default) var defaultFeed
         
         if let firstUser = AppState.main.firstAccount as? UserAccount {
             _savedFeedLoader = .init(wrappedValue: .init(
@@ -56,9 +53,6 @@ struct SavedFeedView: View {
             .scrollContentBackground(.hidden)
             .conditionalNavigationTitle("Saved")
             .navigationBarTitleDisplayMode(.inline)
-            .onChange(of: showRead) {
-                scrollToTopTrigger.toggle()
-            }
             .outdatedFeedPopup(feedLoader: savedFeedLoader)
             .environment(\.feedContext, .saved)
     }
--- a/Mlem/App/Views/Root/Tabs/Feeds/SavedFeedView.swift
+++ b/Mlem/App/Views/Root/Tabs/Feeds/SavedFeedView.swift
@@ -16,15 +16,6 @@ struct SavedFeedView: View {
     @Environment(FiltersTracker.self) var filtersTracker
     @Environment(BackendClient.self) var backendClient
     
-    @Setting(\.post_size) var postSize
-    @Setting(\.feed_showRead) var showRead
-    @Setting(\.tip_feedWelcomePrompt) var showWelcomePrompt
-    @Setting(\.behavior_internetSpeed) var internetSpeed
-    @Setting(\.links_embedLoops) var embedLoops
-    
-    @AppStorage("lastTestFlightUpdate") var lastTestFlightUpdate: URL?
-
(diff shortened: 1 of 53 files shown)


Mlem/App/Views/Root/Tabs/Feeds/SavedFeedView.swift
```diff
@@ -0,0 +1,144 @@
+//
+//  FeedsView.swift
```
EricBAndrews:
```suggestion
//  SavedFeedView.swift
```

Mlem/App/Views/Root/Tabs/Feeds/SavedFeedView.swift
```diff
@@ -0,0 +1,144 @@
...
+                        feedSelection = feed
+                    }
+                } else {
+                    feedSelection = feed
+                }
+            }
+            .symbolVariant(feedSelection == feed ? .fill : .none)
+        }
+    }
+}
```
EricBAndrews: Unused

Mlem/App/Views/Root/Tabs/Feeds/SavedFeedView.swift
```diff
@@ -0,0 +1,144 @@
...
+    var content: some View {
+        FancyScrollView(scrollToTopTrigger: $scrollToTopTrigger) {
+            if AccountsTracker.main.isEmpty, showWelcomePrompt, !appState.firstApi.willSendToken {
+                FeedWelcomeView()
+                    .padding([.horizontal, .bottom], Constants.main.standardSpacing)
+            }
+            if Bundle.main.isTestFlight, let testflightUrl = backendClient.testflightUpdate, lastTestFlightUpdate != testflightUrl {
+                UpdateBannerView(url: testflightUrl)
+                    .padding([.horizontal, .bottom], Constants.main.standardSpacing)
+            }
```
EricBAndrews: IMO these should only be shown on the main feeds view, since that's the start page of the app

Mlem/App/Views/Root/Tabs/Feeds/SavedFeedView.swift
```diff
@@ -0,0 +1,144 @@
...
+struct SavedFeedView: View {
+    @Environment(AppState.self) var appState
+    @Environment(FiltersTracker.self) var filtersTracker
+    @Environment(BackendClient.self) var backendClient
+    
+    @Setting(\.post_size) var postSize
+    @Setting(\.feed_showRead) var showRead
+    @Setting(\.tip_feedWelcomePrompt) var showWelcomePrompt
+    @Setting(\.behavior_internetSpeed) var internetSpeed
+    @Setting(\.links_embedLoops) var embedLoops
```
EricBAndrews: Most of these are unused (all of them if we remove welcome prompt from below)

Mlem/App/Views/Root/Tabs/Feeds/SavedFeedView.swift
```diff
@@ -0,0 +1,144 @@
...
+    
+    @State var scrollToTopTrigger: Bool = false
+    
+    init(feedSelection: FeedSelection? = nil) {
+        // need to grab some stuff from app storage to initialize with
+        @Setting(\.behavior_internetSpeed) var internetSpeed
+        @Setting(\.feed_showRead) var showReadPosts
+        @Setting(\.post_defaultSort) var defaultSort
+        @Setting(\.post_size) var postSize
+        @Setting(\.feed_default) var defaultFeed
```
EricBAndrews:
```suggestion
        @Setting(\.behavior_internetSpeed) var internetSpeed
        @Setting(\.post_size) var postSize
```

Mlem/App/Views/Root/Tabs/Feeds/SavedFeedView.swift
```diff
@@ -0,0 +1,144 @@
...
+    var body: some View {
+        content
+            .background(ThemedColor.themedGroupedBackground)
+            .themedGroupedBackground()
+            .scrollContentBackground(.hidden)
+            .conditionalNavigationTitle("Saved")
+            .navigationBarTitleDisplayMode(.inline)
+            .onChange(of: showRead) {
+                scrollToTopTrigger.toggle()
+            }
```
EricBAndrews: Not relevant to saved
```suggestion
```

EricBAndrews (changes requested):
Nice work! The new design is very clean 😄 

There's a bunch of code in `SavedFeedView` that isn't needed outside of `FeedsView`--see comments.

Sjmarf:
Nice catch, thanks. I agree we should remove the welcome banner and TF banner 👍

Issue #2205, opened 2025-08-14
iOS 26: Search sheets missing search bar


Sjmarf:
This applies to the community search sheet inside of the post composer, the discussion language picker, etc. Anywhere where we use the custom `SearchBar` class inside of a sheet.

PR #2211, opened 2025-08-15
Update Mlem version to 2.3, minimum iOS to 18

EricBAndrews:


Sjmarf (approved):
Nice 👍

.github/workflows/ci_build_16.yml
```diff
@@ -24,6 +24,6 @@ jobs:
     - name: Build
       uses: "./.github/actions/ci_xcodebuild"
       with:
-        xcode_version: "16.0"
-        xcodebuild_destination: "platform=iOS Simulator,name=iPhone 16,OS=18.1"
+        xcode_version: "16.4"
```
Sjmarf, line 27: Are we able to build with Xcode 26? That would allow beta testers who are using the iOS 26 beta to test the new changes

Sjmarf: I assume Xcode 26 is required, anyway. Correct me if I'm wrong

EricBAndrews: This is just the CI build--the actual beta build that goes out to TF is done in XCode cloud. Either way, I'd like to hold off on releasing the 26 version to TF until we've sanded off the rough edges (notably the account switcher ux).

Sjmarf: Makes sense, thanks 👍

Sjmarf: I agree we should hold off on the beta until we've got an account switcher

Sjmarf: #2205 also needs to be fixed before first TestFlight

PR #2218, opened 2025-08-19
Fix removing content crashing app

### LINKED ISSUE #2217: [Bug] App crashes when deleting comment in moderator mode
# Description 

As a moderator
When I try to delete a comment without explaination message 
The app crashes 

# Versions

- iOS 18.6
- iPhone 14 Pro
- app in French
- device in French
- Mlem version 2.2 (463)

# Note

If the issue occurs again I will try ton record screen and bring more details

Sjmarf:
Closes #2217. It does send the removal request correctly, it just crashes afterward.

---------

This was a `@MainActor` issue. I've fixed the issue in this specific case. What do you think about changing MlemMiddleware to run the callback on the main actor? That would be a more general fix that would avoid similar issues cropping up in future.

### DIFF
--- a/Mlem/App/Views/Pages/Editors/ContentRemovalEditorView.swift
+++ b/Mlem/App/Views/Pages/Editors/ContentRemovalEditorView.swift
@@ -71,12 +71,14 @@ struct ContentRemovalEditorView: View {
     
     func send() {
         target.toggleRemoved(reason: reason) { status in
-            switch status {
-            case .success:
-                hapticManager.play(haptic: .success, tier: .low)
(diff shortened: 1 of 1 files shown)


EricBAndrews (approved):
LGTM! 🔥 

Re: running callbacks on `MainActor`--I'd prefer to avoid that if possible, since that would make it easy to unintentionally write heavy code that runs unnecessarily on `MainActor`. Do we know which line was causing the `MainActor` crash?

Sjmarf:
It was `dismiss`.

PR #2221, opened 2025-08-20
Bubble picker fix

### LINKED ISSUE #2143: Bubble picker sometimes broken
From TF feedback. Can't repro

Image

Sjmarf:
Closes #2143

I was hoping for a better fix than this tbh. This might break if we start reordering tabs after creating the bubble picker or something like that. But it's good enough for now and I think it works properly everywhere.


EricBAndrews (approved):
LGTM, if it works it works 🤷

PR #2222, opened 2025-08-21
Comment Update Queue

EricBAndrews:
Implements the update queue for comments. This one has a chunk of shim code in because of how comments and replies are intertwined; I'll do replies next and clean that out.


Sjmarf:
> This one has a chunk of shim code in because of how comments and replies are intertwined; I'll do replies next and clean that out.

In the future we won't have a Reply type - I think it's possible to instead have a Notification type which stores an instance of Comment2. iirc this wasn't possible before we dropped support for 0.19.2. It may not be worth adding an update queue for replies at this stage.

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/Caching/Caches/ReplyCaches.swift
```diff
@@ -33,17 +33,7 @@ class Reply2Cache: ApiTypeBackedCache<Reply2, Reply2Snapshot> {
...
-        let savedManager: StateManager<Bool>
-        
-        if let reply = api.caches.comment2.retrieveModel(cacheId: snapshot.comment.id) {
-            votesManager = reply.votesManager
-            savedManager = reply.savedManager
-        } else {
-            votesManager = .init(wrappedValue: snapshot.votes)
-            savedManager = .init(wrappedValue: snapshot.saved)
-        }
-
```
Sjmarf, line 46: Reply and comment vote/save status no longer syncs correctly. Repro:

- Go to inbox
- Tap on a comment to open it in a new page
- Upvote the comment on that new page
- Go back to the inbox
- Reply doesn't show as upvoted

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Comment/Comment1Providing+Snapshots.swift
```diff
@@ -0,0 +1,49 @@
...
+        } else if let comment1Snapshot = snapshot as? Comment1Snapshot {
+            await snapshotUpdate(with: comment1Snapshot)
+        } else {
+            assertionFailure("Unrecognized comment snapshot")
+        }
+    }
+    
+    @MainActor
+    internal func snapshot1Update(with snapshot: Comment1Snapshot) {
+        comment1.setIfChanged(\.content, snapshot.content)
```
Sjmarf: I suspect we may need to check for `snapshot.removed` here. See this code which you removed from `Comment+CacheExtensions.swift`:

```swift
  // If the comment is removed, the API returns an empty string
        // for the `comment/list` endpoint, but returns the comment content
        // in the modlog endpoint. This `if` statement prevents the comment
        // content being overwritten with that empty string.
        if !snapshot.removed {
            setIfChanged(\.content, snapshot.content)
        }
```

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/UpdateQueues/CommentUpdateQueue.swift
```diff
@@ -0,0 +1,129 @@
...
+    
+    /// Add a task to the queue for a repository call that **does not** return a complete snapshot. The queue will provide the latest verified
+    /// snapshot to the task, which should then modify and return the snapshot according to the repository call result.
+    /// - Note: **only** use this method when absolutely necessary; if the repository returns a complete snapshot, use the variant above.
+    internal func addItem(item: @escaping (any CommentSnapshotProviding) async throws -> any CommentSnapshotProviding) {
+        addItem(.modifiesSnapshot(item))
+    }
+    
+    /// Queues the given upgrade operation for execution
+    /// - Returns: post returned by the upgrade operation
```
Sjmarf:
```suggestion
    /// - Returns: comment returned by the upgrade operation
```

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/UpdateQueues/CommentUpdateQueue.swift
```diff
@@ -0,0 +1,129 @@
...
+        // this method is a unique case because the context it is called from needs to receive its result. This method therefore waits
+        // for any currently queued actions to finish, then blocks the queue from restarting until the upgrade is complete.
+        await semaphore.wait()
+        defer {
+            semaphore.signal()
+            print("DEBUG upgrade complete")
+        }
+        print("DEBUG beginning upgrade")
+        
+        let (snapshot, post) = try await task()
```
Sjmarf:
```suggestion
        let (snapshot, comment) = try await task()
```

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/UpdateQueues/CommentUpdateQueue.swift
```diff
@@ -0,0 +1,129 @@
...
+        lastVerifiedSnapshot = snapshot
+        return post
```
Sjmarf:
```suggestion
        return comment
```

Sjmarf (changes requested):
Nice work! Found a couple of bugs

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/Caching/Caches/ReplyCaches.swift
EricBAndrews: I've implemented a very direct fix (just manually find and update the reply when doing the comment vote/save)--it's not ideal long-term but it should do for now. Once we get the `Notification`, we can just passthrough `votes` and `saved` to the underlying `Notification.comment2`, so it's not worth building anything more sophisticated right now imo.

Sjmarf (approved):
Nice, looks good! 🚀

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/Caching/Caches/ReplyCaches.swift
Sjmarf: Makes sense 👍

Sjmarf:
> > This one has a chunk of shim code in because of how comments and replies are intertwined; I'll do replies next and clean that out.
> 
> In the future we won't have a Reply type - I think it's possible to instead have a Notification type which stores an instance of Comment2. iirc this wasn't possible before we dropped support for 0.19.2. It may not be worth adding an update queue for replies at this stage.

Opened an issue to add this to the new `InboxNotification` type, which is the equivalent of `Reply`: #2336

PR #2223, opened 2025-08-22
Fix search sheet on iOS 26

### LINKED ISSUE #2205: iOS 26: Search sheets missing search bar

Sjmarf:
Closes #2205


Mlem/App/Views/Shared/Search/SearchBar/View+SheetSearchable.swift
```diff
@@ -0,0 +1,75 @@
...
+                            .autocorrectionDisabled()
+                    }
+                    .padding(-10)
+                }
+            }
+    }
+}
+
+extension View {
+    func sheetSearchable(
```
EricBAndrews: *nit*

Something about the "-able" naming pattern for a view modifier feels a little off--I think `.withSearchSheet` might be a little more standard

EricBAndrews (approved):
LGTM! One note re: naming

Mlem/App/Views/Shared/Search/SearchBar/View+SheetSearchable.swift
Sjmarf: I'm basing the name off the SwiftUI `.searchable` modifier here. I agree that it's a strange naming scheme. `.withSearchSheet` seems a little misleading to me - it implies that a new sheet is being added by the modifier, which isn't the case. The modifier just adds a search bar and cancel button to the view, and is only intended for use inside of a sheet. Maybe `.searchInsideSheet` or something? I'm not sure.

EricBAndrews: Fair point. I guess Apple does endorse "searchable," so I'd be fine keeping as is, though I still don't like that verbiage too much. The `.with...` pattern feels more natural for modifiers that add components to views--maybe `.withSheetSearch`?

Sjmarf: I'm happy with `.withSheetSearch` 👍

Issue #2215, opened 2025-08-15
iOS 26: Replace confirmation popovers with alerts


Sjmarf:
Replace all of the "are you sure?" popovers with alerts on iOS 26. We could also switch to alerts on iOS 18 for ease of development, maybe

PR #2225, opened 2025-08-23
Move QuickSwipes to package

### LINKED ISSUE #2094: Move QuickSwipes to package
This could be segmented off pretty easily. Decoupling the swipe actions from our existing action system will make it easier to implement the new action system I've been planning on doing at some point

### LINKED ISSUE #2204: Remove iOS 17 support code from QuickSwipes

Sjmarf:
The only functional change here is that any confirmation popups (e.g. when activating the "block" swipe action) now use `.alert()` on iOS 26 rather than `.confirmationDialog()`. On iOS 18 they continue to use `.confirmationDialog()`.

I've decoupled QuickSwipes from the `Action` system - it now uses a `QuickSwipeAction` type instead. At some point I want to rewrite the action system entirely, and this change will make it easier to do that.

Previously, we specified some of the QuickSwipes configuration (e.g. `minimumDrag`, `iconSize`) in a `behavior` parameter when using the `quickSwipes` view modifier:

``` swift
.quickSwipes(swipeConfiguration, behavior: .standard)
```

Now, these values are applied using environment variables instead:

``` swift
.quickSwipesDisabled(!quickSwipesEnabled)
.quickSwipeThresholds(primary: 60, secondary: 150, tertiary: 240)
.quickSwipeMinimumDrag(20)
.quickSwipeCornerRadius(Constants.main.standardSpacing)
.quickSwipeIconSize(28)
```

Closes #2094, closes #2204, partially completes #2215


EricBAndrews (approved):
LGTM! 🔥

PR #2228, opened 2025-08-24
Lemmy 1.0: Rearrange bridging system

Sjmarf:
> [!note]
> Relies on [this MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/33)

This PR refactors parts of the system that bridges between v3 and v4 sort types in the Lemmy API.

One effect of this is that sorting by "top" now works correctly on Lemmy 1.0.


EricBAndrews (approved):
LGTM! 💯

PR #2231, opened 2025-08-25
Profile editor tweaks

Sjmarf:
Made some tweaks to the avatar and banner sections of the profile editor. The "remove upload" button now uses a `trash` icon rather than an `xmark` and is slightly smaller. I also fixed a bug where the banner image wouldn't fill the frame correctly


EricBAndrews (approved):
Nice 👍

PR #2233, opened 2025-08-26
iOS26 settings icons fix

### LINKED ISSUE #2213: "Settings icons" setting is ignored on iOS 26

EricBAndrews:
## Issues

- closes #2213 

## Description

Implements `ConditionalIconToggleStyle`, which forces toggles' labels to respect `a11y_showSettingsIcons`, and adds it to all views that use `.labelStyle(.conditional)`.


Mlem/App/Views/Root/Tabs/Settings/ZoomSliderSettingsView.swift
```diff
@@ -69,6 +69,7 @@ struct ZoomSliderSettingsView: View {
             .pickerStyle(.inline)
         }
         .labelStyle(.conditional)
+        .toggleStyle(.conditional)
```
Sjmarf: We always use these two modifiers together. Maybe it's worth defining a `.withConditionalLabelStyle()` method on `View` that applies these? We could then define the `ConditionalLabelStyle` and `ConditionalToggleStyle` privately in the same file.

EricBAndrews: That is cleaner, implemented 👍

Sjmarf (approved):
LGTM 👍

PR #2235, opened 2025-08-27
Remove `FeedSelection`

Sjmarf:
We don't need this type any more because `Saved` is no longer a "feed". We're now using `ListingType` directly instead.


EricBAndrews (approved):
LGTM! 💯

PR #2236, opened 2025-08-28
New feed types

### LINKED ISSUE #2120: PieFed: Add "Popular" feed
There's API support for this.

### LINKED ISSUE #2130: Lemmy 1.0: Add "Suggested" feed
See also #2120

Sjmarf:
Added the "popular" feed for PieFed, and the "suggested" feed on Lemmy 1.0.

If you set "popular" or "suggested" as your default feed, Mlem will fall back to "subscribed" if the default feed type is unavailable.

Closes #2120, closes #2130


.swiftlint.yml
```diff
@@ -17,4 +17,6 @@ identifier_name:
   allowed_symbols: ["_"] # these are allowed in type names as we use them in API body arguments
 
 excluded:
+  - "**/.build"
+  - "**/build"
```
Sjmarf, line 21: Before this change, compiling the project outside of Xcode would cause SwiftLint to lint a load of dependency files. This fixes that issue. This doesn't affect how SwiftLint works in Xcode

EricBAndrews (changes requested):
Nice work!

Popular is not appearing in guest mode on piefed.social--I believe that's a bug, since it is available through the web ui when logged out.

Sjmarf:
You're right 👍 Fixed.

EricBAndrews (approved):
LGTM! 💯

PR #2243, opened 2025-08-29
[PATCH] French localizations

Sjmarf:
Copy over the french localizations to the branch for the next 2.2 patch.

PR #2244, opened 2025-08-29
Localization fixes

Sjmarf:
Fix a couple of places where localizations weren't applied properly:
- The "sign out" button in context menus in the account switcher
- The "blur NSFW" settings page

The strings were already translated, but weren't rendered properly.

Also changed "PiedFed" -> "PieFed" in French. The PieFed website refers to itself as "PieFed" in French.

These fixes were also applied in the patch PR: #2243


EricBAndrews (approved):
Nice 👍

PR #2245, opened 2025-08-31
Refactor `RestClient` to avoid `ErrorType` generic argument

Sjmarf:
If `RestClient` is unable to decode the response of a request into the intended type, it attempts to decode it as an error type (`ApiErrorResponse`) instead. Previously, this error type was specified via a generic argument on `RestClient`, and the logic was hard-coded. In this PR, the logic is instead injected into the `RestClient`. So, `RestClient` doesn't need to be generic anymore.

This makes it easier to use `RestClient` in a context where we don't need to decode an error type at all. That scenario doesn't exist yet, but might do in future if we use `RestClient` for the Mlem backend, Fediseer, lemmy-status etc.


EricBAndrews (approved):
Nice! This is a really useful change 😁

Issue #2198, opened 2025-08-10
iOS 26: Tab Bar Redesign


EricBAndrews:
### Requirements

- [x] There are no existing issues for this feature.
- [x] This is a request for a **single** feature (create multiple issues for multiple feature requests).

PR #2246, opened 2025-09-01
Basic iOS 26 account switcher implementation

EricBAndrews:
## Issues

- progress towards #2198 

## Description

Simple implementation of an account switcher swipe gesture compatible with iOS 26. As discussed in Slack, this is a temporary workaround; a robust UIKit solution would likely perform better, but this restores the missing functionality without too much trouble. I've put almost all the logic in a view modifier so it can be easily removed later.


Sjmarf (changes requested):
We should disable this on iPad

Sjmarf:
Forget to push? 🙃

EricBAndrews:
Yep 🤦

Sjmarf (approved):
LGTM 💯

PR #2247, opened 2025-09-03
Circle-based Search Design

EricBAndrews:
Redesigns the search homepage to use circles, plus adds `.gradient` to `Theming`:

image

Also includes a minor rework of `ThemedColor`, removing the `@Environment(\.palette)` declaration and replacing all instances (i.e., the unused `body` var) with functions that take in the palette instead. This is because the `@Environment` declaration in `ThemedColor` was never considered to be installed on a view, and therefore simply read the default value.

### DIFF
--- a/Mlem/Packages/Theming/Sources/Theming/ThemedColor.swift
+++ b/Mlem/Packages/Theming/Sources/Theming/ThemedColor.swift
@@ -9,18 +9,16 @@ import Foundation
 import SwiftUI
 
 public struct ThemedColor: ShapeStyle, Hashable, View, Sendable {
-    @Environment(\.palette) private var palette
-    
     fileprivate let hashString: String
     
     let getColor: @Sendable (Palette) -> Color
     var opacity: CGFloat = 1
     
-    public var body: some View {
+    public func body(palette: Palette) -> some View {
         resolve(with: palette)
     }
     
-    public var gradient: AnyGradient {
+    public func gradient(palette: Palette) -> AnyGradient {
         resolve(with: palette).gradient
     }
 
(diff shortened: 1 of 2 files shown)


Mlem/Packages/Theming/Sources/Theming/ThemedColor.swift
```diff
@@ -9,16 +9,18 @@ import Foundation
...
     var opacity: CGFloat = 1
     
-    public var body: some View {
+    public func body(palette: Palette) -> some View {
         resolve(with: palette)
     }
+    
+    public func gradient(palette: Palette) -> AnyGradient {
+        resolve(with: palette).gradient
+    }
```
Sjmarf, line 23: *nit* Perhaps a better way to do this would be to add an `isGradient` boolean to `ThemedColor`, in the same vein to what we've done for `opacity`. That avoids having to pass the `palette` as an argument whenever we want a gradient. I think it's better to avoid handing the `Palette` as much as possible within Mlem itself.

Sjmarf (approved):
LGTM 🔥

Mlem/Packages/Theming/Sources/Theming/ThemedColor.swift
EricBAndrews: The trouble with that (unless I'm misunderstanding) is that `.gradient` must return `AnyGradient`, where `.opacity` returns `Color`--this means that `resolve`, which should auto-apply such modifiers, would have to either ignore the gradient flag, return `AnyGradient`, or have some awkward middle ground.

One alternative is to add `isGradient` to the existing `.tint` override, which should cover a large majority of cases where we would be calling `.gradient`.

Sjmarf: Ugh, good point. I can't think of a better solution. Happy to merge with this :)

PR #2249, opened 2025-09-04
FeedLoadingState

Sjmarf:
Added a new type, `FeedLoadingState`, which is the same as `LoadingState` but has an extra case: `.initial`. We can't just add this case to `LoadingState` because `LoadingState` is used elsewhere.

No functional changes.


EricBAndrews (approved):
LGTM! 💯

PR #2250, opened 2025-09-04
Support labelled link with no scheme

Sjmarf:
Support links like `[label](example.com/c/comm)` (no scheme). _Unlabelled_ links with no scheme were already working fine, it's just labelled links that were problematic. Tapping on one of these would cause the app to crash.


EricBAndrews (approved):
Nice! 🔥

PR #2256, opened 2025-09-08
Optimize `PersonView` loading

Sjmarf:
Made `PersonView` load slightly faster, in theory, when opening from a link. Also, it was making a second load request in the background that wasn't needed, which I've removed.


EricBAndrews (approved):
LGTM! 🚀

PR #2257, opened 2025-09-08
Add "mark read" context menu action to crosspost list

Sjmarf:
As requested on c/mlemapp.

### DIFF
--- a/Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Post/Post2Providing.swift
+++ b/Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Post/Post2Providing.swift
@@ -81,10 +81,6 @@ public extension Post2Providing {
     
     /// Update the post when its queued mark read operation completes.
     func queuedMarkReadCompleted() {
-        guard post2.readQueued else {
-            assertionFailure("readQueueFlushed called but post was not queued")
-            return
-        }
         // sending this through the updateQueue ensures queue.lastVerifiedSnapshot receives the correct read value
         Task {
             await updateQueue.addItem { snapshot in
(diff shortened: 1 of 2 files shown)


Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Post/Post2Providing.swift
```diff
@@ -81,10 +81,6 @@ public extension Post2Providing {
     
     /// Update the post when its queued mark read operation completes.
     func queuedMarkReadCompleted() {
-        guard post2.readQueued else {
-            assertionFailure("readQueueFlushed called but post was not queued")
-            return
-        }
         // sending this through the updateQueue ensures queue.lastVerifiedSnapshot receives the correct read value
```
Sjmarf, line 84: Not sure if this is what I'm meant to be doing here.

EricBAndrews: This check is here to ensure the queueing system is working correctly--it should only ever be called by the queuing logic in cases where the post is, in fact, queued.

EricBAndrews (commented):
Code looks good! Will approve pending conversation re: `queuedMarkReadCompleted`.

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Post/Post2Providing.swift
Sjmarf: When I call `api.markPostsAsRead`, it triggers this assertion failure. I would expect `markPostsAsRead` to work even if the posts aren't queued. I'm not sure what the best way to fix this is, aside from just removing the assertion failure.

EricBAndrews: Ah, that makes sense--previously we only ever called the batch endpoint for mark read on scroll. I'm fine with removing it 👍

EricBAndrews (approved):
LGTM! 💯

PR #2258, opened 2025-09-08
disable mark read call when on piefed >= 1.1

### LINKED ISSUE #2183: Remove redundant call to mark posts as read when voted on or saved (PieFed >= 1.1)
https://codeberg.org/rimu/pyfedi/commit/18494e3224bafecf164d6b0e4c29896ab7073d39

#2177

EricBAndrews:
## Issues

- closes #2183 

## Description

- Adds `.autoMarkPostReadOnInteract` to `Feature` and logic to indicate it is supported on PieFed 1.1 and up
- Updates `PieFedConnection` to only manually mark read if `.autoMarkPostReadOnInteract` is not supported

### DIFF
--- a/Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/InstanceConnection/LemmyConnection/LemmyConnection+Feature.swift
+++ b/Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/InstanceConnection/LemmyConnection/LemmyConnection+Feature.swift
@@ -40,10 +40,8 @@ public extension LemmyConnection {
              .editAccountSettings, .viewMentionsAndPrivateMessages, .viewReports, .editAndDeletePrivateMessages,
              .reportPrivateMessages, .viewVotes, .purgeContent, .removeCommunity, .banFromInstance,
              .banFromCommunity, .editModeratorList, .commentSearch, .undeletePrivateMessages, .searchLocalPeople,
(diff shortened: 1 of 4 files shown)


Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/InstanceConnection/LemmyConnection/LemmyConnection+Feature.swift
```diff
@@ -42,6 +42,8 @@ public extension LemmyConnection {
              .banFromCommunity, .editModeratorList, .commentSearch, .undeletePrivateMessages, .searchLocalPeople,
              .hidePosts:
             true
+        default:
+            false
```
Sjmarf: I think we should be returning `true` for the new feature, not `false`. Though it doesn't make a difference in practice.

Also, I think we should avoid using `default` here - I've been specifying each case explicitly so that the compiler will complain when I add a new feature. This ensures I don't forget to consider whether Lemmy supports the feature or not.

EricBAndrews: Good point 👍

Sjmarf (approved):
LGTM 👍

PR #2267, opened 2025-09-11
iOS 26 style toolbars

### LINKED ISSUE #2261: Fix `CloseButtonView` on iOS 26
In fact, all toolbar buttons should be symbols rather than text

Sjmarf:
Fixes #2261

- Close buttons are now on the left where possible (on iOS 26 only)
- Toolbar actions associated with confirming something now have a prominent (blue) background
- We now use icons rather than text for toolbar button labels
- Redesigned the "select text" sheet a little:

IMG_2033


Sjmarf:
Build is probably failing because `.glassProminent` button style (new in iOS 26) is only available on the more recent Xcode 26 betas, and not the older Xcode 26 betas.

EricBAndrews (approved):
LGTM! 🔥

PR #2272, opened 2025-09-12
Visit history fixes

### LINKED ISSUE #2254: Potential PieFed account coding error when upgrading
### Requirements

### Description

Reported in Matrix chat.

Before releasing, ensure 2.2 -> 2.3 upgrade does not break PieFed accounts.

### Mlem Version

_No response_

Sjmarf:
If the visit history fails to decode, it'll now reset the visit history and write that to the disk. This will stop the error from appearing next time the user switches to that account. The error will still be shown the first time the user switches to an account where the visit history fails to decode.

When you log out, the visit history file will now be deleted.

Maybe closes #2254. But it's more of a patch over the actual problem.


EricBAndrews (approved):
LGTM! 💯

PR #2273, opened 2025-09-12
Fix markdown toolbar background on iOS 26

### LINKED ISSUE #2214: Markdown toolbar background missing on iOS 26
We should copy the notes app toolbar style (put it in a capsule)

Sjmarf:
Closes #2214


EricBAndrews (approved):
Nice work, the new toolbar looks great! 😁

PR #2274, opened 2025-09-12
Glass jump button

### LINKED ISSUE #2270: Use glass effect for jump button

EricBAndrews:
## Issues

- closes #2270


Sjmarf (approved):
LGTM 👍 

To my eyes, the jump button looks a little too far right in relation to the tab bar. It looks like the right edge of the jump button is aligned with the right edge of the tab bar. I actually wonder whether it would look better if we aligned the center of the jump button to the center of the tab bar's corner radius (where is starts to curve)? What do you think?

Happy to merge as-is; we're short on time

PR #2275, opened 2025-09-12
fix double switcher behavior

EricBAndrews:
Fixes an issue where the account switcher would open twice if the swipe up gesture was held when running on iOS 18


Sjmarf (approved):
Nice 👍

PR #2278, opened 2025-09-13
Fix SwiftUI previews

Sjmarf:
Previews were crashing, this fixes it.

Previews also currently return real data rather than mocked data as they should do - I'll fix this another time.


EricBAndrews (approved):
Nice 👍

Issue #2281, opened 2025-09-13
Change display name on PieFed


Sjmarf:
Pending API support

sunshinehot:
Backend support has been added. 

https://codeberg.org/rimu/pyfedi/commit/d1a75477cd168f3621ac545755d3da8a5070c732

Issue #2106, opened 2025-06-25
PieFed: Edit account settings


Sjmarf:

Sjmarf:
There is API support for this in PieFed 1.2

Sjmarf:
We've implemented profile editing, but not the other settings yet

Sjmarf:
https://codeberg.org/rimu/pyfedi/commit/4792fbd3c4bb42e1535e0bed68ce7c1028be88f6

PR #2282, opened 2025-09-13
Edit profile on PieFed

Sjmarf:
Partially completes #2106

There's no way to edit the display name through the PieFed API currently, so that field is hidden. Have opened #2281 to track this


EricBAndrews (approved):
LGTM! 💯

PR #2287, opened 2025-09-14
Gradients

EricBAndrews:
## Description

Applies gradients to "icon on solid color backdrop" views.

## Implementation Notes

For convenience, I have defined `.gradientTint`. There were three alternatives to this implementation that I considered; of the four choices, I believe this to be the cleanest. The other options are:
- Add `isGradient` flag to the existing `.tint`. This both breaks from the system API pattern and requires defining two completely separate bodies for the actual modifier; at that point, it seems more direct to simply define a separate modifier.
- Call the system `.tint` but pass in the contextually resolved output of `ThemedColor.gradient`. This is extremely verbose and also requires adding a ton of `@Environment(\.palette)`s.
- Create a separate modifier also named `.tint` that takes in a `Palette -> AnyGradient` function, then resolves it to a gradient in the modifier itself. This is the closest to the system behavior, since you can just pass in `ThemedColor.gradient`, but has the annoying quirk that you have to explicitly namespace `ThemedColor` in the invocation (e.g., `.tint(ThemedColor.upvote.gradient)` instead of simply `.tint(.upvote.gradient)`), which makes the simple `.gradientTint` a little more concise.

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Settings/GeneralSettingsView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/GeneralSettingsView.swift
@@ -7,6 +7,7 @@
 
 import Dependencies
 import SwiftUI
+import Theming
 
 struct GeneralSettingsView: View {
     // behavior
@@ -34,7 +35,7 @@ struct GeneralSettingsView: View {
                 description: "Manage your overall setup for Mlem.",
                 icon: .settings.general
             )
-            .tint(.themedNeutralAccent)
+            .tint(ThemedColor.themedNeutralAccent.gradient)
             Section {
                 NavigationLink(
                     "Default Feed",
--- a/Mlem/App/Views/Root/Tabs/Settings/GeneralSettingsView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/GeneralSettingsView.swift
@@ -35,7 +35,7 @@ struct GeneralSettingsView: View {
                 description: "Manage your overall setup for Mlem.",
                 icon: .settings.general
             )
-            .tint(ThemedColor.themedNeutralAccent.gradient)
+            .gradientTint(.themedNeutralAccent)
             Section {
                 NavigationLink(
                     "Default Feed",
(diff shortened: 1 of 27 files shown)


Mlem/App/Views/Root/Tabs/Settings/GeneralSettingsView.swift
```diff
@@ -34,7 +35,7 @@ struct GeneralSettingsView: View {
                 description: "Manage your overall setup for Mlem.",
                 icon: .settings.general
             )
-            .tint(.themedNeutralAccent)
+            .gradientTint(.themedNeutralAccent)
```
Sjmarf, line 38: I wish there was a way to keep using `.tint` here and apply the gradient inside of the label style, but I can't make it work 😔 I'm happy to use `.gradientTint` here instead

Sjmarf (approved):
Nice, looks good! 💯

Mlem/App/Views/Root/Tabs/Settings/GeneralSettingsView.swift
EricBAndrews: Yeah, the best I could get was `.tint(ThemedColor.upvote.gradient)` and that's just too wordy to be worth it ☹️

Sjmarf: Yeah I agree with that choice :)

What I meant was that the `SquircleLabelStyle` will _always_ have a gradient background, so it's unfortunate that we have to specify `gradientTint` when applying a tint to the label rather than just `.tint`. In an ideal world, I think it would be nicer to write:

```
.tint(.themedNeutralAccent)
```

...and then `SquircleLabelStyle` would apply the gradient itself, using something like:

```diff
struct SquircleLabelStyle: LabelStyle {
    func makeBody(configuration: Configuration) -> some View {
        HStack(alignment: .center, spacing: 16) {
            configuration.icon
-              .background(.tint)            
+              .background(.tint.gradient)  
       // ...
       }
   }
}  
```

EricBAndrews: Oh, that would be nice. Alas.

Issue #2191, opened 2025-08-07
iOS 26: Use native section index labels in subscription list


Sjmarf:
There's a native SwiftUI way to implement the subscription list "alphabet bar" in iOS 26. We can't completely replace our existing system until we drop iOS 18 support.

https://www.hackingwithswift.com/quick-start/swiftui/how-to-add-list-section-index-labels

EricBAndrews:
Closing this as completed in #2288, removing the iOS 18 implementation is OOS for this issue

PR #2288, opened 2025-09-14
Use native alphabet scroller for subscription list on iOS 26

### LINKED ISSUE #2286: Alphabet letters in subscription list don't work on iOS 26
We could just use the native implementation on iOS 26

https://github.com/mlemgroup/mlem/issues/2191

EricBAndrews:
## Issues

- closes #2286 
- progress towards #2191 

## Description

Uses the system alphabet scroller when iOS 26 is available.

This PR also includes one minor design change on iOS 26: the star label is now used to indicate favorites + defaults, allowing the user to scroll all the way to the top using the alphabet scroller.

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListView.swift
+++ b/Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListView.swift
@@ -28,12 +28,6 @@ struct SubscriptionListView: View {
     
     @Weak var form: UICollectionView?
     
-    var body: some View {
-        content
-            .listStyle(.sidebar)
-            .navigationTitle("Feeds")
-    }
-    
     var detailDisplayed: Bool {
         if UIDevice.isPad {
             noDetail ? false : navigation.path.isEmpty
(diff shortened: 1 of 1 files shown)


Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListView.swift
```diff
@@ -42,32 +36,39 @@ struct SubscriptionListView: View {
...
-                            }
-                        }
-                        .frame(maxWidth: .infinity, alignment: .leading)
-                        .contentShape(.rect)
-                    }
-                }
+            // TODO: iOS 18 deprecation remove compatibility shim
+            if #available(iOS 26, *) {
+                feeds
+                    .sectionIndexLabel("★")
```
Sjmarf, line 69: There's a version of `.sectionIndexLabel` that accepts a `Text` argument - it may be more idiomatic to write:

```suggestion
                    .sectionIndexLabel(Text(Image(icon: .lemmy.favorite)))
```

We might have to apply `.symbolVariant(.fill)` at the top level, unless SwiftUI applies that automatically. If that can't be made to work, we can use:

```suggestion
                    .sectionIndexLabel(Text(Image(systemName: "star.fill")))
```

Sjmarf (changes requested):
Nice work. It shows an additional label on my device:

IMG_2048

Possibly a result of me using the UK english localisation, which has a different spelling of "favourites". You're checking the section label against the string literal "favorites", which I'm guessing is the cause.

Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListView.swift
EricBAndrews: I can't get this to work--it just doesn't render the image. [This ancient thread](https://stackoverflow.com/questions/5898836/can-use-image-for-sectionindextitlesfortableview) suggests that perhaps it's a limitation of the API.

EricBAndrews:
> Possibly a result of me using the UK english localisation

Yup. I've replaced that logic with an explicit flag in `SubscriptionListSection` to track inclusion in the scroller.

Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListView.swift
Sjmarf: Ok, happy to leave as-is then 👍

Sjmarf (approved):
LGTM 👍

PR #2292, opened 2025-09-15
PieFed 1.2 support

Sjmarf:
> [!note]
> Relies on [this MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/35)

Updates the PieFed API types. In PieFed 1.2, they made a breaking change to the private message deletion endpoint. PieFed 1.2 isn't out just yet but it will be shortly, so this needs to go in Mlem 2.3.

The `pageCursor` -> `page` change is non-breaking; PieFed 1.2 continues to support `pageCursor`


EricBAndrews (approved):
LGTM! 💯

PR #2324, opened 2025-09-25
Hide instance block button on PieFed

Sjmarf:
Hid the "block" button for instances and the "instances" tab in the block list on PieFed.


EricBAndrews (approved):
LGTM! 💯

PR #2332, opened 2025-09-28
Snapshot extensions call base initializer

Sjmarf:
Previously, `...Snapshot+Lemmy.swift` files looked like this:

``` swift
extension Post1Snapshot {
    init(from post: LemmyPost) {
        self.field1 = post.field1
        self.field2 = post.field2
    }
}
```

Now, they look like this:

``` swift
extension Post1Snapshot {
    init(from post: LemmyPost) {
        self.init(
            field1 = post.field1,
            field2 = post.field2
        )
    }
}
```

This is just a refactor, there are no functional changes.

This change makes it possible to move the `...Snapshot+Lemmy.swift` files to a separate package (or target) to where the snapshot is defined. This wouldn't have been possible with how we were doing this before. I haven't actually done this yet.


EricBAndrews (approved):
LGTM! 💯

Issue #2157, opened 2025-07-19
Rethink MlemMiddleware notification types


Sjmarf:
We'll need to remove the `Reply` types entirely, and instead introduce a `Notification` model that looks somethink like this:

```swift
enum NotificationContent {
    case comment(Comment2)
    case post(Post2)
    case message(Message2)
}

@Observable
class Notification {
    var content: NotificationContent
    var read: Bool
}
```

This is similar to what we're already doing for `ModlogEntry`.

To find out: Are the properties in Reply2 a superset of those in Comment2 on 0.19.? I recall that this was not the case on some of the earlier Lemmy versions, but I don't remember whether that was 0.18 or 0.19. Things may get tricky if it's not a superset.

PR #2335, opened 2025-10-01
Bare-bones `InboxNotification` type

Sjmarf:
Partially completes #2157. We need to have a new inbox system working before Lemmy 1.0.

Adds a new type called `InboxNotification`. It can't be called `Notification`, becauase there's a type called `Notification` in the `Foundation` libary.

In this initial implementation, `InboxNotification` only has `id` and `read` properties:

``` swift
public class InboxNotification: ContentModel, Identifiable {
    public static var tierNumber: Int = 1
    public var api: ApiClient

    public let id: Int
    public var read: Bool
    
    // ...
}
```

In future, it will also have a `content: InboxNotificationContent` property, with `InboxNotificationContent` defined as:

``` swift
public enum InboxNotificationContent {
    case reply(Comment2)
    case mention(Comment2)
    case message(Message2)
}
```

For testing purposes, I've added a page labelled "Test Inbox" under developer settings. It loads some notifications and displays the ID and read status of each.

I've created new issues for the work that needs to be done next:
- #2336
- #2337
- #2339
- #2340
- #2338


EricBAndrews (approved):
LGTM! 💯

PR #2355, opened 2025-10-07
Fix changing account in post editor

### LINKED ISSUE #2354: Changing account inside post editor doesn't work
Reported by Sunshine on Matrix

Sjmarf:
`resolveCommunity` just wasn't being called anywhere.

Closes #2354

Do we want to put this in a patch? It's been broken for a while, so clearly no-one is using it, but it's quite a serious bug if someone _does_ use it.


EricBAndrews (approved):
Nice 👍 

Re: patch, not worth it IMO--as you mention it seems very rarely used (discoverability issue maybe?), and we can just point affected users to our TF.

PR #2356, opened 2025-10-07
Show website preview in post editor

Sjmarf:
When you add a link to a post in the post editor, it now shows the link's image and title like you see in the feed.

Screenshot 2025-10-07 at 21 14 37

On Lemmy, the image URL is fetched using `LemmyGetLinkMetadataRequest`. PieFed doesn't have an equivalent yet (I've opened [an issue](https://codeberg.org/rimu/pyfedi/issues/1382) asking for it), so on PieFed we're extracting the metadata from the website directly. This requires a new package dependency.

### DIFF
--- a/Mlem/App/Views/Pages/Editors/CommentEditor/PostEditor/PostEditorView+LinkView.swift
+++ b/Mlem/App/Views/Pages/Editors/CommentEditor/PostEditor/PostEditorView+LinkView.swift
@@ -51,6 +51,9 @@ extension PostEditorView {
     }
     
     private func generatePostLink(url: URL) async throws -> PostLink {
+        if let api = targets.first?.account.api, try await api.supports(.fetchLinkMetadata) {
+            return try await api.getPostLink(url: url)
+        }
         let metadata = try await OpenGraph.fetch(url: url)
         let thumbnailUrl = metadata[.image].map { URL(string: $0) } ?? nil
         return .init(content: url, thumbnail: thumbnailUrl, label: metadata[.title] ?? url.absoluteString)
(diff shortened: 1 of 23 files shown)


Mlem/App/Views/Pages/Editors/CommentEditor/PostEditor/PostEditorView+LinkView.swift
```diff
@@ -51,16 +40,22 @@ extension PostEditorView {
...
     
-    private var linkLabel: String {
-        switch link {
-        case let .value(url):
-            url.absoluteString
-        default:
-            .init(localized: "Add Link")
+    private func generatePostLink(url: URL) async throws -> PostLink {
+        if let api = targets.first?.account.api, try await api.supports(.fetchLinkMetadata) {
+            return try await api.getPostLink(url: url)
```
EricBAndrews, line 55: Is it worth doing this separately depending on backend support? Since we can do it ourselves, it seems simpler and more robust to just always go through OpenGraph. If PieFed adds support for this feature, then I think it makes sense to drop the package dep and just use the API, but until then I don't see a strong reason to use the API

EricBAndrews (changes requested):
Nice work!

Mlem/App/Views/Pages/Editors/CommentEditor/PostEditor/PostEditorView+LinkView.swift
Sjmarf: If the instance has image proxying enabled, it'll return a proxied image when you make a link metadata request. This is better for privacy.

Also, going through Lemmy guarantees that what you see in th post editor is what will be seen on Lemmy. At the moment they're just querying the OpenGraph data, so it's easy to replicate it. But who knows: they might change it, or someone might write a Lemmy plugin that changes it.

EricBAndrews: Fair points 👍

EricBAndrews (approved):
LGTM! 🚀

PR #2358, opened 2025-10-08
Backend V1

EricBAndrews:
- Updates the `BackendClient` to use the new v1 endpoint for instances
- Removes `version` from `InstanceSummary`'s `CodingKeys` (this was removed in 2.2, so all our users should have the new coding by now)
- Adds a dev toggle to use the QC backend

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Settings/DeveloperSettingsView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/DeveloperSettingsView.swift
@@ -65,6 +65,10 @@ struct DeveloperSettingsView: View {
             
             #if DEBUG
                 Section {
+                    Toggle("Use QC Backend",
+                           isOn: .init(get: { BackendClient.main.environment == .qc },
+                                       set: { BackendClient.main.changeEnvironment(to: $0 ? .qc : .prod) }))
+                    
                     Button(String("Trigger Onboarding")) {
                         navigation.showFullScreenCover(.onboarding)
                     }
@@ -72,7 +76,7 @@ struct DeveloperSettingsView: View {
                     Button(String("Reset Feed Welcome Banner")) {
                         showFeedWelcomePrompt = true
                     }
-                
+                    
                     Button(String("Reset Feed TestFlight Banner")) {
                         lastTestFlightUpdate = nil
                     }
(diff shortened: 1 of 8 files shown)


Mlem/App/Views/Root/Tabs/Settings/DeveloperSettingsView.swift
```diff
@@ -65,14 +65,18 @@ struct DeveloperSettingsView: View {
             
             #if DEBUG
                 Section {
+                    Toggle("Use QC Backend",
```
Sjmarf:
```suggestion
                    Toggle(String("Use QC Backend"),
```

Mlem/Localizable.xcstrings
```diff
@@ -9249,6 +9249,9 @@
           }
         }
       }
+    },
+    "Use QC Backend" : {
+
```
```suggestion
```

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Backend Client/BackendClient.swift
```diff
@@ -8,11 +8,22 @@
...
+public enum BackendEnvironment {
+    case qc, prod
+    
+    internal var address: URL {
+        switch self {
+        case .prod:
+            return .init(string: "https://backend.mlemapp.org:8443/")!
+        case .qc:
+            return .init(string: "https://backend.mlemapp.org:2096/")!
+        }
```
Sjmarf, line 18: *nit*

```suggestion
        switch self {
        case .prod: .init(string: "https://backend.mlemapp.org:8443/")!
        case .qc: .init(string: "https://backend.mlemapp.org:2096/")!
        }
```

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Backend Client/BackendClient.swift
```diff
@@ -31,49 +42,53 @@ public class BackendClient {
...
+        self.environment = environment
+        refresh()
+    }
+    
+    private func refresh() {
+        Task {
+            do {
+                try await fetchFlairs()
+                try await fetchTestflightUpdate()
+            } catch {
```
Sjmarf, line 76: Should we also be refreshing the instance array when the QA environment is toggled?

Sjmarf (changes requested):

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Backend Client/BackendClient.swift
EricBAndrews: Good point--I've updated the `ContentView` task that fetches instances to use this. We don't actually store the instances in the backend client, so we can't just hook that into here.

Sjmarf (approved):
LGTM! Nice work 🔥

PR #2362, opened 2025-10-11
Upload custom thumbnail in PostEditorView

Sjmarf:
When you attach a link to a post, you can now change the image used for that link.

If the link already has a default image (obtained using OpenGraph), the "change thumbnail" button is overlaid in the top-left.

If the link does not have a default image, it shows like this:


EricBAndrews (approved):
LGTM! 💯

PR #2363, opened 2025-10-11
Ban users on PieFed

### LINKED ISSUE #2160: PieFed: Ban from community
There's API support for banning users from communities, but there's multiple issues with that endpoint currently:

- https://codeberg.org/rimu/pyfedi/issues/1133
- https://codeberg.org/rimu/pyfedi/issues/1135
- https://codeberg.org/rimu/pyfedi/issues/1137

I'm not going to try to work around these. It'll be much easier to implement once these have been fixed

Sjmarf:
> [!note]
> Relies on [this MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/39).

Only available on PieFed 1.3 and above. Closes #2160.

### DIFF
--- a/Mlem/Packages/Rest/Sources/RestClient/RestClient.swift
+++ b/Mlem/Packages/Rest/Sources/RestClient/RestClient.swift
@@ -117,14 +117,7 @@ public class RestClient {
     func createBodyData(for defintion: any RequestWithBody) throws(RestError) -> Data {
         do {
             let encoder = JSONEncoder()
-            var formatter = ISO8601DateFormatter()
-            // `.withFractionalSeconds` is required for the PieFed banFromCommunity request
-            formatter.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
-            encoder.dateEncodingStrategy = .custom { date, encoder in
-                var container = encoder.singleValueContainer()
-                try container.encode(formatter.string(from: date))
-            }
-
+            encoder.dateEncodingStrategy = .iso8601WithMilliseconds
             let body = defintion.body ?? ""
             return try encoder.encode(body)
         } catch {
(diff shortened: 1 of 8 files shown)


Mlem/Packages/Rest/Sources/RestClient/RestClient.swift
```diff
@@ -117,6 +117,14 @@ public class RestClient {
...
         do {
             let encoder = JSONEncoder()
+            var formatter = ISO8601DateFormatter()
+            // `.withFractionalSeconds` is required for the PieFed banFromCommunity request
+            formatter.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
+            encoder.dateEncodingStrategy = .custom { date, encoder in
+                var container = encoder.singleValueContainer()
+                try container.encode(formatter.string(from: date))
+            }
+
```
Sjmarf: Before this change, it would encode a `Date` as a unix timestamp (integer).

I don't _think_ this will cause any issues. Lemmy explicitly uses `Int` as the type for requests that take date parameters (e.g. banning a user or scheduling a post). I don't think `Date`-encoded body fields are used anywhere aside from PieFed's ban request.

EricBAndrews: Yeah, that checks out 👍 

We may wish to bundle up these encoder settings into a `JSONEncoder` extension `.forPieFedBan` (or similar)--that way if we (a) need it elsewhere or (b) need to add more date encoding formats, we have a more reusable, extensible infrastructure in place

EricBAndrews (approved):
Nice work!

One comment re: extensibility--not blocking but maybe nice to have

Mlem/Packages/Rest/Sources/RestClient/RestClient.swift
Sjmarf: Good idea; I've moved this logic into a custom `dateEncodingStrategy` 👍

PR #2369, opened 2025-10-14
Use `.alert` rather than `.confirmationDialog` on iOS 26

### LINKED ISSUE #2215: iOS 26: Replace confirmation popovers with alerts

Sjmarf:
Closes #2215

Also decoupled `PopoverAnchorModel` from the action system. This will make it easier to rewrite the action system in future.


EricBAndrews (approved):
LGTM! 🔥

Issue #147, opened 2023-06-15
Share Posts and Comments as Image


ShadowJonathan:
Like Apollo, the ability to export a certain comment chain or post as an image;

![image](https://github.com/buresdv/Mlem/assets/22740616/58677339-f4a0-4be3-996a-f7f466ab6dc8)

This seems to suggest its relatively straight forward: https://stackoverflow.com/questions/30696307/how-to-convert-a-uiview-to-an-image

EricBAndrews:
Note from #2370: comments PR should include persistence of configuration settings

PR #2370, opened 2025-10-14
Save Post As Image

EricBAndrews:
## Issues

- progress towards #147 

## Description

Adds a menu option in the expanded post view to share the post as an image.

## Implementation Notes

There are several design choices that are driven by constraints of the image rendering system. Swift offers two basic ways to render a view to an image, one using UIKit and the other using SwiftUI. The UIKit approach doesn't play nice with image sizes (it ends up shrinking the image and adding dead space to the top of the output). There may be a way to fix this, but nothing I tried worked. The SwiftUI approach works more directly, but comes with two limitations:
1. All images except the main post image don't render unless the menu option is activated from within the expanded post view. For this reason, the menu option only appears in the expanded post view; I think this also makes decent sense from a design perspective, since menu space is already limited. I'm not sure what causes this behavior--I even tried an approach that loads the images ahead of time and pipes them directly down into MediaView, to no avail.
2. SFSymbols don't render correctly when interpolated in strings. This means flairs break, so I have disabled those too.


Mlem/App/Logic/ImageFunctions.swift
```diff
@@ -34,6 +34,19 @@ func saveMedia(url: URL) async {
...
 
+@MainActor
+func createImageFromView(_ view: some View, dimensions: CGSize? = nil) -> UIImage? {
+    let renderer = ImageRenderer(content: view)
+    renderer.scale = 3 // boost resolution to look better on larger devices
+    if let dimensions {
+        renderer.proposedSize = .init(dimensions)
+    } else {
+        // assume screen width
+        renderer.proposedSize.width = UIScreen.main.bounds.width
```
Sjmarf, line 45: Will this be a problem on iPad?

Mlem/App/Views/Shared/ExportableViews/ExportablePostEditorView.swift
```diff
@@ -0,0 +1,116 @@
...
+    }
+        
+    var exportablePost: some View {
+        ExportablePostView(post: post, showCommunity: showCommunity, showCreator: showCreator, showStats: showStats)
+            .allowsHitTesting(false)
+    }
+    
+    var snapshot: UIImage? {
+        createImageFromView(exportablePost)
+    }
```
Sjmarf: *nit* imo it's better to use functions rather than computed properties for resource-intensive operations, even when they have no arguments. This makes it clear to the caller that they should call this only once and reuse the value, rather than calling it many times.

Mlem/App/Views/Shared/ExportableViews/ExportablePostView.swift
```diff
@@ -0,0 +1,60 @@
...
+    }
+    
+    var body: some View {
+        content
+            .background(.themedBackground)
+            .clipShape(.rect(cornerRadius: Constants.main.standardSpacing))
+            .padding(Constants.main.standardSpacing)
+            .background(.themedGroupedBackground)
+            .animation(.snappy, value: animationHashValue)
+            .environment(AppState.main)
```
Sjmarf: IMO we should apply this at the call site. In future I want to deprecate access to this singleton

```suggestion
```

```diff
+ @Environment(AppState.self) var appState

// ...

   var exportablePost: some View {
          ExportablePostView(post: post, showCommunity: showCommunity, showCreator: showCreator, showStats: showStats)
              .allowsHitTesting(false)
+             .environment(appState)
    }
```

Mlem/App/Views/Shared/ExportableViews/ExportablePostEditorView.swift
```diff
@@ -0,0 +1,116 @@
...
+    @State var showCommunity: Bool = true
+    @State var showCreator: Bool = true
+    @State var showStats: Bool = true
+    
+    var body: some View {
+        ScrollView {
+            exportablePost
+                .padding(.bottom, 200)
+        }
+        .background(.themedGroupedBackground)
```
Sjmarf: In dark mode, this is the same color as the post background.  Not sure that using a black background for the sheet is the right answer, though... Maybe we could add a shadow to distinguish the post?

Mlem/App/Views/Shared/ExportableViews/ExportablePostEditorView.swift
```diff
@@ -0,0 +1,116 @@
...
+import Nuke
+import Media
+
+struct ExportablePostEditorView: View {
+    @Environment(NavigationLayer.self) var navigation
+    
+    let post: any Post1Providing
+    @State var showCommunity: Bool = true
+    @State var showCreator: Bool = true
+    @State var showStats: Bool = true
```
Sjmarf, line 26: It would be nice if these persisted between sessions. (Or we can do this in a later PR)

Mlem/App/Views/Shared/ExportableViews/ExportablePostEditorView.swift
```diff
@@ -0,0 +1,116 @@
...
+        .buttonStyle(.plain)
+        .padding(.horizontal, Constants.main.halfSpacing)
+    }
+    
+    @ViewBuilder var saveButton: some View {
+        if let imageData = snapshot?.pngData() {
+            Button("Save", icon: .general.import) {
+                Task {
+                    do {
+                        try await ImageSaver().writeImageToPhotoAlbum(imageData: imageData)
```
Sjmarf, line 113: Tapping save should have some feedback: a toast and maybe some haptics.

Sjmarf (changes requested):
Nice work! 😄 A few notes

Mlem/App/Views/Shared/ExportableViews/ExportablePostEditorView.swift
```diff
@@ -0,0 +1,116 @@
...
+        }
+        .toolbar {
+            ToolbarItem(placement: .topBarLeading) {
+                CloseButtonView(ios18Label: .cancel)
+            }
+            ToolbarItem(placement: .topBarTrailing) {
+                Menu("Details", systemImage: "slider.horizontal.3") {
+                    Toggle("Community", isOn: $showCommunity)
+                    Toggle("Creator", isOn: $showCreator)
+                    Toggle("Stats", isOn: $showStats)
```
Sjmarf: It would also be nice to have a "dark mode" toggle. When the system is in dark mode, it still produces a light mode image

Mlem/App/Logic/ImageFunctions.swift
EricBAndrews: I don't think so--the theory is to render it the same way it would appear on the user's device. There is a minor discrepancy between the preview and the output on iPad since the sheet is smaller than the full width of the screen, but I don't think that's too much of an issue.

Sample output from iPad:

image

Mlem/App/Views/Shared/ExportableViews/ExportablePostEditorView.swift
EricBAndrews: Fair, I've changed this to a state var that's recomputed when the toggles are updated 👍

Mlem/App/Views/Shared/ExportableViews/ExportablePostView.swift
EricBAndrews: Good point. As suggested doesn't quite work because environment values don't propagate nicely into ImageRenderer--I've updated this to take in `appState` as a direct parameter, and pass it in from ExportablePostEditorView

Mlem/App/Views/Shared/ExportableViews/ExportablePostEditorView.swift
EricBAndrews: Agreed, I think it'll be nicer to handle that all at once when we implement this feature for comments--hopefully we can reuse some of the post export system for the comment export.

Mlem/App/Views/Shared/ExportableViews/ExportablePostEditorView.swift
EricBAndrews: Oh right, totally forgot to add those 🤦 

I've set it to pop a toast but no haptics, same as saving images from the viewer.

Mlem/App/Views/Shared/ExportableViews/ExportablePostEditorView.swift
EricBAndrews: Whoops, had the wrong background set. The post is now on a slightly lighter card, same as it renders in dark mode normally, and OLED includes a border.

Mlem/App/Views/Shared/ExportableViews/ExportablePostView.swift
```diff
@@ -0,0 +1,67 @@
...
+    }
+    
+    var content: some View {
+        VStack(alignment: .leading, spacing: Constants.main.standardSpacing) {
+            if showCommunity {
+                FullyQualifiedLabelView(post.community_, labelStyle: .medium, showFlairs: false)
+                    .transition(.scale.combined(with: .opacity))
+            }
+            
+            LargePostBodyView(post: post, isPostPage: true, shouldBlur: false)
```
Sjmarf, line 51: Posts where the image proxy failed don't render properly; happy to leave OOS though

Mlem/App/Views/Shared/ExportableViews/ExportablePostEditorView.swift
```diff
@@ -0,0 +1,159 @@
...
+
+    var body: some View {
+        ScrollView {
+            exportablePost
+                .padding(.bottom, 200)
+        }
+        .task(id: snapshotRenderHashValue) {
+            snapshot = createImageFromView(exportablePost)
+        }
+        .background(.themedGroupedBackground)
```
Sjmarf: Now it's the same color as the post background in light mode 😄

Sjmarf (changes requested):

Mlem/App/Views/Shared/ExportableViews/ExportablePostEditorView.swift
EricBAndrews: Ok I think I have this fixed now--the entire sheet uses the global `.themedGroupedBackground` and the preview view applies its configured `.themedGroupedBackground` behind the configured `.themedSecondaryGroupedBackground` applied to the card. The intended behavior is:
- When the configured scheme matches the global scheme, the post card floats on top of a single-color background (sheet and preview have matching backgrounds)
- When the configured scheme does not match the global scheme, the post card is nested in a rectangle of its locally configured `.themedGroupedBackground`

Mlem/App/Views/Shared/ExportableViews/ExportablePostView.swift
EricBAndrews: We have a couple other proxy issues--I want to make that system more robust across the board. Let's leave it OOS and come back to it with better proxy infrastructure.

Sjmarf (approved):
Nice 😄

PR #2374, opened 2025-10-16
fix subscription list tint

### LINKED ISSUE #2216: Context menu tinting issue still present in subscription list
#2016

EricBAndrews:
## Issues

- closes #2216 

## Description

- Removes the default tint from our custom `Form`--as far as I can tell it didn't actually do anything other than prevent the form tint from being locally overridden.
- Tint the subscription list `.themedPrimary`
- Force a `.themedSecondary` tint on the `MediaView` spinner so it always renders as gray rather than picking up parent tint (this is consistent with other spinners in the app)

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListView.swift
+++ b/Mlem/App/Views/Root/Tabs/Feeds/SubscriptionListView.swift
@@ -63,7 +63,7 @@ struct SubscriptionListView: View {
     var content: some View {
         let sections = subscriptions?.visibleSections(sort: sort) ?? []
         
-        Form {
+        Form(tint: .themedPrimary) {
             // TODO: iOS 18 deprecation remove compatibility shim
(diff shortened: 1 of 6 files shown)


Sjmarf:
The icons for destructible actions are no longer red with this fix. Would supporting that be possible?

Sjmarf:
Also, the toggles in Settings are now green - do we want that?

EricBAndrews:
Nice catch. I've modified `Form` to take `tint` as an optional parameter.

Re: destructive actions, it looks like tinting a Form really aggressively applies that tint to everything inside it--we didn't encounter this before because we pretty much only use form in settings. I've updated `MenuButton` to force a `.themedNegative` tint, which gets the buttons correctly colored. I don't think this has any side effects anywhere else, since we were already coloring them that way.

Sjmarf (approved):
Nice 👍

Issue #2338, opened 2025-10-01
Remove `read` property from `Message2`


Sjmarf:
On Lemmy 1.0, messages themselves can no longer be "read", only the notifications associated with them can be.

This issue can't be done until we've pretty much fully replaced the old system with the new one

PR #2375, opened 2025-10-17
New action system

Sjmarf:
This PR implements a proof-of-concept for a new action system. The new system is being used for the context menus of messages in the conversation view. The system is **not** yet being used for messages in the inbox feed.

This PR contains one functional change - the "mark read" option is no longer available for messages in the conversation view. This is not a technical limitation; it needs to be removed anyway (#2338) so I though I'd do it in this PR.

I've described the new system below. Please point out any issues and/or suggest alternative ideas :)

# Technical Details

## Defining an action

Each action is now defined as a struct that conforms to `Action`. `Action` is defined as follows:

``` swift
public protocol Action {
    func createLabel(environment: EnvironmentValues) -> ActionLabel
    
    @MainActor
    func execute(environment: EnvironmentValues)
}
```

A simple implementation might look like:

``` swift
public struct UpvoteAction: Action {
    let entity: any Interactable2Providing

    func createLabel(environment: EnvironmentValues) -> ActionLabel {
        if entity.upvoted {
            .init(title: "Undo Upvote", icon: .lemmy.undoUpvote)
        } else {
            .init(title: "Upvote", icon: .lemmy.upvote)
        }
    }
    
    @MainActor
    func execute(environment: EnvironmentValues) -> ActionLabel {
        entity.toggleUpvoted()
    }
}
```

## Displaying an action

I've defined an `ActionButton` view with a constructor that takes `any Action` as an argument. `ActionButton` calls the `createLabel` method of the action, and renders the returned label on a button. When the button is clicked, it calls the `execute` method of the action.

## Feedback

When the user executes the upvote action, we might want to play haptics or show a toast. In the new system, we'd do this by accessing the `HapticManager` and/or `ToastModel` inside of the `execute` method.

``` swift
    @MainActor
    func execute(environment: EnvironmentValues) -> ActionLabel {
        entity.toggleUpvoted()
        environment.hapticManager.play(haptic: .lightSuccess, priority: .low)
    }
```

Similarly, the `PopupAnchorModel` can be accessed for showing confirmation dialogues, and the `NavigationLayer` can be accessed too.

## Customizability

Actions that may be used in customizable contexts should conform to the `ConfigurableAction` protocol, which is defined like so:

``` swift
public protocol ConfigurableAction: Action {
    static var label: ActionLabel { get }
}

public extension ConfigurableAction {
    func createLabel(environment: EnvironmentValues) -> ActionLabel {
        Self.label
    }
}
```

As you can see, this makes the label of the action static. This will allow us to render the label in the interaction bar editor without having to instantiate the `Action`.

The `ActionSeed` type is a codable representation of an `Action`. Each file that defines a `ConfigurableAction` struct should also include a piece of code that looks like this:

``` swift
extension ActionSeed {
    static let report = ActionSeed("report") { entity in
        switch entity {
        case let entity as any ReportableProviding: ReportAction(entity: entity)
        default: nil
        }
    }
}
```

This allows you to create an array of `ActionSeed` instances like so:

``` swift
let seeds: [ActionSeed] = [.upvote, .downvote, .save]
```

...and generate an array of `any Action` from the seeds like so:

``` swift
seeds.compactMap { $0.createAction(post) }
```

We'd store the user's interaction bar configuration as an array of `ActionSeed`. 

There is only one `ActionSeed` type, which will be used for all content types. This is a departure from our current generics-based system, where each content type has its own enum to represent the possible actions it can provide. Personally I thought that using generics was too difficult to manage, so I think we should switch to using just one type.


EricBAndrews (approved):
LGTM! This system is really clean 😁

PR #2378, opened 2025-10-20
Support new PieFed post link format

### LINKED ISSUE #2377: Support new PieFed link formats
PieFed have got a new format for post links (and possibly other content types?)

https://lemmy.ml/post/37705384/21713703

Sjmarf:
Support links that look like this:

https://piefed.social/c/privacy/p/1390258/printers-leave-a-watermark-on-each-page-indicating-the-exact-printer-that-it-came-from

Closes #2377


EricBAndrews (changes requested):
Nice work!

Link resolution is failing on simple PieFed URLs: `https://piefed.ca/post/286158`

Happy to leave OOS because this was also broken before this PR. The URL above doesn't nicely resolve with `resolve_object` on any instance I tried (even piefed.ca), but I was able to resolve it by fetching the post directly from `https://piefed.ca/api/alpha/post?id=286158` and resolving on the `api_id` of the returned post.

Sjmarf:
Ugh, you're right. Going to leave this for now, as this is a PieFed issue - it works fine on Lemmy. They can probably push a patch faster than we can.

Sjmarf:
Nevermind, it's only expected to work when you're logged in. One of the devs is now working on making it work while logged out, too (like Lemmy does).

EricBAndrews (approved):
LGTM! 🔥

PR #2379, opened 2025-10-20
Disable ban option for non-local communities on PieFed

Sjmarf:
PieFed only supports banning users if the community is on your local instance.

Relevant PieFed code: https://codeberg.org/rimu/pyfedi/src/branch/main/app/api/alpha/utils/community.py#L348


EricBAndrews (approved):
LGTM! 💯

PR #2380, opened 2025-10-20
Update posts, comments from non-entity calls

### LINKED ISSUE #2343: Posts don't update with changes from the remote
Possible updatequeue issue?

Repro:
Open profile on Mlem
Upvote/unupvote one of your own comments from the web UI
Refresh profile on Mlem
Nothing changes

Reported on Matrix by @sunshinehot

> It seems that voyager syncs upvotes quicker than Mlem.
> Like when I check the posts on my profile.
> Like one has 5 on Mlem then when I check it out on Voyager it’s at 13.

EricBAndrews:
## Issues

- closes #2343 

## Description

Adds the `attemptDirectUpdate` method to the `UpdateQueue` for posts and comments. This method takes in a snapshot and, if nothing is currently queued, updates the entity with the snapshot. If the queue is currently executing, nothing happens. Because this snapshot comes from an unknown context, it is not considered a verified snapshot and so `lastVerifiedSnapshot` is not updated.

`updateModel` for posts and comments has been updated to include a call to `attemptDirectUpdate`; this is a shim pending generic, universal update queues.

Also fixes an issue where comments would always use `snapshot1update` regardless of the model's tier. This was caused by the snapshot methods only being declared in extensions and not as part of the base protocol.

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -102,7 +102,7 @@
 		033F84C12C2AD072002E3EDF /* CommentTreeNode.swift in Sources */ = {isa = PBXBuildFile; fileRef = 033F84C02C2AD072002E3EDF /* CommentTreeNode.swift */; };
 		033F84C32C2B12AA002E3EDF /* InstanceSummary.swift in Sources */ = {isa = PBXBuildFile; fileRef = 033F84C22C2B12AA002E3EDF /* InstanceSummary.swift */; };
 		033F84C82C2B193D002E3EDF /* MlemStats.swift in Sources */ = {isa = PBXBuildFile; fileRef = 033F84C72C2B193D002E3EDF /* MlemStats.swift */; };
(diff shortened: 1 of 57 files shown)


Sjmarf (changes requested):
Nice! Do we also need to do this for the `InboxNotification` update queue?

EricBAndrews:
I decided not to add this to the inbox queues because this architecture will likely change somewhat when we get generic update queues built and I'd prefer not to add code I intend to rip up later unless it's important. I don't _think_ people are going to be refreshing inbox to watch vote counts change, but I don't mind too much either way--I'll add it to the inbox if you feel strongly that it would be a value add.

Sjmarf:
I could see someone refreshing the inbox to see the `read` value change if they update it on another client. IMO that's worth it, I don't mind if we don't update the `InboxNotification` content though :)

EricBAndrews:
That's a good point--I hadn't considered that case. We don't have an `updateModel` defined for `CoreCache`, so I just call the direct update outright if the model is retrieved. This behaves identically to posts and comments, just without the `updateModel` function wrapper.

Sjmarf (approved):
LGTM! 💯

Issue #931, opened 2024-02-17
Swift 6 strict concurrency mode


boscojwho:
We should get a head start on compiling with strict concurrency completely enabled before Swift 6 arrives:
- Initial build resulted in at least 2 errors and over 100 warnings (all related to Swift concurrency).
<img width="596" alt="Screenshot 2024-02-16 at 4 19 57 PM" src="https://github.com/mlemgroup/mlem/assets/2549615/b7977e71-4238-41ff-abf1-7e950ad8a06a">
<img width="85" alt="Screenshot 2024-02-16 at 4 20 39 PM" src="https://github.com/mlemgroup/mlem/assets/2549615/beb2b262-4581-47fb-8aa1-3c22be56330f">

Sjmarf:
https://www.swift.org/documentation/concurrency/

EricBAndrews:
Most of our packages right now are compiled with swift tools 5.9, which do not require strict concurrency. Bumping them to 6.2 requires updating a bunch of things to conform to `Sendable`.

PR #2387, opened 2025-10-24
Unified Logging

### LINKED ISSUE #124: Unified Logging Solution

EricBAndrews:
## Issues

- closes #124 

## Description

Replaces our print statement logging with a Swift Logger. Loggers in the app should be initialized using the `.mlemLogger` extension, which automatically uses the file name as the category. Subsystem should always be the package, though I couldn't find a nice way to get this automagically.

### Log Levels

I've migrated the existing prints into log levels according to the following basic categorization:
- Error: errors
- Warning: unexpected but recoverable behavior
- Info: important status changes (e.g., loaded a page)
- Debug: granular status changes (e.g., each step of loading a page). The idea here is that Info should indicate that the event happened, but not go into detail; debug traces step-by-step.

There is also a trace level available, but Xcode doesn't have a nice way to distinguish debug from trace in the output console (at least, not that I've found), so I have not gone to that level of detail.

### Singletons

There are also two singleton loggers:
- `Logger.universal`: sometimes access to a struct or class logger is not available. Logs from these contexts should be sent to the universal logger.
- `Logger.dev`: convenience logger for temporary dev logging. The app will not pass CI if statements to this logger are present.


Mlem/Packages/MlemLogger/Sources/MlemLogger/MlemLogger.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+//  Created by Eric Andrews on 2025-10-24.
+//
+
+import os
+
+public extension Logger {
+    static func mlemLogger(subsystem: String, file: String = #fileID) -> Logger {
+        Logger(subsystem: subsystem, category: file)
+    }
```
Sjmarf: Re: Having to pass the package name whenever we call `mlemLogger`

There are some options here. One would be to retrieve the package name from the file path, which can be obtained using `#file`. Another option would be to omit the package name entirely on the premise that our file names are (afaik) unique across all packages anyway.

I'm also happy to stick with the solution you've got here

Mlem/Packages/MlemLogger/Package.swift
```diff
@@ -0,0 +1,23 @@
+// swift-tools-version: 6.2
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "MlemLogger",
+    platforms: [.iOS(.v17)],
```
Sjmarf, line 8:
```suggestion
    platforms: [.iOS(.v18)],
```

Mlem/Packages/QuickSwipes/Sources/QuickSwipes/QuickSwipesViewModifier.swift
```diff
@@ -10,9 +10,13 @@ import Haptics
 import Icons
 import SwiftUI
 import Theming
+import MlemLogger
+import os
```
Sjmarf, line 14: We can avoid the double import by adding this to `MlemLogger.swift`:

```swift
public typealias Logger = os.Logger
```

Then, we can replace all cases of `import os` with `import MlemLogger`, which should clutter up the namespace a little less.

Sjmarf (approved):
Nice! We've needed something like this for a while 👀 A couple of suggestions

Mlem/Packages/MlemLogger/Sources/MlemLogger/MlemLogger.swift
EricBAndrews: Oh, good point--I've implemented a fully automated system that just parses `#file`. Using the package name for subsystem is still useful because you can use that to filter them.

Mlem/Packages/QuickSwipes/Sources/QuickSwipes/QuickSwipesViewModifier.swift
EricBAndrews: Unfortunately this causes build issues with package dependencies, since many of them use `Logger`. We could define a different typealias, but we would then also end up needing `MlemLogger` imports in a _lot_ of files (for some reason, once `os` is imported somewhere, `Logger` tends to resolve elsewhere in the package without issues)

Mlem/Packages/MlemLogger/Package.swift
EricBAndrews: This is actually a huge pain, since it forces minimum `.v18` and therefore Swift 6 on everything downstream--see #931

Sjmarf: Good point. We can add `.swiftLanguageMode(.v5)` to the `swiftSettings` array on the target to switch it back

```
        .target(
            name: "Actions",
            dependencies: [
                .byName(name: "Icons")
            ],
            swiftSettings: [
                .swiftLanguageMode(.v5),
            ]
        )
```

Mlem/Packages/QuickSwipes/Sources/QuickSwipes/QuickSwipesViewModifier.swift
Sjmarf: Ah ok, let's leave it as-is then 👍

Mlem/Packages/MlemLogger/Package.swift
EricBAndrews: True. For scope reasons I'd rather leave this as-is, I can open a small PR shortly after that bumps all the packages to .v18 and forces Swift 5 where needed.

PR #2390, opened 2025-10-26
Apply new action system to comments

### LINKED ISSUE #2389: Expand new action system to comment context menus

Sjmarf:
The ellipsis menu for comments now uses the new action system, under the following conditions:

- The comment is not in the mod mail feed (as I haven't implemented the "resolve" action yet)
- The setting "Moderation" -> "Moderator Actions" is set to "divider"

Only the ellipsis menu is using the new system, not the context menu. This is because the context menu always collapses actions into a disclosure group, which I haven't implemented yet.

-------

This PR contains one functional change. If you are both a mod and an admin, we would previously show two separate "ban" buttons if the user is banned from the community but not from the instance, or vice versa.

Now, we show a single ban button in this case. When tapped, this alert is shown:


Mlem/App/Actions/BanAction.swift
```diff
@@ -0,0 +1,161 @@
...
+        case (true, true, _, true), (false, true, false, true):
+            label = .init("Unban", icon: .lemmy.unbanFromInstance)
+        case (_, _, true, true):
+            label = .init("Ban...", icon: .lemmy.banFromInstance, isDestructive: true)
+        case (true, _, true, false):
+            label = .init("Unban", icon: .lemmy.unbanFromCommunity)
+        case (false, _, true, false):
+            label = Self.label
+        case (_, _, false, false):
+            return Self.label.withVisibility(.hidden)
```
EricBAndrews: This is super hard to parse--a brief explanatory comment of what gets displayed when would go a long way improving maintainability

Mlem/App/Actions/PurgeAction.swift
```diff
@@ -0,0 +1,69 @@
...
+
+extension PurgeAction {
+    static let label: ActionLabel = .init("Purge", icon: .lemmy.purge, isDestructive: true)
+    static let verboseLabel: ActionLabel = .init("Purge User", icon: .lemmy.purge, isDestructive: true)
+    
+    func createLabel(environment: EnvironmentValues) -> ActionLabel {
+        if useVerboseLabel {
+            if !(entity is any Person1Providing) {
+                assertionFailure()
+            }
```
EricBAndrews:
```suggestion
            assert(entity is any Person1Providing)
```

EricBAndrews (approved):
LGTM! 💯 

Just a couple nits

Mlem/App/Actions/BanAction.swift
Sjmarf: Good point. How's this?

EricBAndrews: Nice, that's much better 😁

PR #2391, opened 2025-10-26
Swift 6 Packages

EricBAndrews:
## Issues

- progress towards #931 

## Description

Updates the packages to use swift tools 6.2 and iOS v18, with a manual .v5 swift language mode enforced to make the whole thing compile. The Theming package also required a couple of `nonisolated` declarations.


Sjmarf (approved):
Nice 👍

PR #2393, opened 2025-10-28
Fix build

Sjmarf:
It builds just fine in Xcode but fails in CI.

> Unable to find module dependency: 'Haptics' import Haptics
> QuickSwipesViewModifier.swift:9


EricBAndrews (dismissed):
Nice 👍

EricBAndrews (approved):
LGTM! 💯

PR #2394, opened 2025-10-29
Use new action system for comments when using separate moderator menu

Sjmarf:
In #2390, I mentioned that the new action system was not yet used when using the separate moderator menu. This has now been implemented.


EricBAndrews (approved):
LGTM! 🔥

PR #2395, opened 2025-10-29
New saved feed loader

### LINKED ISSUE #2127: Rewrite saved feed loader to use separate endpoints
This is required for the saved feed to work on Lemmy 1.0.

Instead of using the `GetPersonDetails` feed loader for the saved feed, use a new feed loader with two child loaders - one for posts and one for comments.

On PieFed and Lemmy < 1.0, we can use `ListPostsRequest` and `ListCommentsRequest`. 

On Lemmy 1.0, we can use `ListPersonSavedRequest` and specify the `type_` parameter. This will require slightly more calls than we technically need. Once we drop 0.19.0 support (and if PieFed updates their API to match Lemmy) we can change the system again to just use `ListPersonSavedRequest`.

Sjmarf:
Closes #2127

# Rationale

Before this PR, we used `GetPersonRequest` and `PersonContentFeedLoader` to fetch the saved feed. In Lemmy 1.0, it is no longer possible to fetch both saved posts and comments from a single request. In this PR, I've introduced a new feed loader that uses the parent-child feed loader architecture, with saved posts and comments being fetched from different endpoints.

The new system is also used for the "upvoted" feed.

I haven't tested this on Lemmy 1.0. Any changes needed to support it should only be in `LemmyConnection`, which should be pretty easy.

# Functional Changes

Both the "Saved" and "Upvoted" feeds now have tabs:

# Technical Changes

`PersonContentFeedLoader`, which we used previously, has been renamed to `SingleSourceMixedFeedLoader`. The new feed loader is called `DualSourceMixedFeedLoader`. This is the best way I could think of to distinguish them

`SavedFeedView` and `UpvotedFeedView` have been merged into a single `VisitAgainView`, which takes in a parameter to specify the feed type.


EricBAndrews (approved):
LGTM! 🪨

EricBAndrews:
That was supposed to be 🚀 lol

Sjmarf:
😆

PR #2396, opened 2025-10-29
Fix ban actions sometimes not showing

Sjmarf:
Ban actions sometimes wouldn't show. Fixes bug introduced in #2390; don't include in patch notes


EricBAndrews (approved):
Nice 👍

Issue #2398, opened 2025-10-30
Edit community description on PieFed


Sjmarf:
Endpoint is currently broken I think
https://codeberg.org/rimu/pyfedi/issues/1425

Issue #981, opened 2024-03-20
Edit Community Settings


Sjmarf:
Allow moderators to edit the settings of a community

PR #2399, opened 2025-10-30
Edit community description

Sjmarf:
Progress towards #981

Added an "edit" button to the top-right of the description:

If the community has no description yet, you see this:

The API endpoint is currently bugged on PieFed, so this is only available on Lemmy for now. I've opened #2398 to track this


Mlem/App/Views/Pages/Editors/CommunityDescriptionEditorView.swift
```diff
@@ -0,0 +1,131 @@
...
+    @State var presentationSelection: PresentationDetent = .large
+
+    init(community: Community2) {
+        self.community = community
+        textView.text = community.description ?? ""
+    }
+
+    var body: some View {
+
+        CollapsibleSheetView(presentationSelection: $presentationSelection, canDismiss: !textHasChanged) {
```
EricBAndrews: *nit*
```suggestion
        CollapsibleSheetView(presentationSelection: $presentationSelection, canDismiss: !textHasChanged) {
```

EricBAndrews (changes requested):
Works really well! The edit button does shift all the text in the box down, though, which isn't ideal--some possible alternatives:
- Put the edit button at the bottom; this would mirror the "Add Moderator" button
- Put the edit button in the ellipsis menu. Like in Profile, this would require the edit sheet to include options for editing all the community info to satisfy user expectations.

Sjmarf:
Fair point 🤔

> Put the edit button at the bottom; this would mirror the "Add Moderator" button

This requires the user to scroll all the way to the bottom, which could be annoying if someone edits the description frequently.

> Put the edit button in the ellipsis menu. Like in Profile, this would require the edit sheet to include options for editing all the community info to satisfy user expectations.

This means that the button would be available on all pages of the `CommunityView`, which I don't particularly like. We _could_ only show the option on the "about" tab, but that feels weird

Maybe balancing it with a divider and title is enough? What do you think?

EricBAndrews (approved):
Works for me! 🔥

Issue #2197, opened 2025-08-10
iOS 26


EricBAndrews:
Update the app to support iOS 26, leverage new APIs to clean up existing code, and integrate the new Liquid Glass design language.

## Compatibility
- #2191 

## Cleanup
- #2196

## Design Changes
- #2198

Sjmarf:
Counting this as done

PR #2407, opened 2025-11-03
Liquid Glass Image Viewer Controls

EricBAndrews:
## Issues

- progress towards #2197

## Description

Uses liquid glass instead of material for the image viewer on iOS 26+

### DIFF
--- a/Mlem/App/Views/Pages/ImageViewer+Views.swift
+++ b/Mlem/App/Views/Pages/ImageViewer+Views.swift
@@ -288,13 +288,21 @@ extension ImageViewer {
     }
     
     @ViewBuilder
-    func videoStateButtonLabel(isOn: Bool, icons: (on: Icon, off: Icon)) -> some View {
-        Image(icon: isOn ? icons.on : icons.off)
-            .symbolVariant(.fill)
-            .scaledToFit()
-            .frame(width: 22, height: 22)
-            .contentTransition(.symbolEffect(.replace, options: .speed(2)))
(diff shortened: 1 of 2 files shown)


Mlem/App/Views/Pages/ImageViewer+Views.swift
```diff
@@ -246,32 +237,137 @@ extension ImageViewer {
...
+    
+    @ViewBuilder
+    var muteButtonContent: some View {
+        Image(icon: controlState.muted ? .general.mute : .general.unmute)
+            .scaledToFit()
+            .symbolVariant(.fill)
+            .frame(width: 22, height: 22)
+            .contentTransition(.symbolEffect(.replace, options: .speed(2)))
+            .padding(Constants.main.standardSpacing + 4) // +3 to match .title2 implicit padding plus offset
+    }
```
Sjmarf, line 389: There's some opportunity for re-use here.

Also, buttons labels should be of type `Label` with `.labelStyle(.iconOnly)`, which is better for accessibility than plain images. We've not been keeping up with accessibility particularly well, but worth noting anyway

EricBAndrews: Good point, implemented 👍 

One minor implementation note: developer tools now uses a `LocalizedStringResource` instead of a verbatim string. This is both technically more straightforward to pass into the label and IMO correct, since this is a tool that is available to users in the production build.

Sjmarf (approved):
LGTM! 🔥

PR #2411, opened 2025-11-06
Glass Animation Control Layer

EricBAndrews:
## Issues

- progress towards #2197 

## Description

Implements Liquid Glass for inline animation controls and the NSFW blur button. Also updates the blur button to match the mute button and uses `Label` for all overlay buttons.

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -150,10 +150,10 @@
 		035BE08D2BDE88EC00F77D73 /* NavigationLayerView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 035BE08C2BDE88EC00F77D73 /* NavigationLayerView.swift */; };
 		035BE08F2BDE911900F77D73 /* NavigationLayer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 035BE08E2BDE911900F77D73 /* NavigationLayer.swift */; };
 		035BE0912BDEA01E00F77D73 /* View+NavigationSheetModifiers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 035BE0902BDEA01E00F77D73 /* View+NavigationSheetModifiers.swift */; };
-		035DFA232EB3F7240021DE8C /* CommunityAboutView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 035DFA222EB3F7240021DE8C /* CommunityAboutView.swift */; };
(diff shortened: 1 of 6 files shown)


Sjmarf:
I'm not necessarily sure that we should be doing this. As far as I can tell, Apple exclusively uses liquid glass for controls that float _above_ the content. All of the controls mentioned here are part of the content (they scroll up and down inside of the `ScrollView`). IMO we should continue to use materials here.

Apple uses materials rather than liquid glass for their inline "play" buttons (e.g. App Store).

EricBAndrews:
The App Store is actually a bit of a mess--the play buttons are material, but the mute buttons are glass 🤔. We could just follow that directly, but IMO that's weirder than just picking one.

This is the relevant guidance from Apple's [materials guidelines](https://developer.apple.com/design/human-interface-guidelines/materials):

> ...use [standard materials](https://developer.apple.com/design/human-interface-guidelines/materials#Standard-materials) for elements in the content layer, such as app backgrounds. An exception to this is for controls in the content layer with a transient interactive element like [sliders](https://developer.apple.com/design/human-interface-guidelines/sliders) and [toggles](https://developer.apple.com/design/human-interface-guidelines/toggles); in these cases, the element takes on a Liquid Glass appearance to emphasize its interactivity when a person activates it.

Sjmarf:
> The App Store is actually a bit of a mess--the play buttons are material, but the mute buttons are glass 🤔.

For me they're both material 🤔



> This is the relevant guidance from Apple's [materials guidelines](https://developer.apple.com/design/human-interface-guidelines/materials):

> ..use [standard materials](https://developer.apple.com/design/human-interface-guidelines/materials#Standard-materials) for elements in the content layer, such as app backgrounds. An exception to this is for controls in the content layer with a transient interactive element like [sliders](https://developer.apple.com/design/human-interface-guidelines/sliders) and [toggles](https://developer.apple.com/design/human-interface-guidelines/toggles); in these cases, the element takes on a Liquid Glass appearance to emphasize its interactivity when a person activates it.
 
The proposed mute and play buttons liquid glass aren't "transient" in the way that the sliders and toggles are. The sliders and toggles are only liquid glass when you interact with them.

EricBAndrews:
Interesting, Apple isn't even consistent within its own App Store. This one's glass regardless of interaction:

Screenshot 2025-11-06 at 3 48 14 PM

Sjmarf:
Huh, that's interesting. Certainly seems to be against their own guidelines 🤔

Sjmarf (approved):
Nice 💯 

This was an issue before this PR, but maybe worth fixing here: the play button icon should be white in dark mode. It's currently black

PR #2415, opened 2025-11-07
Fix "description" field of post editor being blank when editing post

Sjmarf:
If a post has an image, the "description" field would be blank when you edit the post. This bug was introduced in #2356.


EricBAndrews (approved):
LGTM! 💯

PR #2423, opened 2025-11-09
Better `InstanceView` error handling

### LINKED ISSUE #2383: Hide "communities" tab in `InstanceView` for private communities
Reported by @sunshinehot on Matrix

We should also feature a "private" label more prominently, so people don't ask where the communities tab is

Sjmarf:
`InstanceView` now shows an `ErrorView` in two places:
- Over the whole page, if the instance data fails to load
- In the "communities" tab, if the community list fails to load

For two specific errors, I've added custom messages:

(First image is aussie.zone)

Closes #2383


Mlem/App/Views/Pages/Instance/InstanceView+Logic.swift
```diff
@@ -130,4 +130,43 @@ extension InstanceView {
...
+                await refresh()
+                return true
+            }
+
+            if case let ApiClientError.decoding(data, _) = error {
+                let string = String(data: data, encoding: .utf8)
+                if string?.contains("<title>Just a moment...</title>") ?? false {
+                    errorDetails?.title = .init(localized: "Blocked by Cloudflare")
+                    errorDetails?.icon = .general.cloudflare
+                    errorDetails?.body = .init(localized: "This instance uses Cloudflare, which prevents us from loading the page.")
```
EricBAndrews: *nit*

Can be reworded to avoid using "us":

```suggestion
                    errorDetails?.body = .init(localized: "This page cannot be loaded because of the instance's Cloudflare protection policy.")
```

EricBAndrews (changes requested):
Nice work! One note re: wording

Mlem/App/Views/Pages/Instance/InstanceView+Logic.swift
Sjmarf: Hmm, maybe "This page can't be displayed because Cloudflare blocked the request"?

EricBAndrews: Yeah, I like that 👍

EricBAndrews (approved):
LGTM! 🚀

PR #2424, opened 2025-11-09
Fix Loops embeds

### LINKED ISSUE #2421: Loops embeds broken
iirc we're doing some HTML5 shenanigans to get the video URL. It's possible that they changed the page format, causing this to no longer work

EricBAndrews:
## Issues

- closes #2421 

## Description

Loops recently changed how their webpage is displayed, so parsing the video source URL now requires two steps. First we have to fetch the page from the provided URL. This does not directly point us to the source URL; however, it _does_ contain some thumbnail information, which conveniently re-uses the API ID for the actual video (note that the original URL uses a one-way hash of the API ID, e.g., `https://loops.video/v/6DMa4tT2LZ` points to a video with API ID `119275713371843581`). With this information in hand we can then query the API directly and extract the source URL.


Sjmarf (approved):
Nice 💯

PR #2425, opened 2025-11-10
Theme fixes

### LINKED ISSUE #2021: Fix Dracula theme contrast
The tappable link contrast isn't great:

Sjmarf:
- Fixed the theme not applying to `UploadConfirmationView` background
- Fediseer Endorsement color now has its own value in `Palette` (rather than using a numbered accent)
- Fediseer Hesitations now use the `caution` color rather than a numbered accent
- Fediseer Censurse now use the `warning` color rather than a numbered accent
- Tweaked some dracula theme colors (see comparison below)
  - Turned up the brightness of `.secondary` slightly and made it slightly translucent
  - Reduced brightness of `.tertiaryGroupedBackground`
  
These color changes are technically a departure from the exact color values used by Dracula, but I think that's the lesser of two evils in this case.

| Old | New  |
| ------- | --- | 
| IMG_2178 | IMG_2177 |

Closes #2021


EricBAndrews (approved):
LGTM! 👍

PR #2426, opened 2025-11-10
Fix markdown toolbar hitbox

Sjmarf:
The top ~25% of the markdown toolbar was not registering taps. This fixes it. The top of the view was outside of the bounding box for the keyboard accessory.

Also fixed the dividers in the toolbar having uneven padding on iOS 26.


EricBAndrews (approved):
Nice 👍

sunshinehot:
Nice catch!

Issue #1642, opened 2025-01-22
Adopt full-size action sheet for posts/comments as we add more actions


Sjmarf:
Over time, we'll need to add more actions to the post/comment context menus. Rather than making those menus longer, I think we should add a "More..." button at the bottom of the context menu that opens a sheet something like we see in Mail:

<img src="https://github.com/user-attachments/assets/ebfe352e-d6d1-4252-b5a4-50af3759ff23" width=50%>

Inside of this menu, we could have an "Edit" button that lets the user pin/unpin actions to the context menu in the same way we do for the feed sort menu.

PR #2427, opened 2025-11-13
Action Sheet

Sjmarf:
This is only used in the test inbox for now, but I think we should adopt this everywhere down the line.

Partially implements #1642

The context menu of a notification now only shows a few items, and the others are hidden within a sheet. IMO the actions shown in the image below are sensible ones to make available all the time. upvote/downvote/save/reply aren't shown here because they are available in the interaction bar.

In future, we can make which actions appear here customizable.

The full sheet looks like this:

When you tap on an item that opens a sheet (e.g. Reply), the `NavigationPage` being opened _replaces_ the action sheet, re-using the same sheet. This makes the transition nice and fast, which I think is better than closing the action sheet and opening a new one. This change required updates to the navigation system.


EricBAndrews (approved):
LGTM! 💯 

This is a very smooth UX

PR #2431, opened 2025-11-14
URLQueryItemEncoder improvements

Sjmarf:
`URLQueryItemEncoder` has hard-coded support for the following data types:
- `String`
- `Int`
- `Double`
- `Bool`
- `URL`

Before this PR, if you wanted to add support to a type _not_ in the above list, you would have to declare conformance to `URLQueryItemEncodable`, which required implementation of `encodeInQueryItemFormat`:

``` swift
protocol URLQueryItemEncodable {
    func encodeInQueryItemFormat() -> String?
}
```

In this PR, I've removed `URLQueryItemEncodable`. Instead, types that you want to be encodable as a query item should conform to `Encodable`, and write one of the hard-coded types listed above to a `encoder.singleValueContainer()`. E.g.

``` swift
struct MyStruct: Encodable {
    func encode(to encoder: any Encoder) {
        var container = try encoder.singleValueContainer()
        try container.encode("Hello world")
    }
}
```

This change makes the system more idiomatic.

Also, I moved `URLQueryItemEncoder` to a separate target within the `Rest` package, for better organisation.


EricBAndrews (approved):
LGTM! 🚀

PR #2439, opened 2025-11-16
Update Lemmy 1.0

Sjmarf:
> [!note]
> See also the [MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/41) and the [lemmy-swift-codegen PR](https://github.com/mlemgroup/lemmy-swift-codegen/pull/20)

Lemmy made some changes:
- Raw enum values are now all snake_case rather than PascalCase
- Requests for deleting now use the DELETE method rather than POST

This PR adds support for those changes. This fixes logging in to Lemmy 1.0, but everything else is still broken on 1.0.


EricBAndrews (changes requested):
Code looks good!

I think something got borked in the endpoint version tracking, though--toggling "hide read" from the Subscribed feed on lemmy.ml trips the `assertionFailure` on `LemmyConnection:158`

Sjmarf:
Nice catch. I needed to pass the endpoint to the `JSONEncoder` used in the `PersistenceRepository`

EricBAndrews (approved):
LGTM! 💯

PR #2440, opened 2025-11-18
Better MlemStats error handling

Sjmarf:
Cloudflare is currently down, which causes the Mlem backend to be unreachable. This would cause an error toast on startup.

In this PR, I've removed that error toast. Instead, an `ErrorView` is shown in the instance list.

The `MlemStats` instance list is also used for `HandleThreadiverseLinksModifier`, but that code is capable of working fine without it. If the backend is offline, it'll just take a little longer to open some links while it contacts the instance directly.


EricBAndrews (approved):
LGTM! 🔥

PR #2441, opened 2025-11-18
Add new action system to `QuickSwipes`

Sjmarf:
You can now swipe on an item in the test inbox to mark it as read. This uses the new action system.


EricBAndrews (approved):
LGTM! 🚀

PR #2444, opened 2025-11-19
Lemmy 1.0 fixes

Sjmarf:
I have started testing against `voyager.lemmy.ml`, which runs Lemmy 1.0. This PR makes a few simple fixes. The feed now loads on Lemmy 1.0.

Also added an `ErrorView` to:
- `CommunityView`, if an upgrade fails
- `PersonView`, if an upgrade fails
- `PersonContentGridView`

The aforementioned `ErrorView` cases all appear when using Lemmy 1.0, because the associated requests currently fail.


EricBAndrews (approved):
LGTM! 💯

Issue #2361, opened 2025-10-09
Improved Blocking


EricBAndrews:
### Requirements

- [x] There are no existing issues for this feature.
- [x] This is a request for a **single** feature (create multiple issues for multiple feature requests).

### Description

Enhance the current blocking to better meet our users' needs:
- Differentiate keyword and string literal blocking. Currently we only offer keywords, but this doesn't work if a user wants to block tags like `[M]`, especially since those are sometimes embedded in words without spacing. String literal blocking would simply check if the given string appears anywhere in the post title.
- Add website blocking. I've seen a couple posts around Lemmy asking if this feature exists anywhere. Simply filter out posts that point to a particular host.

Sjmarf:
Lemmy 1.0 has backend support for both keyword filtering and website host filtering iirc. PieFed also has these but I'm not sure of the API status

IMO if both platforms support this, we should use the backend implementation rather than doing it locally.

EricBAndrews:
The downside of the backend implementation is that it is tied to a single account, while presumably users want these filters to be global. We have a couple ways to approach that:
- Use the backend implementation, and let the users handle synchronizing accounts
- Use the backend implementation and give the users a tool to synchronize filters between accounts
- Use a wholly client-side implementation
- Offer both client-side and backend filtering. This is the most powerful, but probably confusing

Sjmarf:
Yeah, that's a good question 🤔 I'm torn between these two:

> Use the backend implementation and give the users a tool to synchronize filters between accounts
> Offer both client-side and backend filtering. This is the most powerful, but probably confusing

I think including the backend system is important. We don't control whether a user sets filters up on desktop, for instance. 

If we stuck to just the backend system, we'd lose the ability to turn the filters on/off with a simple toggle.

EricBAndrews:
UI note: https://github.com/mlemgroup/mlem/pull/2454#discussion_r2551123672

PR #2454, opened 2025-11-21
Literal Filtering

EricBAndrews:
## Issues

- progress towards #2361 

Adds a literal filter, which filters out posts if any filtered literal appears as an exact substring of the post's title.

Also refactors filters to inherit from a common `FilterProviding` class, eliminating a bunch of redundant code.

### DIFF
--- a/Mlem/App/Globals/Definitions/FiltersTracker.swift
+++ b/Mlem/App/Globals/Definitions/FiltersTracker.swift
@@ -81,7 +81,7 @@ class FiltersTracker {
         rawKeywords = filteredKeywords
     }
     
-    func resetFilteredLiterals(to filteredLiterals: Set<String>) async {
+    func resetFilteredLiterals(to filteredLiterals: Set<String>) {
         literals = filteredLiterals
     }
     
--- a/Mlem/App/Globals/Definitions/FiltersTracker.swift
+++ b/Mlem/App/Globals/Definitions/FiltersTracker.swift
@@ -81,6 +81,10 @@ class FiltersTracker {
         rawKeywords = filteredKeywords
     }
     
+    func resetFilteredLiterals(to filteredLiterals: Set<String>) async {
+        literals = filteredLiterals
+    }
+    
     func postWouldBeFiltered(_ post: any Post) -> Bool {
         (keywordFilterEnabled && post.title.failsKeywordFilter(keywords: keywords, phrases: phrases)) ||
         (literalFilterEnabled && post.title.failsLiteralFilter(literals: literals))
--- a/Mlem/App/Globals/Definitions/FiltersTracker.swift
+++ b/Mlem/App/Globals/Definitions/FiltersTracker.swift
@@ -18,6 +18,8 @@ class FiltersTracker {
             (self.keywords, self.phrases) = parseKeywordsAndPhrases(from: rawKeywords)
         }
     }
+    @ObservationIgnored @Setting(\.filters_literalFilterEnabled) var literalFilterEnabled
+    @ObservationIgnored @Setting(\.filters_literals) var literals
     
     var isAdmin: Bool
     var moderatedCommunityActorIds: Set<ActorIdentifier>
@@ -33,7 +35,8 @@ class FiltersTracker {
             isAdmin: isAdmin,
             moderatedCommunityActorIds: moderatedCommunityActorIds,
             filteredKeywords: keywordFilterEnabled ? keywords : .init(),
-            filteredPhrases: keywordFilterEnabled ? phrases : .init()
+            filteredPhrases: keywordFilterEnabled ? phrases : .init(),
+            filteredLiterals: literalFilterEnabled ? literals : .init()
         )
     }
     
--- a/Mlem/App/Globals/Definitions/FiltersTracker.swift
+++ b/Mlem/App/Globals/Definitions/FiltersTracker.swift
@@ -60,7 +60,7 @@ class FiltersTracker {
     }
     
     func addFilteredKeyword(_ keyword: String) async {
-        rawKeywords = rawKeywords.union([keyword])
+        rawKeywords.insert(keyword)
     }
     
     func removeFilteredKeyword(_ keyword: String) async {
@@ -69,7 +69,7 @@ class FiltersTracker {
     }
     
     func addFilteredLiteral(_ literal: String) async {
-        literals = literals.union([literal])
+        literals.insert(literal)
     }
     
     func removeFilteredLiteral(_ literal: String) async {
--- a/Mlem/App/Globals/Definitions/FiltersTracker.swift
+++ b/Mlem/App/Globals/Definitions/FiltersTracker.swift
@@ -45,6 +45,8 @@ class FiltersTracker {
         hasher.combine(moderatedCommunityActorIds)
         hasher.combine(rawKeywords)
         hasher.combine(keywordFilterEnabled)
+        hasher.combine(literals)
+        hasher.combine(literalFilterEnabled)
         return hasher.finalize()
     }
     
@@ -66,12 +68,22 @@ class FiltersTracker {
         rawKeywords = rawKeywords.subtracting([keyword])
     }
     
+    func addFilteredLiteral(_ literal: String) async {
+        literals = literals.union([literal])
+    }
+    
+    func removeFilteredLiteral(_ literal: String) async {
+        assert(literals.contains(literal), "Filtered literals do not contain \(literal)")
+        literals.remove(literal)
+    }
+    
     func resetFilteredKeywords(to filteredKeywords: Set<String>) async {
         rawKeywords = filteredKeywords
     }
     
     func postWouldBeFiltered(_ post: any Post) -> Bool {
-        keywordFilterEnabled && post.title.failsKeywordFilter(keywords: keywords, phrases: phrases)
+        (keywordFilterEnabled && post.title.failsKeywordFilter(keywords: keywords, phrases: phrases)) ||
+        (literalFilterEnabled && post.title.failsLiteralFilter(literals: literals))
     }
     
     static var main: FiltersTracker = .init()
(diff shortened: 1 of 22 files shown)


Mlem/App/Globals/Definitions/FiltersTracker.swift
```diff
@@ -63,12 +68,22 @@ class FiltersTracker {
         rawKeywords = rawKeywords.subtracting([keyword])
     }
     
+    func addFilteredLiteral(_ literal: String) async {
+        literals = literals.union([literal])
```
Sjmarf:
```suggestion
        literals.insert(literal)
```

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
```diff
@@ -12,27 +12,38 @@ struct FiltersSettingsView: View {
...
         @Dependency(\.persistenceRepository) var persistenceRepository
     }
     
     var body: some View {
         Form {
-            Section {
-                Toggle("Enable Keyword Filters", icon: .settings.keywordFilter, isOn: $keywordFilterEnabled)
-            }
+            SettingsHeaderView(
+                title: "Filters",
```
Sjmarf, line 37: This page is now titled "Filters", but the `NavigationLink` that leads to it is still labelled "Keyword filters". IMO we should be consistent here.

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
```diff
@@ -12,27 +12,38 @@ struct FiltersSettingsView: View {
...
             } footer: {
                 // swiftlint:disable:next line_length
-                Text("Posts with these keywords in their titles will be hidden. If you are a moderator or administrator of a matching post, it will appear in your feed but require you to tap to view its content.")
+                Text("Hide posts with titles containing these whole words or phrases. Ignores case and punctuation (e.g., the keyword \"john\" will also filter \"John's\").")
+            }
+            
+            Section("Literals") {
+                literalSection
+            } footer: {
+                Text("Hide posts with titles containing containing these precise character sequences")
```
Sjmarf:
```suggestion
                Text("Hide posts with titles containing containing these precise character sequences.")
```

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
```diff
@@ -111,4 +124,51 @@ struct FiltersSettingsView: View {
...
+    var literalSection: some View {
+        Toggle("Enable", icon: .settings.keywordFilter, isOn: $literalFilterEnabled)
+        
+        TextField("New Literal...", text: $newLiteral)
+            .textCase(.lowercase)
+            .textInputAutocapitalization(.never)
+            .submitLabel(.done)
+            .onSubmit {
+                saveNewLiteral()
+            }
```
Sjmarf, line 151: I wonder whether we should have a "submit" button to the right of the text field, in addition to the field being submittable using the return button in the keyboard. I could see people being confused about where the submit button is.

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
```diff
@@ -12,27 +12,38 @@ struct FiltersSettingsView: View {
...
+                Text("Hide posts with titles containing these whole words or phrases. Ignores case and punctuation (e.g., the keyword \"john\" will also filter \"John's\").")
+            }
+            
+            Section("Literals") {
+                literalSection
+            } footer: {
+                Text("Hide posts with titles containing containing these precise character sequences")
             }
         }
         .withConditionalLabelStyle()
```
Sjmarf, line 56: Currently there's no way to dismiss the keyboard

```suggestion
        .scrollDismissesKeyboard(.interactively)
        .withConditionalLabelStyle()
```

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
```diff
@@ -12,27 +12,38 @@ struct FiltersSettingsView: View {
...
     
     var body: some View {
         Form {
-            Section {
-                Toggle("Enable Keyword Filters", icon: .settings.keywordFilter, isOn: $keywordFilterEnabled)
-            }
+            SettingsHeaderView(
+                title: "Filters",
+                // swiftlint:disable:next line_length
+                description: "Hide posts containing certain words, phrases, or character sequences from your feed. If you are a moderator or administrator of a filtered post, it will appear in your feed but require you to tap to view its content.",
```
Sjmarf: (Optional)

We could consider hiding the second sentence if the user has no moderator accounts, which can be determined using:

```
AccountsTracker.main.highestLevelAccountType >= .moderator
```

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
```diff
@@ -111,4 +124,51 @@ struct FiltersSettingsView: View {
...
+                
+                Spacer()
+                
+                // using a Button to do this makes the whole row register tap gestures :/
+                Image(icon: .general.delete)
+                    .foregroundStyle(.themedWarning)
+                    .onTapGesture {
+                        deleteLiteral(literal)
+                    }
+            }
```
Sjmarf, line 165: Possible opportunity for code re-use with the `ForEach` a little further up

Sjmarf (changes requested):
Nice, this is very useful 😄 Some comments

Mlem/App/Globals/Definitions/FiltersTracker.swift
Sjmarf: There's also another case of this further up the file; can't submit a suggestion on it though as it's outside the diff.

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
Sjmarf, line 1: The import/export functionality doesn't currently include literals.

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
Sjmarf: Also, the "Keyword filters" navigation link has some secondary text. Currently, it displays "None" when there are no keyword filters, even if there are literal filters configured. We should change that

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
EricBAndrews: That's a good idea, but I'm leaving OOS for now--this entire UI is going to get an overhaul with the blocklists concept discussed in Slack. I've added a note to the improved blocking issue to track this. The scope of this PR is primarily refactoring and extending the filtering backend.

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
EricBAndrews: See https://github.com/mlemgroup/mlem/pull/2454#discussion_r2553577025

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
EricBAndrews: Nice idea! Implemented.

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
EricBAndrews: Nice catch!

I've updated it to handle both. This required updating the serialization schema. The user can still import the old format, but the app will prompt to re-export in the new format; in a future version we can remove this compatibility logic and just use the new format.

Mlem/App/Globals/Definitions/FiltersTracker.swift
```diff
@@ -81,6 +81,10 @@ class FiltersTracker {
         rawKeywords = filteredKeywords
     }
     
+    func resetFilteredLiterals(to filteredLiterals: Set<String>) async {
+        literals = filteredLiterals
+    }
```
Sjmarf:
```suggestion
    func resetFilteredLiterals(to filteredLiterals: Set<String>) {
        literals = filteredLiterals
    }
```

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
Sjmarf: Nice 👍

Mlem/App/Views/Root/Tabs/Settings/FiltersSettingsView.swift
```diff
@@ -20,16 +19,20 @@ struct FiltersSettingsView: View {
...
+        ""
     }
     
     var body: some View {
         Form {
             SettingsHeaderView(
                 title: "Filters",
-                // swiftlint:disable:next line_length
-                description: "Hide posts containing certain words, phrases, or character sequences from your feed. If you are a moderator or administrator of a filtered post, it will appear in your feed but require you to tap to view its content.",
+                description: "Hide posts containing certain words, phrases, or character sequences from your feed.\(descriptionDetail)",
```
Sjmarf: We could avoid exposing our method of joining these strings to the `Localizable.xcstrings`. A good way to do this might be

```swift
var headerDescription: String {
        var output = String(localized: "Hide posts containing certain words, phrases, or character sequences from your feed.")
        if AccountsTracker.main.highestLevelAccountType >= .moderator {
            output += " "
            output += String(localized: "If you are a moderator or administrator of a filtered post, it will appear in your feed but require you to tap to view its content.")
        }
        return output
```

Sjmarf (approved):
LGTM, a couple of minor nits

PR #2455, opened 2025-11-22
Modlog user filter

### LINKED ISSUE #1588: Filter Modlog by Moderator
This could be a picker with all active moderators rather than a sheet. At the bottom of the picker could be a disclosure group for ex-moderators, and another disclosure group for admins. That should cover all possible moderators.

### LINKED ISSUE #1589: Filter Modlog by Target User

Sjmarf:
Closes #1588, closes #1589

You can now filter the modlog by target user and by moderator. There's a new "modlog..." button in the profile ellipsis menu that opens the modlog for that user.


EricBAndrews (approved):
LGTM! 🚀 

A couple minor notes:
- Do we need the ellipsis on `Modlog...`? Right now AFAIK we only use it for actions that open a full sheet (e.g., `More...`), while alert-style interactions (e.g., block) omit the ellipsis
- It would be great to skip the dialog if the user has no moderated communities, but that requires a `Person3Providing`. I can't think of a nice way to do that off the top of my head, but once we get the unified models in place this should be relatively straightforward

Sjmarf:
> Do we need the ellipsis on Modlog...? Right now AFAIK we only use it for actions that open a full sheet (e.g., More...), while alert-style interactions (e.g., block) omit the ellipsis

Good point. I don't mind either way but agree we should be consistent - I'll remove it so that we're consistent with Block.

> It would be great to skip the dialog if the user has no moderated communities, but that requires a Person3Providing. I can't think of a nice way to do that off the top of my head, but once we get the unified models in place this should be relatively straightforward

In theory less taps is indeed better. However, having the alert show only some of the time could harm muscle memory. Happy to revisit once there's unified models. 

If we get the full-size action sheet in-place for users, we could have separate "modlog as subject" and "modlog as moderator" buttons, and only show the relevant one. I don't want to do this rn because it would take up a lot of space in the menu, but that wouldn't be an issue if we had a full-size sheet.

PR #2457, opened 2025-11-25
Bugfix scroll to top on tab reselect

EricBAndrews:
iOS 26 implemented a default navigate back on tab reselect, which made reselect to scroll to top stop working. This adds a check to suppress that system behavior if there is a `.tabReselectConsumer` present.


Sjmarf (approved):
Nice 👍

PR #2461, opened 2025-11-26
Add swipe action colors to new action system

### LINKED ISSUE #2459: Add customizable colors to new action system

Sjmarf:
Closes #2459. The "mark read" swipe action in the test inbox is now purple.

Also fixed a bug where `RemoveAction` would still be labelled "remove" when a comment is already removed, rather than "restore". This affected the ellipsis menu in comments, which uses the new action system.


EricBAndrews (approved):
LGTM! 🔥

PR #2462, opened 2025-11-28
Enable Comment Search on PieFed 1.3

### LINKED ISSUE #2134: PieFed: comment search
Pending API support https://codeberg.org/rimu/pyfedi/issues/1014

Sjmarf:
Closes #2134

The API is much slower than Lemmy, but it does eventually return


EricBAndrews (approved):
Nice 👍

PR #2464, opened 2025-11-30
Fix newly-created comment not showing when post had zero comments

Sjmarf:
If you replied to a post that has zero comments, it would not show the new comment. This has been fixed.

This may not completely fix the comments-not-showing up issue in the general case, I'm not sure.


EricBAndrews (approved):
LGTM! 🚀

PR #2465, opened 2025-11-30
Support granular Lemmy 1.0 instance-wide vote settings

### LINKED ISSUE #2450: Lemmy 1.0 support: split `Instance.enableDownvotes` into separate settings for posts and comments

Sjmarf:
In Lemmy 0.19, there was a `downvotesEnabled` instance setting. In Lemmy 1.0, this has been replaced with four new settings:
- Post upvotes
- Post downvotes
- Comment upvotes
- Comment downvotes

Each setting is a three-case enum:
- On
- Off
- Local only

This PR adds support for those settings.

I have _not_ added support for hiding the upvote button when "post upvotes" or "comment upvotes" is disabled. This is because the action system will be replaced soon anyways. I have added support for this to the new action system.

Closes #2450


EricBAndrews (approved):
LGTM! 💯

PR #2466, opened 2025-12-01
Fix haptic error AGAIN

### LINKED ISSUE #2446: "Haptic manager failed to start" shows when dev mode off
There's some code that's _meant_ to stop this from showing when dev mode is off, but apparently errors are still slipping through.

```
Optional(HapticManager player failed to start. Underlying error: Error Domain=[com.apple](http://com.apple/).CoreHaptics Code=-4804 "(null)" UserInfo={Error =self.client != nil})
```

```swift
    func handleHapticError(_ error: HapticError) {
        let silent = switch error {
        case .failedToStartEngine: developerMode
        default: false
        }
        handleError(error, silent: silent)
    }
```

Sjmarf:
I'm hoping this finally fixes it

Closes #2446

### DIFF
--- a/Mlem/App/Views/Root/ContentView+Logic.swift
+++ b/Mlem/App/Views/Root/ContentView+Logic.swift
@@ -51,10 +51,6 @@ extension ContentView {
     }
     
     func handleHapticError(_ error: HapticError) {
-        let silent = switch error {
-        case .failedToStartEngine, .failedToStartPlayer, .failedToMakePlayer: !developerMode
-        default: false
-        }
-        handleError(error, silent: silent)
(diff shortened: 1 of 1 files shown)


EricBAndrews (changes requested):
Can we just make `silent = !developerMode` across the board? I haven't encountered any issues with the haptic engine not running at all in a very long time, which is the only case where even displaying haptic-related info to the user might be useful.

Sjmarf:
Good idea 👍

EricBAndrews (approved):
LGTM! If this doesn't bury the issue I don't know what will 😅

PR #2470, opened 2025-12-03
Tap and hold version number to copy

Sjmarf:


EricBAndrews (changes requested):
This might be smoother as a single tap that opens a "Copy Item" dialog, like how Passwords works. Also fine to merge as-is if that's a hassle, this isn't a flow we need to bikeshed too hard

Sjmarf:
Good point. I tried doing it with a `Menu` but it didn't feel right. How about making tapping the version number copy the number to clipboard directly, without any secondary button at all?

EricBAndrews (approved):
LGTM! 🚀

PR #2471, opened 2025-12-03
Disable sign up on PieFed

### LINKED ISSUE #2238: Disable sign up on PieFed
We currently allow users to fill out the whole form with no warning that it won't work when they hit "submit"

Sjmarf:
If you click "sign up" from the account list, PieFed instances are now grayed out in the search list.

Closes #2238


EricBAndrews (approved):
LGTM! 🔥

PR #2478, opened 2025-12-06
Use new action system for `Person`

Sjmarf:
The new action system is now used for `Person` models, in all places.


Mlem/App/Actions/AppointAdminAction.swift
```diff
@@ -0,0 +1,103 @@
...
+
+        return label.withVisibility(visibility(environment))
+    }
+
+    private func visibility(_ environment: EnvironmentValues) -> ActionVisiblity {
+        guard let myPerson = entity.api.myPerson else { return .hidden }
+        guard entity.api.canInteract(appState: environment.appState) else { return .hidden }
+        guard entity.api.isAdmin else { return .hidden }
+        guard !(entity.isAdmin_ ?? false) || entity.api.isHigherAdmin(than: entity) else { return .hidden }
+        guard entity.apiIsLocal else { return .hidden }
```
EricBAndrews: *nit* Is it nicer to have all these in one clause since they all have the same `else`? Ditto other actions below

EricBAndrews (approved):
LGTM! 🔥

Mlem/App/Actions/AppointAdminAction.swift
Sjmarf: Yep, updated 👍

PR #2483, opened 2025-12-07
Fix only top-level comments showing on Lemmy 1.0

Sjmarf:
On Lemmy 0.19, comments were sorted by depth ascending, which our logic relied upon. In Lemmy 1.0, they are sorted by depth descending.

### DIFF
--- a/Mlem/App/Views/Shared/ExpandedPost/CommentTreeTracker.swift
+++ b/Mlem/App/Views/Shared/ExpandedPost/CommentTreeTracker.swift
@@ -174,8 +174,10 @@ class CommentTreeTracker: Hashable {
         var output: [CommentTreeNode] = clear ? [] : nodes
         var commentsKeyedById: [Int: CommentTreeNode] = [:]
         var commentsKeyedByActorId: [ActorIdentifier: CommentTreeNode] = clear ? [:] : nodesKeyedByActorId
+
+        var sortedComments = newComments.sorted { $0.depth < $1.depth }
         
-        for comment in await self.sortComments(newComments) {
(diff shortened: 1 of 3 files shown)


Mlem/App/Views/Shared/ExpandedPost/CommentTreeTracker.swift
```diff
@@ -201,6 +194,14 @@ class CommentTreeTracker: Hashable {
         nodes = output
         nodesKeyedByActorId = commentsKeyedByActorId
     }
+
+    private func sortComments(_ comments: [Comment2]) async -> any Sequence<Comment2> {
+        guard let first = comments.first else { return [] }
+        if (try? await first.api.supports(.commentTreeSortedByDepthAscending)) ?? false {
+            return comments
+        }
+        return comments.sorted { $0.depth < $1.depth }
```
EricBAndrews: Is it worth tracking and checking `.commentTreeSortedByDepthAscending` instead of just always sorting? `comments.sorted { $0.depth < $1.depth }` should be trivially fast if they're already sorted (Swift uses introsort, which on sorted arrays is O(n)). It's technically a couple more cycles, but in return we never need to worry about how the API gives us the comments and we cut out a `guard` and make the function synchronous.

EricBAndrews (approved):
LGTM! 💯

Mlem/App/Views/Shared/ExpandedPost/CommentTreeTracker.swift
Sjmarf: Yeah good point, that's much nicer 😄

PR #2488, opened 2025-12-12
Export comment thread

### LINKED ISSUE #147: Share Posts and Comments as Image

EricBAndrews:
## Issues

- closes #147 

## Description

Adds threading support to the comment export tool.


Mlem/App/Views/Shared/ExportableViews/ExportableCommentEditorView.swift
```diff
@@ -36,56 +51,72 @@ struct ExportableCommentEditorView: View {
...
+                            Button {
+                                assert(threadLength > 1, "Cannot decrease thread length below 1")
+                                threadLength -= 1
+                            } label: {
+                                Image(icon: .general.remove)
                             }
-                            .menuActionDismissBehavior(.disabled) // this doesn't work but I think that's a bug
+                            .disabled(threadLength == 1)
+                            
+                            Text("\(threadLength - 1)")
```
Sjmarf:
```suggestion
                            Text(verbatim: "\(threadLength - 1)")
```

Mlem/App/Views/Shared/ExportableViews/ExportableCommentLoader.swift
```diff
@@ -0,0 +1,65 @@
...
+                assertionFailure("Could not cast to Post2Providing post-upgrade")
+                throw ApiClientError.unsuccessful
+            }
+            
+            Task { @MainActor in
+                self.data = .init(comments: comments, post: post)
+            }
+        } catch {
+            handleError(error)
+            self.error = error
```
Sjmarf: There's a `handleError` variant designed to handle `ErrorDetails` that we could use here. I don't remember why we have that 🤔 it might be important

Mlem/App/Views/Shared/ExportableViews/ExportableCommentLoader.swift
```diff
@@ -0,0 +1,65 @@
...
+        }
+    }
+}
+
+struct ExportableCommentData {
+    let comments: [any Comment2Providing]
+    let post: any Post3Providing
+    
+    var numComments: Int { comments.count }
```
Sjmarf: We should call this `commentCount` IMO, or just use `comments.count`.

Mlem/App/Views/Shared/ExportableViews/ExportableCommentEditorView.swift
```diff
@@ -36,56 +51,72 @@ struct ExportableCommentEditorView: View {
...
+                            
+                            Button {
+                                assert(
+                                    threadLength < min(8, data.numComments),
+                                    "Cannot increase thread length beyond \(min(8, data.numComments))"
+                                )
+                                threadLength += 1
+                            } label: {
+                                Image(icon: .general.add)
+                            }
```
Sjmarf: We should provide a title and hide it with `.labelStyle`, which is better for accessibility

```suggestion
                            Button("Add Comment", icon: .general.add) {
                                assert(
                                    threadLength < min(8, data.numComments),
                                    "Cannot increase thread length beyond \(min(8, data.numComments))"
                                )
                                threadLength += 1
                            }
                            .labelStyle(.iconOnly)
```

Sjmarf (changes requested):

Mlem/App/Views/Shared/ExportableViews/ExportableCommentLoader.swift
EricBAndrews: Ooh, I'd totally forgotten about that variant. IIRC it's built for situations where the error is reflected in the view rather than via toast; I've updated this to use it.

Sjmarf (approved):
Nice, works perfectly 😄

PR #2489, opened 2025-12-12
Use yes/no for votes settings in instance details

### LINKED ISSUE #2473: Use "yes" and "no" labels for upvote/downvote settings in `InstanceView` details tab
We currently use "enabled"/"disabled"; we should be consistent with the other settings

EricBAndrews:
closes #2473


Sjmarf (approved):
Nice 👍

PR #2490, opened 2025-12-12
Update Settings headers to match iOS format

### LINKED ISSUE #2456: Update Settings Header Stylng
### Requirements

### Description

iOS 26 uses left alignment instead of centering for many design elements, including settings headers. We should update our styling to match.

EricBAndrews:
## Issues

- closes #2456 

## Description

- Aligns headers to the leading edge
- Updates header detail text to `.body`/`.themedSecondary`

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Settings/Components/SettingsHeaderView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/Components/SettingsHeaderView.swift
@@ -29,7 +29,7 @@ struct SettingsHeaderView<IconView: View>: View {
     
     var body: some View {
         Section {
-            VStack(spacing: 12) {
+            VStack(alignment: .leading, spacing: 12) {
                 icon
                     .symbolVariant(.fill)
                     .padding(.bottom, 5)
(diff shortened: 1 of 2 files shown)


Mlem/App/Views/Root/Tabs/Settings/Components/SettingsHeaderView.swift
```diff
@@ -29,7 +29,7 @@ struct SettingsHeaderView<IconView: View>: View {
     
     var body: some View {
         Section {
-            VStack(spacing: 12) {
+            VStack(alignment: .leading, spacing: 12) {
                 icon
                     .symbolVariant(.fill)
                     .padding(.bottom, 5)
```
Sjmarf: *nit* I think this spacing is a little closer to iOS settings

```suggestion
            VStack(alignment: .leading, spacing: 4) {
icon
.symbolVariant(.fill)
.padding(.bottom, 11)
```

Sjmarf (approved):
Nice 😄

Mlem/App/Views/Root/Tabs/Settings/Components/SettingsHeaderView.swift
EricBAndrews: Yeah, that is closer 👍

PR #2492, opened 2025-12-14
Add user notes

Sjmarf:
Requires PieFed 1.4 or later (test using crust.piefed.social).

At the moment notes are only shown in `PersonView`; in future they will be shown in the feed too.


EricBAndrews (approved):
LGTM! 🔥 

Maybe the context menu option should read "Add Note" instead of "Edit Note" if no note is present?

Sjmarf:
Good idea 👍

PR #2496, opened 2025-12-22
Support PieFed in the test inbox

### LINKED ISSUE #2339: Create PieFed mappings for `InboxNotification`
At the time of writing only Lemmy < 1.0 has API Type -> `InboxNotificationSnapshot` mappings in place.

Ideally we'd use PieFed's fancier notification system here, rather than the Lemmy-adjacent one. However, that's currently impossible pending [this PieFed issue](https://codeberg.org/rimu/pyfedi/issues/1368) being completed. If Lemmy 1.0 is on the horizon and that issue isn't fixed, we should implement support for the Lemmy-adjacent API so we can get the new inbox system in.

Sjmarf:
Closes #2339

I'm considering rewriting the current inbox feed loader to use `InboxNotification` soon-ish. It looks relatively easy to do. It may need rewriting again to support the Lemmy 1.0 code, but we've got too many things in a half-finished state and I'd like to get this new system into production.

### DIFF
--- a/Mlem/App/Views/Shared/TestInboxView.swift
+++ b/Mlem/App/Views/Shared/TestInboxView.swift
@@ -86,11 +86,11 @@ private struct TestInboxSectionView: View {
         do {
             switch type {
             case .reply:
-                notifications = try await appState.firstApi.getReplyNotifications()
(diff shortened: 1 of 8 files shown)


EricBAndrews (approved):
LGTM! 💯

Issue #2435, opened 2025-11-15
Tier-Abstracted Content Models


EricBAndrews:
### Requirements

- [x] There are no existing issues for this feature.
- [x] This is a request for a **single** feature (create multiple issues for multiple feature requests).

### Description

Create unified models that abstract tiers altogether.

Summarizing Slack design discussion:

`AnyPost` contains a `PostPropertiesProviding`, a new protocol that defines the set of properties a post can have. Software-specific providers (e.g., `LemmyPostPropertiesProviding`, `PiefedPostPropertiesProviding`) conform to the protocol and internally contain raw snapshots. Accessing a property from a provider returns `.present(property)`, `.absent(property)` (indicating upgrade required), or `.unavailable` (indicating no API support). If `.absent` is returned from the provider, `AnyPost` fetches the higher tier data before returning the property.

This has the consequence that every property would be `async`, which would require us to implement more placeholder views, but in return tier becomes completely irrelevant. Any post function would be available from anywhere the post is present, we only need to use the single model throughout the whole codebase, and writing generic classes (e.g., `UpdateQueue`s) becomes much simpler. We can also initialize models with no data (effectively a `PostStub`), on which any property access triggers an upgrade.

EricBAndrews:
See also https://github.com/mlemgroup/mlem/issues/1154

PR #2499, opened 2025-12-23
Unified Post Model

EricBAndrews:
> [!NOTE]
> This PR is huge, since there is no nice way to incrementally replace these models. A vast majority of the line count here is name replacements or trivial compatibility changes (and a lot of commented out mocks). I have placed comments in every file with meaningful code changes outlining what those changes are; files without comments can generally be assumed to contain only trivial alterations.

- Progress towards #2435 

## Architecture

Replaces the existing tiered post structure with a new, unified architecture. This is built on the following two constructs:

`PostProperties`
This struct holds all the actual properties. Any properties that are not present in the lowest tier snapshot (`Post1Snapshot`) are optional; such properties which were already optional are now double optionals, where nil/nil indicates the value is not resolved and present/nil indicates the value resolved to nil. Double optionals are somewhat awkward, but I believe the most canonical way to represent unresolved nullables. (note: double optionals are not currently in use, as all higher-tier post snapshots return exclusively non-optional information; however, that is not guaranteed to always hold)

`ExpectedValue`
This class handles the on-demand value fetching. It must be provided a function that, when called, populates the underlying `PostProperties` with the relevant data. `Post.expectedValue` makes this somewhat more ergonomic.

`Post` stores a `PostProperties` and provides access via either passthroughs or `ExpectedValue`s as appropriate.

In cases where functions rely on potentially absent data, the publicly exposed function is a computed var that returns `nil` if the requisite information is not present. In checking for that data, it triggers an upgrade operation. UI elements that rely on these functions should therefore simply conditionally display the element on whether this function exists, and once the relevant data appears the element will be shown as well.

The new `ExpectedView` struct is built to streamline displaying `ExpectedValue`s. It takes an `ExpectedValue<T>`, a closure to render the view when that value is present, and an optional placeholder closure.

## Naming Conventions

There is some minor friction between the new system and the old, especially in protocols where previously a concrete value was available but now only an `ExpectedValue` is present. I have used the following two prefixes to denote certain types of interim code:
- Shim: Anything starting with "Shim" (e.g., `ShimVotable`) is a temporary workaround that I expect to be completely removed once all the tier-abstracted models are implemented
- Unified: anything starting with "Unified" (e.g., `UnifiedUpdateQueue`) is a new implementation of an existing system that I expect to supersede the old system entirely, but can be developed and used in parallel with the existing system.

## Outstanding Work

For the sake of clearing up the logjam this PR is causing, I have omitted some minor tasks:
- Mocking (this should wait until all the new models are in)
- Small UI polish

These areas are annotated with `// TODO: NOW` or `// TODO: updated mocks`, and will be addressed in follow-up PRs.


Mlem/App/Views/Dev/ExpectedViews/ExpectedTextView.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+    
+    var body: some View {
+        ZStack { // ZStack to make the animation work correctliy
+            if let text = text.value {
+                Text(text)
+                    .transition(.scale)
+            } else {
+                MockTextView()
+                    .frame(height: font.leadingHeight)
```
Sjmarf: If this is a permanent view we'll need some sort of `@ScaledMetric` in here to account for dynamic type.

Another option might be to use `redacted(reason: )` ([Docs](https://developer.apple.com/documentation/swiftui/view/redacted(reason:))) on some placeholder text. I've not used that API before, though, so don't know how it looks.

Yet another solution would be to `.overlay` some placeholder text with a rectangle

Sjmarf (commented):
Looks good 😄 This'll be much better than what we have now 👀

Mlem/App/Actions/BanAction.swift
```diff
@@ -72,7 +72,12 @@ extension ActionSeed {
     static let banCreator = ActionSeed("banCreator") { entity in
         switch entity {
         case let entity as any Comment2Providing: BanAction(entity: entity.creator)
-        case let entity as any Post2Providing: BanAction(entity: entity.creator)
+        case let entity as Post:
```
EricBAndrews, line 75: Some actions now require an additional check for the presence of the relevant data; they all follow this pattern.

Mlem/App/Actions/HideAction.swift
```diff
@@ -33,10 +33,11 @@ extension HideAction {
     static var label: ActionLabel { hideLabel }
 
     func createLabel(environment: EnvironmentValues) -> ActionLabel {
-        if entity.hidden {
-            Self.showLabel.withVisibility(visibility(environment))
+        guard let hidden = entity.hidden.value else { return Self.showLabel.withVisibility(.hidden) }
```
EricBAndrews, line 36: Likewise, if an action label requires an `ExpectedValue` that hasn't populated yet, it is hidden.

Mlem/App/Actions/HideAction.swift
```diff
@@ -55,9 +56,10 @@ extension HideAction {
 extension HideAction {
     @MainActor
     func execute(environment: EnvironmentValues) {
-        entity.toggleHidden()
+        guard let hidden = entity.hidden.value, let toggleHidden = entity.toggleHidden else { return }
```
EricBAndrews, line 59: And if the action somehow gets executed without the required data, nothing happens

Mlem/App/Utility/Extensions/Content Models/Interactable1Providing+Extensions.swift
```diff
@@ -161,13 +161,14 @@ extension Interactable1Providing {
         )
     }
     
-    func replyAction(appState: AppState, commentTreeTracker: CommentTreeTracker? = nil) -> BasicAction {
-        .init(
+    func replyAction(appState: AppState, commentTreeTracker: CommentTreeTracker? = nil) -> BasicAction? {
+        guard api.canInteract(appState: appState) else { return nil }
```
EricBAndrews: It seems cleaner to simply not show an action that will not do anything

Mlem/App/Utility/Extensions/Content Models/Post+Extensions.swift
```diff
@@ -0,0 +1,56 @@
+//
+//  Post+Extensions.swift
```
EricBAndrews, line 2: Contains convenience extensions for building UI elements. Mostly copy/paste from the old system.

Mlem/App/Utility/Extensions/Content Models/Post+Interactions.swift
```diff
@@ -0,0 +1,174 @@
+//
+//  Post+Interactions.swift
```
EricBAndrews, line 2: This file contains all the wrapper functions around the primitive interactions defined in `Post`; the functions in here take in feedback, callbacks, offer toggle functionality, etc.

Mlem/App/Utility/Extensions/Content Models/Post+Interactions.swift
```diff
@@ -0,0 +1,174 @@
...
+
+extension Post: ShimVotable, ShimSaveable {
+    public var shimToggleSaved: (() -> Void)? {
+        if let toggleSaved {
+            return { toggleSaved([]) }
+        }
+        return nil
+    }
+    
+    public var toggleVote: ((ScoringOperation) -> Void)? {
```
EricBAndrews, line 24: This is a canonical example of how functions that rely on potentially absent data are handled under the new system

Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/Feed Post Components/CrossPostListView.swift
```diff
@@ -14,12 +14,19 @@ struct CrossPostListView: View {
...
     @Environment(NavigationLayer.self) private var navigation
     
-    let post: any Post3Providing
+    let post: Post
     
     @State private var isExpanded: Bool = false
     
     var body: some View {
-        if !post.crossPosts.isEmpty {
+        ExpectedView(post.crossPosts) { crossPosts in
```
EricBAndrews: Updated to `ExpectedValue` for crossposts, passes the value down the view tree once resolved

Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/CompactPostView.swift
```diff
@@ -61,22 +61,31 @@ struct CompactPostView: View {
...
+                        }
+                        
                     }
                     Spacer()
-
-                    if differentiateWithoutColor, readPostIndicator == .checkmark, post.read_ ?? false {
-                        ReadCheck()
+                    
+                    if differentiateWithoutColor, readPostIndicator == .checkmark {
+                        ReadCheck(read: post.read)
```
EricBAndrews, line 80: Note that this now takes an `ExpectedValue`

Mlem/App/Views/Shared/ExpandedPost/CommentStubResolutionPage.swift
```diff
@@ -0,0 +1,51 @@
+//
+//  CommentStubResolutionPage.swift
```
EricBAndrews, line 2: Serves as the navigation destination for comment stubs. Performs the resolution work and redirects to `CommentPage` once resolved.

This works, but the replace is somewhat jarring. I will investigate a smoother technique in a future PR.

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView+Logic.swift
```diff
@@ -59,22 +59,17 @@ extension ExpandedPostView {
...
             return tracker?.nodesKeyedByActorId.count == 0
         }
-        return (post?.commentCount_ ?? -1) == 0
+        return (post.commentCount.value ?? -1) == 0
     }
     
     var showLoadingSymbol: Bool {
         // Don't need to show ProgressView if there's nothing to scroll to
         if scrollTargetedComment == nil { return false }
-
```
EricBAndrews, line 68: Removed `ContentLoader` code

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView.swift
```diff
@@ -47,41 +45,32 @@ struct ExpandedPostView<Content: View>: View {
...
         // be a delay between the `content()` appearing and calling `scrollTo`
         VStack {
-            if let post = post as? any Post {
-                content(post: post, isLoading: isLoading)
-                    .externalApiWarning(entity: post, isLoading: isLoading)
-                    .task(id: tracker == nil) {
-                        if let tracker, post.api == appState.firstApi, tracker.loadingState == .idle {
-                            post.markRead()
-                        }
+            viewContent
```
EricBAndrews, line 65: Removed `ContentLoader`-associated loading statuses

Mlem/App/Views/Shared/ExpandedPost/PostStubResolutionPage.swift
```diff
@@ -0,0 +1,37 @@
+//
+//  PostStubResolutionPage.swift
```
EricBAndrews, line 2: Same as `CommentStubResolutionPage`

Mlem/App/Views/Shared/ExpectedViews/ExpectedTextView.swift
```diff
@@ -0,0 +1,33 @@
+//
+//  ExpectedTextView.swift
```
EricBAndrews, line 2: Convenience view for `ExpectedValue<String>`

Mlem/App/Views/Shared/ExpectedViews/ExpectedView.swift
```diff
@@ -0,0 +1,33 @@
+//
+//  ExpectedView.swift
```
EricBAndrews, line 2: View to display an `ExpectedValue` or a placeholder if that value is not resolved, and animate nicely between them.

Mlem/App/Views/Shared/ExpectedViews/String+Placeholders.swift
```diff
@@ -0,0 +1,11 @@
+//
+//  String+Placeholders.swift
```
EricBAndrews, line 2: Standard strings to use as placeholders

Mlem/App/Views/Shared/ExportableViews/ExportableCommentLoader.swift
```diff
@@ -38,13 +38,8 @@ class ExportableCommentLoader {
                 comments = [comment]
             }
             
-            guard let post = try await comment.post.upgrade() as? any Post3Providing else {
```
EricBAndrews, line 41: No longer necessary

Mlem/App/Views/Shared/Navigation/NavigationPage+View.swift
```diff
@@ -84,15 +84,29 @@ extension NavigationPage {
...
-            PostPage(post: post, scrollTargetedComment: scrollTargetedComment?.wrappedValue)
-                .environment(\.communityContext, communityContext?.wrappedValue)
-        case let .comment(comment, comments: comments, showViewPostButton, exposeRemovedContent):
+            // TODO: NOW better way of doing this embed
+            ExpandedPostView(post: post, tracker: nil, scrollTargetedComment: scrollTargetedComment?.wrappedValue) {
+                CrossPostListView(post: post)
+                    .padding(.horizontal, Constants.main.standardSpacing)
+            }
+            .environment(\.communityContext, communityContext?.wrappedValue)
+        case let .postStub(post, _):
```
EricBAndrews, line 93: Adds this and `.commentStub` as navigation targets for when only a stub is available

Mlem/App/Views/Shared/Navigation/NavigationPage.swift
```diff
@@ -355,10 +378,6 @@ enum NavigationPage: Hashable {
         advancedSorting(.init(wrappedValue: sort))
     }
     
-    static func createPostImage(_ post: any Post) -> NavigationPage {
```
EricBAndrews, line 358: `HashWrapper` is no longer needed so neither is this function

Mlem/App/Views/Shared/Navigation/NavigationPage+View.swift
```diff
@@ -84,15 +84,29 @@ extension NavigationPage {
...
+            PostStubResolutionPage(stub: post.wrappedValue)
+        case let .comment(comment, post, comments, showViewPostButton, exposeRemovedContent):
             CommentPage(
-                comment: comment,
+                comment: comment.wrappedValue,
+                post: post,
```
EricBAndrews, line 98: Note that post now needs to be passed in; this is temporary until the unified comment model is created

Mlem/App/Views/Shared/HandleThreadiverseLinksModifier.swift
```diff
@@ -162,24 +162,24 @@ struct HandleThreadiverseLinksModifier: ViewModifier {
             // https://piefed.social/c/politics/p/1385905/will-the-supreme-court-hand-government-contractors-blanket-immunity
             if components.count > 4, components[2] == "p" {
                 let newUrl = url.removingPathComponents().appendingPathComponent("post/\(components[3])")
-                return .post(PostStub(api: appState.firstApi, url: newUrl))
+                return .postStub(PostStub(api: appState.firstApi, url: newUrl))
```
EricBAndrews, line 165: Update these to point to the new stub pages

Mlem/App/Views/Shared/ReadCheck.swift
```diff
@@ -8,19 +8,33 @@
 import Icons
 import SwiftUI
 import Theming
+import MlemMiddleware
 
 struct ReadCheck: View {
     let dimension: CGFloat
+    let read: ExpectedValue<Bool>
```
EricBAndrews, line 15: This is now basically an `ExpectedView` wrapper

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/Caching/Caches/PostCache.swift
```diff
@@ -0,0 +1,57 @@
+//
+//  PostCache.swift
+//  MlemMiddleware
+//
+//  Created by Eric Andrews on 2026-01-03.
+//
+
+public enum AnyPostSnapshot: CacheIdentifiable {
```
EricBAndrews, line 8: This unified snapshot makes handling types in the generic update queue very easy

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/Caching/Caches/PostCache.swift
```diff
@@ -0,0 +1,57 @@
...
+    public var cacheId: Int {
+        switch self {
+        case let .post1(snapshot): snapshot.cacheId
+        case let .post2(snapshot): snapshot.cacheId
+        case let .post3(snapshot): snapshot.cacheId
+        }
+    }
+}
+
+class PostCache: ApiTypeBackedCache<Post, AnyPostSnapshot> {
```
EricBAndrews, line 22: Normal cache for the new posts

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Community/Community2Providing+Snapshots.swift
```diff
@@ -8,10 +8,10 @@
 extension Community2Providing {
     internal func takeSnapshot2() -> Community2Snapshot {
         .init(community: community1.takeSnapshot1(),
-              subscription: .init(total: subscriberCount, local: localSubscriberCount, subscribed: subscribed, pending: false), // TODO: NOW pending?
```
EricBAndrews, line 11: These TODOs will be irrelevant with unified community (hopefully)

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Interactable/Interactable2Providing.swift
```diff
@@ -8,7 +8,14 @@
 import Foundation
 
 // Content that can be upvoted, downvoted, saved etc
-public protocol Interactable2Providing: Interactable1Providing, RemovableProviding, PurgableProviding, CanModerateProviding {
+public protocol Interactable2Providing:
+    Interactable1Providing,
+    RemovableProviding,
+    PurgableProviding,
+    CanModerateProviding,
+    ShimVotable,
```
EricBAndrews: Added shim protocols that use an `ExpectedValue` instead of a direct one for specific relevant fields and functions

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Post/Post+Conformance.swift
```diff
@@ -0,0 +1,235 @@
+//
+//  Post+Conformance.swift
```
EricBAndrews, line 2: All the protocol conformance code goes here, mostly copy/paste from previous system.

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Post/PostProperties.swift
```diff
@@ -0,0 +1,147 @@
+//
+//  PostProperties.swift
```
EricBAndrews, line 2: Contains all properties that come from the API

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Report/ReportTarget.swift
```diff
@@ -34,25 +34,32 @@ public enum ReportTarget {
...
     }
     
-    public var creator: Person1 {
+    // TODO: UnifiedCommentModel, UnifiedMessageModel remove this shim
+    public var creator: ExpectedValue<any Person1Providing> {
         switch self {
         case let .post(post): post.creator
-        case let .comment(comment): comment.creator
-        case let .message(message): message.creator
+        case let .comment(comment): .init(
```
EricBAndrews, line 47: Added shims to convert to `ExpectedValue`

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Shared/UnifiedModelProviding.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  UnifiedModelProviding.swift
+//  MlemMiddleware
+//
+//  Created by Eric Andrews on 2026-01-07.
+//
+
+public protocol UnifiedModelProviding: AnyObject, CacheIdentifiable, ContentModel {
```
EricBAndrews, line 8: Primarily used to make update queue generic (along with `UnifiedPropertiesProviding` below)

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/UpdateQueues/UnifiedUpdateQueue.swift
```diff
@@ -0,0 +1,122 @@
+//
+//  UnifiedUpdateQueue.swift
```
EricBAndrews, line 2: Fully generic update queue. All the actual queuing logic is the same, but there are a couple key changes:

- Upgrade is now a normal queued operation since we don't need to swap the actual object out. However, it still gets its own function because it has some special queuing requirements.
- `UpdateTask` now works with properties instead of snapshots

Sjmarf:
> Mocking (this should wait until all the new models are in)

Yeah don't worry about this, the entire system is broken anyway (#2279). It need a major rewrite at some point

Mlem/App/Views/Shared/InteractionBar/InteractionBarView.swift
Sjmarf, line 1: The upvote/downvote counter isn't displaying a number for me, and can't be interacted with.

Image

Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/Feed Post Components/CrossPostListView.swift
Sjmarf, line 1: When a post has no crossposts, there's too much space between the post and the first comment

Mlem/App/Utility/Extensions/Content Models/Post+Actions.swift
```diff
@@ -551,26 +232,247 @@ extension Post1Providing {
...
+            ) {
+                moderatorMenuActions(
+                    appState: appState,
+                    feedback: feedback,
+                    showAllActions: showAllActions,
+                    navigation: navigation,
+                    report: report
+                )
+            }
+        }
```
Sjmarf, line 397: Moderator actions aren't shown in the post context menu for a community I moderate. They show correctly in the ellipsis menu (which uses the new action system).

Mlem/App/Utility/Extensions/Content Models/Interactable1Providing+Extensions.swift
Sjmarf: This causes strange behaviour in guest mode

Image

I'm on board with hiding useless actions, but maybe we should leave that OOS and solve it when the full-sized action sheet gets put in (like we've got in the test inbox)?

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/FeedLoaders/Filtering/Filters/Post/PostKeywordFilter.swift
```diff
@@ -7,10 +7,15 @@
 
 import Foundation
 
-class PostKeywordFilter: FilterProviding<Post2> {
-    override public func shouldPassFilter(_ post: Post2) -> Bool {
+class PostKeywordFilter: FilterProviding<Post> {
+    override public func shouldPassFilter(_ post: Post) -> Bool {
+        // community should always exist for posts going through the feed loader
+        guard let community = post.community.value_ else {
+            assertionFailure("No community found in filter-eligible post")
```
Sjmarf, line 14: I consistently hit this `assertionFailure` when opening community links. I don't have any keyword filters

```
MlemMiddleware/PostKeywordFilter.swift:14: Fatal error: No community found in filter-eligible post
```

Mlem/App/Views/Shared/InteractionBar/InteractionBarView.swift
EricBAndrews: Nice catch! I've fixed this with a new `ShimInteractable2Providing`, which bridges `Post` and the existing `Interactable2Providing`. There is some duplicated logic now between `Post` and `ShimInteractable2Providing`, but once the other interactables are converted to the unified system it should all collapse neatly down into a unified protocol.

Mlem/App/Utility/Extensions/Content Models/Post+Actions.swift
EricBAndrews: Whoops, messed up the clause structure in `moderatorMenuActions`. Should be fixed now.

Sjmarf: Now it shows, but "ban user" is disabled

IMG_2273

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView+Logic.swift
Sjmarf, line 1: The app hangs for a second when I open this post from the feed. https://lemmy.ml/post/41476191

Mlem/App/Views/Shared/InteractionBar/InteractionBarView.swift
Sjmarf: Nice, works great 👍

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/FeedLoaders/Filtering/Filters/Post/PostKeywordFilter.swift
EricBAndrews: This was a sneaky one! It triggered because the community was loaded in an async task due to the update going through the queue, and therefore was sometimes absent when the filter got to the post. I've fixed this by adding a synchronous "soft update" that bypasses the queue and only populates _new_ properties; this gets called on `update`. I have not been able to trigger the crash since this change.

Sjmarf: Nice. It hasn't occurred for me yet either, good job 👍

Mlem/App/Views/Shared/ExpandedPost/CommentStubResolutionPage.swift
Sjmarf, line 1: We need a `.themedGroupedBackground()` in here somewhere.

Mlem/App/Views/Shared/ExpandedPost/CommentStubResolutionPage.swift
```diff
@@ -0,0 +1,51 @@
...
+    let stub: any CommentStubProviding
+    let comments: [Comment2]?
+    let showViewPostButton: Bool
+    let exposeRemovedContent: Bool
+    
+    @State var upgradeError: Error?
+    
+    var body: some View {
+        if let upgradeError {
+            ErrorView(.init(error: upgradeError))
```
Sjmarf: We should probably hook this up with a `refresh:` callback

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView+Logic.swift
Sjmarf: Can't repro this anymore today. It used to happen every single time for that post, and not for any other posts 🤔

Will mention if I see it again; happy to resolve this for now

Mlem/App/Views/Shared/ExpandedPost/PostStubResolutionPage.swift
```diff
@@ -0,0 +1,37 @@
...
+struct PostStubResolutionPage: View {
+    @Environment(NavigationLayer.self) var navigation
+    
+    let stub: any PostStubProviding
+    
+    @State var upgradeError: Error?
+    
+    var body: some View {
+        if let upgradeError {
+            ErrorView(.init(error: upgradeError))
```
Sjmarf: Ditto above comment about adding `refresh:` here

Mlem/App/Views/Shared/ExpandedPost/PostStubResolutionPage.swift
Sjmarf, line 1: Ditto above comment about `.themedGroupedBackground()`

Mlem/App/Views/Shared/ExpectedViews/ExpectedView.swift
```diff
@@ -0,0 +1,33 @@
...
+
+import SwiftUI
+import MlemMiddleware
+
+struct ExpectedView<Value, Content: View, Placeholder: View>: View {
+    let value: ExpectedValue<Value>
+    @ViewBuilder let view: (Value) -> Content
+    @ViewBuilder let placeholder: (() -> Placeholder)
+    
+    init(_ value: ExpectedValue<Value>, view: @escaping (Value) -> Content, placeholder: @escaping (() -> Placeholder) = { EmptyView() }) {
```
Sjmarf: _nit_ maybe we should multi-line this for readability


```suggestion
    init(
        _ value: ExpectedValue<Value>, 
        view: @escaping (Value) -> Content,
        placeholder: @escaping () -> Placeholder = { EmptyView() }
    ) {
```

Mlem/App/Utility/Extensions/Content Models/Post+Actions.swift
Sjmarf: Can no longer repro this

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView+Logic.swift
EricBAndrews: I think there are some really negative performance implications to the way property updates currently work (i.e., replacing the whole struct every time). I have a follow-up planned to play around with optimizations (e.g., `setIfChanged`), but in the interest of getting the big picture architecture working the brutalist approach works well enough for me for now

Mlem/App/Utility/Extensions/Content Models/Interactable1Providing+Extensions.swift
EricBAndrews: Fair point, works for me 👍 

I've reverted the standard actions to the previous behavior. Mod actions I decided to keep hidden unless available, since we only display them if we've already checked upstream that they're available so there's no situation I can think of in which we would even want the disabled mod action

Mlem/App/Utility/Extensions/Content Models/Post+Actions.swift
EricBAndrews: This was caused by the post creator's community ban status not being updated. That status is now set any time a `PostProperties` is initialized or updated and the relevant data is available

Mlem/App/Views/Shared/ExpandedPost/CommentStubResolutionPage.swift
```diff
@@ -0,0 +1,66 @@
...
+        content
+            .themedGroupedBackground()
+    }
+    
+    @ViewBuilder
+    var content: some View {
+        if let upgradeError {
+            ErrorView(.init(
+                error: upgradeError,
+                        refresh: fetchComment))
```
Sjmarf: _nit_


```suggestion
            ErrorView(.init(
                error: upgradeError,
                refresh: fetchComment
            ))
```

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView.swift
Sjmarf: Technically OOS: this should be `groupedBackground`

Mlem/App/Views/Shared/ExpandedPost/PostStubResolutionPage.swift
```diff
@@ -0,0 +1,52 @@
...
+        content
+            .themedGroupedBackground()
+    }
+    
+    @ViewBuilder
+    var content: some View {
+        if let upgradeError {
+            ErrorView(.init(
+                error: upgradeError,
+                refresh: fetchPost))
```
Sjmarf: _nit_


```suggestion
            ErrorView(.init(
                error: upgradeError,
                refresh: fetchPost
            ))
```

Mlem/App/Views/Shared/ExpectedViews/ExpectedView.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Created by Eric Andrews on 2025-12-28.
+//
+
+import SwiftUI
+import MlemMiddleware
+
+struct ExpectedView<Value, Content: View, Placeholder: View>: View {
+    let value: ExpectedValue<Value>
+    @ViewBuilder let view: (Value) -> Content
+    @ViewBuilder let placeholder: (() -> Placeholder)
```
Sjmarf: _nit_

```suggestion
    @ViewBuilder let placeholder: () -> Placeholder
```

Mlem/App/Views/Shared/Navigation/NavigationLayer.swift
```diff
@@ -56,6 +56,19 @@ class NavigationLayer: Identifiable {
...
     }
+    
+    /// Replaces the current top level page with the given NavigationPage. Useful for redirecting from interim
+    /// pages that should not appear in navigation history.
+    @MainActor
+    func replace(_ page: NavigationPage) {
+        if hasNavigationStack {
+            // This prevents keyboard animation glitches when navigating whilst the keyboard is open
+            UIApplication.shared.firstKeyWindow?.endEditing(true)
+            path[path.count - 1] = page
```
Sjmarf: The root page is stored in a separate `root` variable, and all other pages are stored in `path`. Therefore this logic doesn't work with the root page. I don't think this is a problem in reality but thought I'd mention it

Mlem/App/Views/Shared/Navigation/NavigationPage.swift
```diff
@@ -24,12 +24,28 @@ enum NavigationPage: Hashable {
...
     case post(
-        _ post: AnyPost,
+        _ post: Post,
         scrollTargetedComment: HashWrapper<any CommentStubProviding>? = nil,
         communityContext: HashWrapper<any Community1Providing>? = nil,
         navigationNamespace: Namespace.ID? = nil
     )
-    case comment(_ comment: AnyComment, comments: [Comment2]?, showViewPostButton: Bool, exposeRemovedContent: Bool)
+    case postStub(
+        _ post: HashWrapper<any PostStubProviding>,
```
Sjmarf: What about:

```suggestion
        _ post: PostStub
```

Could we also remove `PostStubProviding` altogether?

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Post/Post+Conformance.swift
```diff
@@ -0,0 +1,237 @@
...
+// MARK: CacheIdentifiable
+
+public extension Post {
+    var cacheId: Int { id }
+}
+
+// MARK: ContentModel
+
+public extension Post {
+    static var tierNumber: Int =  4
```
Sjmarf:
```suggestion
    static var tierNumber: Int = 4
```

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Post/Post.swift
```diff
@@ -0,0 +1,477 @@
...
+    lazy var updateQueue: UnifiedUpdateQueue<Post> = .init(parent: self)
+    
+    public var api: ApiClient
+    public var properties: PostProperties {
+        didSet {
+            pinnedCommunityPending = false
+            pinnedInstancePending = false
+            lockedPending = false
+            nsfwPending = false
+            removedPending = false
```
Sjmarf, line 56: Using `setIfChanged` here could result in a significant speed-up

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Post/PostProperties.swift
```diff
@@ -0,0 +1,191 @@
...
+//  MlemMiddleware
+//
+//  Created by Eric Andrews on 2026-01-07.
+//
+
+import Foundation
+
+public struct PostProperties: UnifiedPropertiesProviding {
```
Sjmarf, line 10: If we run into performance issues, this may need to be a class. Because this is a value type, changing an arbitrary property will cause a state update for all views that reference _any_ property of the post. Happy to leave for now though

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/UpdateQueues/UnifiedUpdateQueue.swift
Sjmarf: Nice, this is really cool 🎉

Sjmarf (changes requested):

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Post/PostProperties.swift
EricBAndrews: Yeah, that's my basic plan long term. I have #2526 to keep track of this

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Post/Post.swift
EricBAndrews: Tagging #2526

Mlem/App/Views/Shared/Navigation/NavigationPage.swift
EricBAndrews: So we can! Nice catch

Mlem/App/Views/Shared/Navigation/NavigationLayer.swift
EricBAndrews: Good point, I've updated it to replace root if the path is empty

Mlem/App/Views/Shared/ExpandedPost/ExpandedPostView.swift
EricBAndrews: Easy enough to fix 👍

Mlem/App/Views/Shared/Navigation/NavigationLayer.swift
```diff
@@ -64,7 +64,11 @@ class NavigationLayer: Identifiable {
         if hasNavigationStack {
             // This prevents keyboard animation glitches when navigating whilst the keyboard is open
             UIApplication.shared.firstKeyWindow?.endEditing(true)
-            path[path.count - 1] = page
+            if path.count == 0 {
```
Sjmarf: _nit_

```suggestion
            if path.isEmpty {
```

Sjmarf (approved):
LGTM 🚀  Great work! This will make things much easier.

PR #2501, opened 2025-12-23
New action system for posts

Sjmarf:
The new action system is now used in post ellipsis menus and in the `ExpandedPostView` toolbar. It is not yet used in post context menus.

Functional changes:
- Admins previously had separate "pin instance" and "pin community" buttons. These have now been merged into one, in the same way that we did for the "ban" action.
- The "create image" action is no longer shown for comments when they appear outside of `ExpandedPostView`


Sjmarf:
Holding off until 2.4 is out

EricBAndrews (approved):
LGTM! 🚀

PR #2502, opened 2025-12-24
Fix user who submitted report sometimes being shown as creator

Sjmarf:
This is a pretty big bug. Couldn't actually repro it myself, weirdly


EricBAndrews (approved):
Nice 👍

PR #2503, opened 2025-12-24
Resolve reports when separate mod menu enabled

EricBAndrews:
## Issues

- closes #2500 

## Description

Implements `ResolveAction` under the new action system and adds it to the separate moderator menu.

PR #2505, opened 2025-12-26
Fix resolve action only showing in ellipsis menu for posts, not context menu

Sjmarf:
This was also an issue in 1.3 I think; unlike #2503 which fixed something we broke in TF


EricBAndrews (approved):
Nice 👍

PR #2511, opened 2025-12-31
2.4.1

EricBAndrews:
Version bump, no changes


Sjmarf:
Should this be 2.5?

EricBAndrews:
If we need to release a hotfix it should be 2.4.1, so I don't want to bump all the way up to 2.5 until we have a full release ready--if we don't need any hotfixes we can always just bump it to 2.5 before release

Sjmarf (approved):
👍

PR #2514, opened 2026-01-01
Fix missing "edit" and "block community" actions for posts

Sjmarf:
Currently, blocking a community will result in a "failed to block" message. This is because an error occurs in the `ApiClient`. It was also broken before this PR, but there was no error message. I will fix it in a separate PR


EricBAndrews (approved):
LGTM! 👍

PR #2515, opened 2026-01-01
Fix blocking community on Lemmy (and update types)

### LINKED ISSUE #2513: Post doesn't disappear from feed when community is blocked
This was an issue before #2501

Sjmarf:
> [!note]
> See also [the MlemApiTypes PR](https://github.com/mlemgroup/MlemApiTypes/pull/45)

Closes #2513

There was a decoding error due to our schema not matching what was returned on Lemmy pre-1.0. The block would still take effect on the server, but the posts wouldn't disappear from the feed. We should put this in a patch release at some point.


EricBAndrews (approved):
LGTM! 🔥

PR #2518, opened 2026-01-07
Use `InboxNotification` in inbox tab

Sjmarf:
`InboxFeedLoader` now uses `InboxNotification`. No functional changes.

This means that `Reply` is (afaik) no longer used. I will remove the `Reply` models in an upcoming PR.


EricBAndrews (approved):
LGTM! 🔥

PR #2525, opened 2026-01-17
New action system for `Instance`

Sjmarf:
- Instances now use the new system
- The new action system is now used by the ellipsis menu in `MessageFeedView`
- Removed the old action system code for `Instance`
- Removed the old action system code for `Person`


Mlem/App/Actions/BlockAction.swift
```diff
@@ -159,50 +203,101 @@ extension BlockAction {
...
+    func execute(entity content: Content, environment: EnvironmentValues) {
+        if content.blocked(environment: environment) {
+            submit(content: content, environment: environment)
+            return 
+        }
+
+        let label: LocalizedStringResource = switch content {
+        case .blockable(_ as any Person): "Really block this user?"
+        case .blockable(_ as any Community): "Really block this community?"
+        default: "Really block this instance?"
```
EricBAndrews: *nit*

I think it's better to have a `case .instance` for this one, since the output text is instance-specific rather than a generic default

EricBAndrews (approved):
LGTM! 🔥

Issue #2331, opened 2025-09-27
Rewrite ItemCache using Actor


EricBAndrews:
### Requirements

- [x] There are no existing issues for this feature.
- [x] This is a request for a **single** feature (create multiple issues for multiple feature requests).

### Description

`ItemCache` is currently implemented using an `Atomic` to achieve thread safety, but we are getting memory crashes on `itemCache.put`. Rewriting this class as an `actor` may resolve this, and either way is more canonical Swift.

PR #2531, opened 2026-01-19
Unified Post Model Improvements

### LINKED ISSUE #2526: Unified Models: Performance
Improve performance of unified model system

EricBAndrews:
## Issues

- closes #2526 

## Description

Makes a couple key changes to the unified post models to improve performance and architectural cleanliness:

### Properties In Model

Properties are now stored directly on the model, rather than being completely wrapped in `PostProperties`. This significantly speeds up access and updates, though comes at the slight penalty of an awkward `init` due to `ExpectedValue` requiring access to `self`. `ExpectedValue` is simplified to take an initial value and an update function, rather than a value-providing getter; this also improves performance by removing a function call to the access stack.

### Streamlined Model Creation Pipeline

`PostProperties` remains useful, however, as an intermediate model between Snapshots and full Models. They serve multiple roles in the updated architecture:
- Represent a Model at a particular moment in time, providing convenience methods for data merging etc. that render them ergonomic for use in the `UpdateQueue` (and I think will also eventually be useful for multi-account stuff)
- Serve as a narrow waist for Model pre-initialization. Post can _only_ be initialized with a PostProperties. The Properties initializer now goes out to the caches to fetch the necessary Models to build the final Post, and performs on-init work such as updating creator ban state. This allows us to ensure that this work, if possible, is done for every post initialization, but separates it from the actual instantiation of the Post; once #2331 is done, this will allow us to move all the post pre-init off the main thread altogether, which should yield further performance gains.


Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Post/PostProperties.swift
```diff
@@ -48,92 +48,38 @@ public struct PostProperties: UnifiedPropertiesProviding {
...
-        
-        crossPosts = crossPosts ?? properties.crossPosts
-        
-        if let creator, let creatorBannedFromCommunity {
-            creator.person1.updateKnownCommunityBanState(id: communityId, banned: creatorBannedFromCommunity)
-        }
-    }
-    
     /// Constructs a PostProperties from a given snapshot
     /// - Note: External models (e.g., Creator) will NOT be included!
```
Sjmarf:
```suggestion
```

Sjmarf (approved):
Nice 💯

PR #2532, opened 2026-01-20
Remove `Message1.read` property

### LINKED ISSUE #2338: Remove `read` property from `Message2`

Sjmarf:
`read` is also stored on `InboxNotification`, so this property was redundant.

Closes #2338


EricBAndrews (approved):
LGTM! 💯

PR #2536, opened 2026-01-22
Disable send message button while message is being sent

### LINKED ISSUE #2535: Direct Message send button can be pressed multiple tomes
### Requirements

### Description

If the direct messages have to think about sending, the send button can still be pressed and when it eventually processes it sends the messages multiple times. 

### Mlem Version

2.4.1 (662)

Sjmarf:
Closes #2535

### DIFF
--- a/Mlem/App/Views/Pages/MessageFeedView/MessageFeedView+Logic.swift
+++ b/Mlem/App/Views/Pages/MessageFeedView/MessageFeedView+Logic.swift
@@ -16,8 +16,9 @@ extension MessageFeedView {
     }
     
     func sendMessage(_ scrollProxy: ScrollViewProxy) async {
+        self.isSending = true
+        defer { self.isSending = false }
         do {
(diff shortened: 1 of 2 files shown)


EricBAndrews (changes requested):
It looks like `isSending` doesn't get reset in some cases--if I perform the following:
- Send a message
- Immediately spam the button a couple times
Then I am locked out from sending further messages until I navigate away and back into the chat feed

Maybe we want to include a visual indication that the button is disabled while sending? It may feel broken if it stays blue. Also happy to leave that OOS to get a TF up quickly.

Sjmarf:
Thanks, fixed.

I've lowered the opacity of the button while sending. Not sure that we need a longer-term progress indicator (e.g. `ProgressView`) because sending messages is basically instant, in my experience

EricBAndrews (approved):
LGTM! 🚀

PR #2539, opened 2026-01-25
Unified Comment Model

### LINKED ISSUE #2527: Unified Models: Comment

EricBAndrews:
## Issues

- closes #2527 

## Description

Implements the unified content models for comments.

## Implementation Notes

Most of the changes here are either boilerplate or deleting old comment stuff. The noteworthy points:
- `ValueProviding`: this is a new protocol that `ExpectedValue` and the new `ResolvedValue` struct both conform to. There's a corner case where we might get an `ExpectedValue` or we might get a resolved value passed in, so this lets us treat them interchangeably.
- Removed `Post` and `Interactable2Providing` shim code. `Post` and `Comment` conform to `InteractableProviding` now.
- Light restructuring: all the `toggleStatus(feedback)` functions live in a `Model+Toggles` file
- Made properties that are never expected to change (e.g., `actorId`) `let` values instead of `var`s and excluded them from updates
- Created a separate `refresh()` method on `UnifiedUpdateQueue`. `upgrade()` can now only be called once; this prevents a rare bug where navigating to a post that loaded into feed and was then removed/deleted/purged caused an infinite `upgrade()` loop. In a future PR I will update `refresh()` to only return once the refresh is complete (rather than once it's queued) so the spinners will behave correctly.


Mlem/App/Utility/Extensions/Content Models/Interactable/InteractableProviding+Extensions.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+
+import MlemMiddleware
+
+// Utility extensions for InteractableProviding
+
+extension InteractableProviding {
+    var downvotesEnabled: Bool {
+        api.voteFederationMode.postDownvote != .disable
+    }
```
Sjmarf: There are separate `postDownvote` and `commentDownvote` values. Having this defined in `InteractableProviding` incorrectly uses `postDownvote` for both.

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Comment/Comment.swift
```diff
@@ -0,0 +1,390 @@
...
+        setIfNil(\.creator.value_, properties.creator)
+        setIfNil(\.post.value_, properties.post)
+        setIfNil(\.community.value_, properties.community)
+        setIfChanged(\.commentCount.value_, properties.commentCount)
+        setIfChanged(\.creatorIsModerator.value_, properties.creatorIsModerator)
+        setIfChanged(\.creatorIsAdmin.value_, properties.creatorIsAdmin)
+        setIfChanged(\.creatorBannedFromCommunity.value_, properties.creatorBannedFromCommunity)
+        setIfChanged(\.votes.value_, properties.votes)
+        setIfChanged(\.saved.value_, properties.saved)
+    }
```
Sjmarf, line 130: Something doesn't seem to be working properly - if I upvote or un-upvote a comment through the website and then refresh the page in Mlem, it doesn't show the update.

Sjmarf (changes requested):

Mlem/App/Utility/Extensions/Content Models/Interactable/InteractableProviding+Extensions.swift
EricBAndrews: Nice catch! I have moved the implementations of `downvotesEnabled` into `Post` and `Comment` to use the correct values.

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Comment/Comment.swift
EricBAndrews: Nice catch! The `softUpdate` technique I was using in the cache caused updates to not propagate when refreshing the post page. I've updated it to perform both a synchronous `softUpdate` (to ensure high-tier data always present if available) and a queued `attemptDirectUpdate` (to perform other updates, if the queue allows).

Mlem/App/Utility/Extensions/Content Models/Interactable/InteractableProviding+Extensions.swift
Sjmarf: Nice 👍

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Comment/Comment.swift
Sjmarf: Nice, works perfectly 👍

Sjmarf (approved):
LGTM 🚀 🚀

PR #2541, opened 2026-01-26
Polls (initial work)

Sjmarf:
Still to do:
- #2542
- #2543
- #2544


Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/PostPollView.swift
```diff
@@ -0,0 +1,123 @@
...
+                if showResults {
+                    // This creates a half-capsule
+                    UnevenRoundedRectangle(
+                        topLeadingRadius: 0,
+                        bottomLeadingRadius: 0,
+                        bottomTrailingRadius: .greatestFiniteMagnitude,
+                        topTrailingRadius: .greatestFiniteMagnitude
+                    )
+                    .fill(.themedAccent)
+                    .frame(width: proxy.size.width * CGFloat(choice.voteCount ?? 0) / CGFloat(poll.totalVotes))
```
EricBAndrews: *nit*

Logically speaking we should never see 0 votes when we display this, but maybe worth forcing the denominator here to be at least 1 to avoid div by 0 behavior?


```suggestion
                    .frame(width: proxy.size.width * CGFloat(choice.voteCount ?? 0) / CGFloat(max(1, poll.totalVotes)))
```

EricBAndrews (approved):
LGTM! 🔥

Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/PostPollView.swift
Sjmarf: Good spot, definitely worth having this. Added 👍

Issue #1787, opened 2025-02-21
Notification count broken in Lemmy 1.0


Sjmarf:
In 1.0, the schema of `ApiGetUnreadCountResponse` has changed:

```swift
public struct ApiGetUnreadCountResponse: Codable, Hashable, Sendable {
    /// Lemmy availability: unavailable after 0.19.11
    public let replies: Int?
    /// Lemmy availability: unavailable after 0.19.11
    public let mentions: Int?
    /// Lemmy availability: unavailable after 0.19.11
    public let privateMessages: Int?
    /// Lemmy availability: available from 1.0.0-alpha onwards
    public let count: Int?
}
```

Following these changes, we will either need to:
- Remove the setting that lets the user choose which types of notifications to include in the unread count.
- Or, determine the unread count by fetching the first page of inbox results (this might be more complexity than it's worth).

Worth considering is that Piefed *does* give us the separate notification types, and I don't think it's likely they'll take that away as Lemmy did.

Sjmarf:
https://github.com/LemmyNet/lemmy/issues/6329

PR #2550, opened 2026-01-30
Implement inbox on Lemmy 1.0

### LINKED ISSUE #2157: Rethink MlemMiddleware notification types

### LINKED ISSUE #2340: Create Lemmy 1.0 mappings for `InboxNotification`
At the time of writing only Lemmy < 1.0 has API Type -> `InboxNotificationSnapshot` mappings in place

Sjmarf:
On Lemmy 1.0, you can now load the inbox and mark items as read.

Closes #2340, closes #2157

Known issues:
- When you open the inbox, you get a 404 error. This is caused by fetching the mod mail, which we haven't implemented yet.
- The inbox tab count may not stay in sync. This is a separate issue (#1787)

### DIFF
--- a/Mlem/App/Views/Shared/TestInboxView.swift
+++ b/Mlem/App/Views/Shared/TestInboxView.swift
@@ -86,11 +86,26 @@ private struct TestInboxSectionView: View {
         do {
             switch type {
             case .reply:
-                notifications = try await appState.firstApi.getReplyNotifications(page: 1, limit: 5, unreadOnly: false)
+                notifications = try await appState.firstApi.getReplyNotifications(
+                    page: 1,
+                    cursor: nil,
+                    limit: 5,
+                    unreadOnly: false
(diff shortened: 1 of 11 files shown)


EricBAndrews (approved):
LGTM! 🔥

PR #2552, opened 2026-02-02
Vote in polls

### LINKED ISSUE #2543: Voting in polls
Note that there are two types of polls:
- Single-option
- Multi-option

We should support both

Sjmarf:
There are two types of poll: single-choice and multi-choice. Once you've voted, you can't retract your vote.

For convenience, I've made a few poll posts in the mlemtesting@piefed.social community, and there's some in polls@piefed.social too

Closes #2543

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/LargePostBodyView.swift
+++ b/Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/LargePostBodyView.swift
@@ -30,7 +30,7 @@ struct LargePostBodyView: View {
 
             switch post.type {
             case let .poll(poll):
-                PostPollView(poll: poll)
+                PostPollView(post: post, poll: poll)
                 if post.content != nil {
                     Divider().padding(.horizontal, -Constants.main.standardSpacing)
                 }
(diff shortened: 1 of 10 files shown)


EricBAndrews (approved):
LGTM! 💯 

Do we need the "submitted" info at the bottom? It feels a bit unnecessary when revisiting the poll--maybe we can turn the check gray to indicate interaction is no longer allowed, and just pop a toast for the confirm?

Sjmarf:
> Do we need the "submitted" info at the bottom? It feels a bit unnecessary when revisiting the poll--maybe we can turn the check gray to indicate interaction is no longer allowed, and just pop a toast for the confirm?

There are two problems that the current design solves:
- You need to be able to select multiple actions and _then_ confirm your choice. For this reason, we can't use a confirmation alert.
- When you submit, what happens to the "show results" button? If we just hide it, the whole post changes size which I found to be jarring.

We _could_ hide the "submitted" button if you visit a poll for a second time 🤔 I'm not sure whether it's better to hide or keep it in that scenario

EricBAndrews:
Hmm, that's fair. I'm happy to merge as-is and we can refine the design later if inspiration strikes

PR #2564, opened 2026-02-08
New action system for communities

Sjmarf:
Functional changes:

The "Unsubscribe" action is no longer red. I think this is slightly better; it's not really _that_ destructive. If you disagree, I don't mind keeping it red.

The "subscribe" swipe action now uses the `plus.circle` symbol rather than `person.crop.circle.badge.plus`. This is a technical limitation with the new action system. I may fix this in future, but for now I think this icon is fine.

Increased the border radius of `WarningView` to match the increased border radius on iOS 26. This fixes a rendering bug in the "purge" sheet.


EricBAndrews (approved):
LGTM! 🔥

PR #2565, opened 2026-02-08
Unified Person

### LINKED ISSUE #2049: 1.0: Move `banned` property from `Person1` -> `Person2`
It's no longer possible to get this value from a `Person1`. At the moment, all users are assumed to be unbanned when using 1.0.

### LINKED ISSUE #2529: Unified Models: Person

### LINKED ISSUE #2554: ExpandedPostView no longer auto-resolves when switching account using "keep place" mode
I think this was broken as a result of the post model being replaced

EricBAndrews:
## Issues

- closes #2529 
- closes #2049 
- closes #2554 

## Description

Unified Person model

## Implementation Notes

`updateIfChanged`: I realized there is a technically a possibility for higher tier fields to be unset if a model is updated with properties from a lower tier snapshot. This shouldn't happen as long as things go through the update queue, since properties handle this logic in their merge, but in case we ever do something weird I've set higher tier properties on the model to only be updated if the incoming value is non-nil.

`User`: currently what was previously `Person4` is incorporated into `Person`. This causes a lot of unwrapping annoyance; it may be worth separating that model out into a `User` that is guaranteed to have the user-specific fields and contains a `Person`. OOS here, as this PR is just supposed to plug in the new models with no major refactors.

`CodedData`: now only encodes as `Person1`. This forces a fetch for user stats in the visit history, but that's a good thing IMO since it ensures up-to-date data.

Removed function to get tier 3 person by URL: this logic is now implicit in the upgrade system


Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Person/Person.swift
Sjmarf, line 1: Moderator actions do not appear on posts in a community I moderate

Sjmarf, line 1: On PieFed, if I edit the user note on myself, my avatar disappears.

Sjmarf: I've also had my own user note disappear by just navigating around. Maybe a lower-tier model is getting loaded and setting it to `nil`?

Mlem/App/Views/Pages/Person/PersonView.swift
Sjmarf, line 1: #2554

This issue also applies to `PersonView` in this PR

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Person/Person.swift
EricBAndrews: Should be fixed 👍

EricBAndrews: Nice catch! The avatar disappearance was caused by profile updates not going through the queue, which in turn caused the queue to have a stale set of properties; `editNote` is a property-modifying operation, and so when it triggered the older properties were assigned over the updated ones.

EricBAndrews: The disappearing note is reproducible on dev, I've opened #2569 for it but it's out of scope here

Mlem/App/Views/Pages/Person/PersonView.swift
EricBAndrews: Nice catch, I've implemented a fix. There is now a `reloadOnAccountSwitch` modifier that handles this logic; for now, the relevant protocol requirements for resolving the new model are on `UnifiedModelProviding` to avoid conflicts with the existing `Community` `Resolvable` conformance, but once `Community` is migrated to the new system I will move those requirements into `Resolvable`.

Sjmarf: Nice 👍 

It works for the post itself, but the comment chain also needs to be resolved on the new API.

Sjmarf: Nice, works 👍

Sjmarf, line 1: My banner doesn't show in `PersonView`. It shows correctly in the profile editor

EricBAndrews: Nice catch, I forgot to shim the `Profile1Providing`

Sjmarf (approved):
LGTM 🚀 🚀

PR #2570, opened 2026-02-10
Fix piefed.zip

Sjmarf:
Piefed.zip is running a new version of PieFed that breaks Mlem completely (and other clients, apparently). I expect that either they'll roll back to an earlier version or PieFed will fix the bug.

This fixes the issue on our side - if it's not fixed on their side by the time you see this, we could put out a TestFlight?


Sjmarf:
piefed.zip fixed it

PR #2585, opened 2026-02-16
Sort subscription list by instance secondarily

Sjmarf:
The subscription list is now sorted by name and then by instance if two communities have the same name.


EricBAndrews (approved):
LGTM! 👍

PR #2590, opened 2026-02-17
Don't play haptics when tapping ended poll, show toast instead

### LINKED ISSUE #2589: Don't play haptics when tapping poll answer if poll has ended
Tapping a poll row plays a haptic. At the moment, this haptic still plays even if the poll has ended. We should change it so that no haptic plays in this case.

un1970ix:
## Issues

- closes #2589

## Description

Prevents haptic feedback when tapping a poll answer after the poll has ended, and shows an informative toast instead.

## Implementation Notes

- Added `poll.hasEnded` guard at the top of the tap gesture handler in `PostPollView.choiceView`, before the haptic and selection logic
- Shows a toast ("Poll has ended") matching the existing `hasVoted` pattern for consistency


Sjmarf (approved):
Nice 👌

PR #2591, opened 2026-02-17
Remove unused Face ID usage description

### LINKED ISSUE #2344: Remove FaceID intent
We used this for the "lock with FaceID" feature, which has since been removed

un1970ix:
## Issues

- closes #2344

## Description

Removes the `NSFaceIDUsageDescription` plist key.

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -3309,7 +3309,6 @@
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = Mlem/Info.plist;
 				INFOPLIST_KEY_CFBundleDisplayName = Mlem;
-				INFOPLIST_KEY_ITSAppUsesNonExemptEncryption = NO;
 				INFOPLIST_KEY_LSApplicationCategoryType = "public.app-category.entertainment";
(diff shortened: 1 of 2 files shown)


Mlem.xcodeproj/project.pbxproj
```diff
@@ -3357,8 +3356,8 @@
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = Mlem/Info.plist;
 				INFOPLIST_KEY_CFBundleDisplayName = Mlem;
+				INFOPLIST_KEY_ITSAppUsesNonExemptEncryption = NO;
```
Sjmarf: Presumably Xcode added this automatically? Not sure whether we need to change this. 

Apparently this is used to avoid having to re-submit some security questions every time you push a build to TestFlight. So it may be worth adding this flag, but either way we should leave it out of this PR.

@EricBAndrews

https://developer.apple.com/documentation/security/complying-with-encryption-export-regulations

Sjmarf (changes requested):

Mlem.xcodeproj/project.pbxproj
un1970ix: Yes, it was added automatically. Now deleted, thanks for the heads-up.

Sjmarf (approved):
LGTM 👍

PR #2593, opened 2026-02-18
Fix search filter persistence, unify filtersActive handling, and bypass hide-read in search

### LINKED ISSUE #2381: Post search refreshes on navigate back

un1970ix:
## Issues

- closes #2381

## Description

Fixes search filters resetting and refresh when navigating back from a result, and makes the filter toggle behave consistently across all search tabs. Also prevents the global "hide read" setting from hiding posts in search results.

## Implementation Notes

- `setupFilters()` now guards with `communityFilters == nil` to avoid recreating filter objects (and resetting their values) when `.task(id:)` re-fires on navigate-back
- `refreshPosts` and `refreshComments` now check `filtersActive` before applying sort, creator, and location filters, matching the existing pattern in `refreshCommunities`/`refreshPeople`/instances
- Added `.onChange(of: appState.firstApi)` to nil out filter objects so they get recreated with correct defaults on account switch
- Passed `alwaysShowRead: true` to `PostGridView` in search results since hiding read posts in search is not useful

### DIFF
--- a/Mlem/App/Views/Shared/Search/SearchView+Logic.swift
+++ b/Mlem/App/Views/Shared/Search/SearchView+Logic.swift
@@ -49,6 +49,7 @@ extension SearchView {
     }
     
     func setupFilters() async {
+        guard communityFilters == nil else { return }
         do {
             let software = try await appState.firstApi.software
             communityFilters = .init(software: software)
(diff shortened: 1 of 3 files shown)


Sjmarf (approved):
Nice, works perfectly 💯

PR #2594, opened 2026-02-18
Show toast when hide read is toggled

### LINKED ISSUE #2308: Show toast when "hide read" toggled

un1970ix:
## Issues

- closes #2308

## Description

Shows a toast when the user taps "Hide Read" or "Show Read" in the feed or inbox.

## Implementation Notes

- Added toast in `FeedToolbarOptions` and `InboxView+Views`
- Uses `.success()` toast to match existing patterns
- Uses explicit `LocalizedStringResource` type so Xcode picks up the strings for localization

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Inbox/InboxView+Views.swift
+++ b/Mlem/App/Views/Root/Tabs/Inbox/InboxView+Views.swift
@@ -141,7 +141,7 @@ extension InboxView {
         Button {
             showRead.toggle()
             let message: LocalizedStringResource = showRead ? "Showing Read" : "Hiding Read"
-            ToastModel.main.add(.success(message))
+            toastModel.add(.success(message))
         } label: {
             Label("Hide Read", icon: .general.filterMenu)
                 .symbolVariant(showRead ? .none : .fill)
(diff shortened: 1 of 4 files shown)


Mlem/App/Views/Root/Tabs/Inbox/InboxView+Views.swift
```diff
@@ -140,6 +140,8 @@ extension InboxView {
     var hideReadButton: some View {
         Button {
             showRead.toggle()
+            let message: LocalizedStringResource = showRead ? "Showing Read" : "Hiding Read"
+            ToastModel.main.add(.success(message))
```
Sjmarf: _nit_

It's more idiomatic to grab this from the environment, rather than using the static property.

```
@Environment(ToastModel.self) var toastModel
```

In future I want to mark the singleton as deprecated to make this more clear (which we've done for `HapticManager` already). Don't do this rn, though - first we need to check that it isn't used anywhere where we can't access the environment

Feel free to merge once resolved :)

Sjmarf (approved):

Mlem/App/Views/Root/Tabs/Inbox/InboxView+Views.swift
un1970ix: Thanks for the tip, updated!

PR #2595, opened 2026-02-18
Fix username shown twice for posts in mod mail

### LINKED ISSUE #2507: Username is shown twice for posts in mod mail

un1970ix:
## Issues

- closes #2507

## Description

Fixes the post creator name showing up twice on reported posts in mod mail.

## Implementation Notes

- In reports, `favoredLink: .creator` is passed, which already shows the creator at the top
- The `alwaysShowCreator` setting then shows it again below
- Added a check to skip the second one if the creator is already shown at the top
- Applied to both `HeadlinePostView` and `LargePostView`
- Normal feed views are not affected


Sjmarf (approved):
LGTM 👍

PR #2596, opened 2026-02-18
Auto-dismiss TestFlight banner when navigating to update post from feed

### LINKED ISSUE #2365: TestFlight banner should disappear if you navigate to the update post from the feed

un1970ix:
## Issues

- closes #2365

## Description

Hides the TestFlight update banner when the user opens the update post from their feed.

## Implementation Notes

- Added `.onChange(of: navigation.path)` to `UpdateBannerView`
- Checks if the user navigated to a post matching the banner URL
- Uses `allResolvableUrls` to match the URL regardless of which instance the user is on
- Calls the existing `dismiss()` function


Sjmarf (approved):
Nice :100:

PR #2597, opened 2026-02-18
Preserve removed comment content from being overwritten (Modlog)

### LINKED ISSUE #2548: `Comment.content` field should NOT be updated with incoming changes if `Comment.removed` is `true`
Repro:
- Open the modlog
- Go to a removed comment
- Comment appears empty

Introduced in #2527

un1970ix:
## Issues

- closes #2548

## Description

Fixes removed comments showing up empty in the modlog. When navigating to a removed comment, the API returns empty content, which was overwriting the locally stored content.

## Implementation Notes

- Added a guard in `Comment.update(with:)` to skip updating `content` when the incoming data has `removed` set to `true`


Sjmarf (approved):
LGTM 👍

PR #2604, opened 2026-02-21
Add "Open in Mlem" action extension

### LINKED ISSUE #128: Add "Share" feature
## Introduction
I am testing MLem.
Congratulations on the project.

## Requirements
I suggest improving the app by adding the share feature, so it will be possible - for example - to share content from Safari directly to MLem.

un1970ix:
## Issues

- closes #128

## Description

Adds an "Open in Mlem" action extension to the iOS share sheet. When viewing a Lemmy post or community in Safari, users can tap Share > Open in Mlem to open the link directly in the app.

## Implementation Notes

- The action extension (`OpenInMlem`) uses JavaScript to get the page URL, swaps the scheme to `mlem://`, and redirects the browser to open the app
- `MlemApp.onOpenURL` catches the `mlem://` scheme, converts it back to `https://`, and sets `NavigationModel.main.pendingOpenURL`
- `HandleThreadiverseLinksModifier` observes `pendingOpenURL` via `.onChange` and routes it through the existing `didReceiveURL` pipeline, same path as the paste button in search
- French localization included for the extension display name


Mlem/App/Views/Shared/HandleThreadiverseLinksModifier.swift
```diff
@@ -43,6 +43,12 @@ struct HandleThreadiverseLinksModifier: ViewModifier {
     func body(content: Content) -> some View {
         content
             .environment(\.openURL, OpenURLAction(handler: didReceiveURL))
+            .onChange(of: NavigationModel.main.pendingOpenURL) { _, url in
+                if let url {
+                    NavigationModel.main.pendingOpenURL = nil
+                    _ = didReceiveURL(url)
+                }
+            }
```
Sjmarf: In general I prefer to avoid using `.main` on any of the singleton models where possible. We used to use them everywhere, but I now think that was a mistake - it caused a lot of headaches in our SwiftUI `#Preview` code, and I imagine could cause problems in tests too if we ever add any.

```suggestion
            .onChange(of: navigation.model?.pendingOpenURL) { _, url in
                if let url {
                    navigation.model?.pendingOpenURL = nil
                    _ = didReceiveURL(url)
                }
            }
```

Mlem/App/Views/Root/MlemApp.swift
```diff
@@ -44,6 +44,13 @@ struct MlemApp: App {
     var body: some Scene {
         WindowGroup {
             ContentView()
+                .onOpenURL { url in
+                    guard url.scheme == "mlem" else { return }
+                    var components = URLComponents(url: url, resolvingAgainstBaseURL: false)
+                    components?.scheme = "https"
+                    guard let targetURL = components?.url else { return }
+                    NavigationModel.main.pendingOpenURL = targetURL
+                }
```
Sjmarf: Ditto with using the `.main` property. 

If we put this code inside `ContentView`, could we use `self.navigationModel.pendingOpenURL = targetURL` instead?

`ContentView` has this code currently:

```swift
    var navigationModel: NavigationModel { .main }
```

Eventually I want to change this to:

```swift
    var navigationModel: NavigationModel = .init()
```

...and remove `.main` entirely.

Sjmarf (changes requested):
Nice, great job on this 🔥 A couple minor nits

Sjmarf (approved):
Nice 🚀

PR #2610, opened 2026-02-22
Customizable action sheet

Sjmarf:
Added customizability to the action sheet used in the test inbox. At the bottom of the action sheet, there's a "customize" button that opens an editor. Changes are persisted to the disk.


EricBAndrews (approved):
LGTM! 🚀

PR #2623, opened 2026-02-28
Add `contextMenu` property to all `-BarConfiguration` classes

Sjmarf:
In #2610, I made the test inbox action sheet customisable. The configuration is tracked in `ReplyBarConfiguration.contextMenu`. 

In this PR, I added `contextMenu` to the other `-BarConfiguration` classes, and made it a requirement of the protocol. No functional changes. There is NO customisable action sheet for these content types yet.


Mlem/App/Enums/Interaction/PostBarConfiguration.swift
```diff
@@ -194,7 +205,8 @@ struct PostBarConfiguration: InteractionBarConfiguration {
             leadingSwipes: [.upvote, .downvote],
             trailingSwipes: [.save, .reply],
             readouts: [.created, .comment],
-            availableWidgets: .init(CounterType.defaultWidgets.map { .counter($0) } + ActionType.defaultWidgets.map { .action($0) })
+            availableWidgets: .init(CounterType.defaultWidgets.map { .counter($0) } + ActionType.defaultWidgets.map { .action($0) }),
+            contextMenu: [.selectText, .share, .blockCreator, .report, .edit, .delete, .remove, .banCreator]
```
Sjmarf: I think this is a good default configuration. Compared to the current iteration, this does not include `.upvote`, `.downvote`, `.reply`, or `.save`. I think these would be redundant because they are available in the interaction bar by default. `.hide` is also hidden by default. What do you think?

Comments have the same default configuration - notably missing `.createImage`. 

Changing this will be tricky after this PR is merged, so worth discussing what we want the default to be now.

Sjmarf:
Actually I think it's better to use `nil` to represent the default state, which makes it much easier to change the default later

EricBAndrews (approved):
LGTM! 💯

PR #2626, opened 2026-03-01
Unified Community

### LINKED ISSUE #2528: Unified Models: Community

EricBAndrews:
## Issues

- closes #2528 

## Description

Implements the unified community model

## Implementation Notes

Most of these changes are the usual fare for unified model updates, with the notable exception of the `ValueProviding` system. This has been significantly extended:
- `RealizedValueProviding` is now its own protocol offering `.realizedValue` to directly access the non-optional value
- Added two new types of `ValueProviding`: `SyntheticExpectedValue` and `SyntheticRealizedValue`. Both are `ValueSynthesizer`s, which replaces the old `SyntheticStateManager`
- All instances of `SyntheticStateManager`now use `SyntheticExpectedValue` or `SyntheticRealizedValue`, as appropriate

`BubblePickerView` no longer throws an assertion failure if instantiated with an invalid tab; instead, the selected tab is gracefully defaulted to 0. This behavior is expected if you have a person's moderated communities open and you remove them from their only moderated community in a different tab.


Mlem/App/Views/Pages/Instance/InstanceCommunityListView.swift
```diff
@@ -38,7 +41,7 @@ struct InstanceCommunityListView: View {
         }
         .animation(.easeOut(duration: 0.1), value: communityLoader.items.isEmpty)
         .task {
-            print(errorDetails)
+            logger.error("\(String(describing: errorDetails))")
```
Sjmarf: We should use `errorDetails.errorText()` so that tokens get properly redacted

I wonder whether we could be smarter about our redaction system, to avoid leaks accidentally slipping in (OOS) 🤔

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Shared/ValueProviding/ValueSynthesizer.swift
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Allows optionals to be used as MergeableValue
+extension Optional: MergeableValue where Wrapped: MergeableValue & Equatable {
+    public func merge(with other: Optional<Wrapped>, using mergeType: ValueMergeType) -> Optional<Wrapped> {
+        return self.map { value in
+            return other.map { otherValue in
+                return value.merge(with: otherValue, using: mergeType)
+            } ?? value
+        } ?? other
```
Sjmarf: It might be clearer to use a `switch` statement here

```suggestion
        switch self, other {
            case let (a, b):
                a.merge(with: b, using: mergeType)
            case (let a, nil):
                a
            case (nil, let b):
                b
            case (nil, nil):
                nil
        }
```

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/InstanceConnection/LemmyConnection/LemmyConnection+Community.swift
```diff
@@ -168,3 +163,5 @@ public extension LemmyConnection {
         )
     }
 }
+
+import os
```
Sjmarf: Dead code?


```suggestion
```

Sjmarf (changes requested):
Nice. A few minor code comments. Will do some proper testing later

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Shared/ValueProviding/ValueSynthesizer.swift
EricBAndrews: I've implemented a slightly different approach, which is more readable than the `map` method but more concise than the switch:

```
        if let other {
            return self?.merge(with: other, using: mergeType) ?? other
        }
        return self
```

The switch is still a little more explicit, but I don't expect this logic to change so I think this way is sufficiently comprehensible.

Interestingly, the map approach was the only one that would compile when I was originally doing development, but the more "normal" ways all work fine now. Just xcode being xcode, I suppose.

Sjmarf: Nice, that works 👍

Sjmarf (changes requested):
If you switch account in "keep place" mode, it scrolls to the top of the `CommunityView`. It didn't do this previously

EricBAndrews:
Nice catch! We had a `.id(entity.uid)` on `reloadOnAccountSwitch` that was causing this. I've removed it, and manually tagged `PersonView` with `.id(person.uid)` to preserve the existing behavior. `ExpandedPostview` is the only other place using `reloadOnAccountSwitch`, and it behaves normally after removing that `id`.

Sjmarf (approved):
Nice 🚀🚀

Issue #2617, opened 2026-02-25
Community, User, and Instance list gesture customization


Coolie2:
### Requirements

- [x] There are no existing issues for this feature.
- [x] This is a request for a **single** feature (create multiple issues for multiple feature requests).

### Description

A feature I had requested for Arctic that was added to its TestFlight build was the ability to swipe to block communities on the community list. Perhaps such functionality could be added to Mlem as well, perhaps in the same style as Swipe Actions for posts and comments are currently handled. Other optional functions that could be swiped include Block for all three lists, Send Message and Modlog for users, Favorite/Unfavorite, Subscribe/Unsubscribe, and New Post for communities, along with possibly other options from their popup menus.

While the functionality itself can be used via the pop-up menu, swipes would make conducting some tasks in bulk, such as blocking, faster.

PR #2630, opened 2026-03-04
Customisable community swipe actions

### LINKED ISSUE #2621: Settings header descriptions are no longer localized

Sjmarf:
Added a setting for customising community swipe actions. There's a new section called "communities" at the top level of settings that opens the editor directly. Not sure if this is the best place to put the setting?

Partially completes #2617

### DIFF
--- a/Mlem/App/Configuration/User Settings/SettingsValues.swift
+++ b/Mlem/App/Configuration/User Settings/SettingsValues.swift
@@ -115,6 +115,7 @@ class SettingsValues: Codable { // swiftlint:disable:this type_body_length
     var interactionBar_post: PostBarConfiguration
     var interactionBar_comment: CommentBarConfiguration
     var interactionBar_reply: ReplyBarConfiguration
+    var interactionBar_community: CommunityActionConfiguration
     var interactionBar_postReport: PostBarConfiguration
     var interactionBar_commentReport: CommentBarConfiguration
     var interactionBar_alternateReportLayout: Bool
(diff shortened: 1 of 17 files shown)


Mlem/App/Configuration/User Settings/SettingsValues.swift
```diff
@@ -260,6 +261,7 @@ class SettingsValues: Codable { // swiftlint:disable:this type_body_length
         self.interactionBar_post = try container.decodeIfPresent(PostBarConfiguration.self, forKey: ._interactionBar_post) ?? .default
         self.interactionBar_comment = try container.decodeIfPresent(CommentBarConfiguration.self, forKey: ._interactionBar_comment) ?? .default
         self.interactionBar_reply = try container.decodeIfPresent(ReplyBarConfiguration.self, forKey: ._interactionBar_reply) ?? .default
+        self.interactionBar_community = try container.decodeIfPresent(CommunityActionConfiguration.self, forKey: ._interactionBar_community) ?? .init()
         self.interactionBar_postReport = try container.decodeIfPresent(PostBarConfiguration.self, forKey: ._interactionBar_postReport) ?? .reportDefault_
         self.interactionBar_commentReport = try container.decodeIfPresent(CommentBarConfiguration.self, forKey: ._interactionBar_commentReport) ?? .reportDefault_
         self.interactionBar_alternateReportLayout = try container.decodeIfPresent(Bool.self, forKey: ._interactionBar_alternateReportLayout) ?? false
```
Sjmarf, line 267: `interactionBar` is an outdated name, because these structs store more than just the interaction bar now. Do we want to stick with the naming scheme for the community action configuration, or call it something like `actions_community`?

`CommunityActionConfiguration` is very different from the `-BarConfiguration` structs at the moment, but eventually they'll all be similar and conform to the same protocols.

Mlem/App/Views/Root/Tabs/Settings/NewSwipeActionEditorView.swift
```diff
@@ -0,0 +1,73 @@
...
+//  NewSwipeActionEditorView.swift
+//  Mlem
+//
+//  Created by Sjmarf on 2026-03-04.
+//
+
+import Actions
+import SwiftUI
+
+struct NewSwipeActionEditorView: View {
```
Sjmarf, line 11: This is copy-pasted from `SwipeActionEditorView` and adapted for the new system. Eventually `SwipeActionEditorView` will be deleted, and this will replace it.

Mlem/App/Configuration/User Settings/SettingsValues.swift
EricBAndrews: Let's keep it as-is--IMO one stale naming schema is easier to read and maintain than two parallel ones. We should update this naming eventually, but that should be done all at once.

EricBAndrews (changes requested):
Super clean, works well!

The setting location feels a bit jarring, which I think would be solved by pushing it one layer deeper to match Posts and Comments. One fix would be to put the subscription list in Communities instead of the top layer, and maybe pad out the view with a toggle for hide/show community avatar--that's three different settings, which should make the page feel normal.

Sjmarf:
Yeah that's nicer, thanks. Thoughts?

Screenshot 2026-03-05 at 13 57 06

EricBAndrews (approved):
LGTM! 🔥

PR #2633, opened 2026-03-07
Show community name in post navigation bar

### LINKED ISSUE #2582: Show community name at top of open post view
Often when reading long comment threads I realise I didn't actually see which community the post was in, and this affects the context of comments. I think the only way to find it is scroll all the way to the top again?

It would be great if there was an option to show the community in the header bar once scrolling down into the comments.

EricBAndrews:
## Issues

- closes #2582 

Note that longer community names get left-aligned; this is Apple system behavior. Keeping it like that seemed neater than truncating or trying to hack font size down.


Sjmarf (approved):
Nice 💯

PR #2650, opened 2026-03-16
New swipe actions system

Sjmarf:
Swipe actions for posts, comments and the inbox now use the new action system.

# Functional changes

- There are now more actions available to be set as swipe actions.
- The "collapse comment" swipe action now shows a plus symbol rather than a minus when the comment is already collapsed.
- The three "collapse" actions are now colored orange rather than purple.
- Tapping "apply to all" in the editor for the alternate mod mail swipe actions now applies the configuration _only_ to mod mail swipe actions.

# Technical details

In #2630, I introduced `CommunityActionConfiguration`, which uses the new codable action format. In this PR, I extracted that interface out to a protocol `SwipeActionConfiguration`, which `CommunityActionConfiguration` conforms to:

```swift
protocol SwipeActionConfiguration {
    var savedSwipes: ActionSeedSwipeConfiguration? { get set }

    static var availableActions: ActionSeedSections { get }
    static var defaultSwipes: ActionSeedSwipeConfiguration { get }
}
```

Then, I made the `InteractionBarConfiguration` protocol inherit from `SwipeActionConfiguration`. This means that all three of the `-BarConfiguration` classes now use that new format on the disk.

Eventually `InteractionBarConfiguration` will be removed, and we'll have individual `SwipeActionConfiguration`, `BarActionConfiguration` and `ContextActionConfiguration` protocols.

# Testing

I have tested:
- All of the non-moderator and moderator actions. I haven't tested the admin-only ones
- Upgrading from the previous version converts the format correctly
- Alternate mod mail layouts still behave as expected


EricBAndrews (approved):
LGTM! 🔥 

Nice to see everything aligning under the new standards 💯

PR #2653, opened 2026-03-17
Fix link resolution sometimes not working

Sjmarf:
Links to posts or comments could fail to resolve in cases where the linked instance isn't in our instance list.


EricBAndrews (approved):
LGTM! 💯

PR #2655, opened 2026-03-18
Fix swipe freezing when gesture is cancelled

### LINKED ISSUE #2644: Swipe action animation freeze glitch
### Requirements

### Description

I've noticed the swipe actions getting frozen in place before, but just found a way to make it happen consistently. If you hold a swipe action in place while turning off the device with the power button, the swipe action will be frozen in place when returning to the app:

Image

### Mlem Version

Mlem 2.4.1 (730)

EricBAndrews:
## Issues

- closes #2644


Sjmarf (approved):
LGTM 💯

PR #2656, opened 2026-03-19
Use `RestClient` for `BackendClient`

Sjmarf:
- Moved `BackendClient` to a new `MlemBackend` package
- `BackendClient` now uses `RestClient`

`RestClient` includes the URL response when it throws a decoding error. This change will make it easier to debug some issues people were reporting in the Matrix room.


EricBAndrews (approved):
LGTM! 🔥

Issue #2658, opened 2026-03-19
`BubblePicker` tab misalignment glitch


Sjmarf:
Seems to happen randomly

<img width=50% alt="Image" src="https://github.com/user-attachments/assets/c7cb32c8-bd15-4768-8827-337f32c4f9f0" />

PR #2661, opened 2026-03-21
Possible bubble picker fix

Sjmarf:
Apparently this might be partially to do with the value of a tab changing? I've added an `.id` to recreate the `ChildSizeReader` if the value changes.

Progress towards #2658


EricBAndrews (approved):
LGTM 👍

PR #2662, opened 2026-03-21
Generic context menu configuration

Sjmarf:
No functional changes.

---

In #2650 I created the `SwipeActionConfiguration` protocol, which `InteractionBarProtocol` inherits from. In this PR, I've created a similar `ContextMenuConfiguration` protocol, which `InteractionBarConfiguration` also inherits from, and moved the relevant parts of `InteractionBarConfiguration` to it.

In #2610, I made the action sheet (which is available in the test inbox) customizable:

Before this PR, the action sheet customization system was tightly coupled to `ReplyBarConfiguration` specifically. Now, it's generalised to work with any `ContextMenuConfiguration`.


EricBAndrews (approved):
LGTM! 🔥

PR #2665, opened 2026-03-23
Improve quick swipes/interactive navigation gesture conflict

EricBAndrews:
Addresses a longstanding complaint that the swipe gestures make interactively navigating back awkward. Quick swipe gestures now do not trigger on a small leading buffer zone, allowing the interactive dismissal gesture to be triggered more easily. Gestures on views that do not abut the edge (e.g., nested comments) should be unaffected.


Sjmarf:
LGTM 💯 It's good to address some of these long-standing issues :)

Sjmarf (approved):

PR #2679, opened 2026-04-03
Customisable post/comment context menus

Sjmarf:
- Post/comment context menus are now customisable
- Removed "disclosure group" case for the "moderator action grouping" setting
- Messages in the inbox view no longer have a separate moderator action menu (replies didn't have this, so it was inconsistent)

I have tested that:
- Enabling the "Alternate mod mail configuration" setting produces different context menus for the mod mail
- "Separate menu" moderator action grouping still works as expected


EricBAndrews (approved):
LGTM! 🔥

PR #2682, opened 2026-04-06
Unified Instance

### LINKED ISSUE #2530: Unified Models: Instance

EricBAndrews:
## Issues

- closes #2530 

## Description

Implements a unified instance model, along with some additional tidying:
- `Profile1Providing` and `Profile2Providing` are now a single `ProfileProviding`. The only real consequence of this is that `Profile2Providing.created: Date` is now `ProfileProviding.profileCreated: Date?`, which is then implemented as either a passthrough to `.created` or computed to `nil`.
- `Instance` is `Blockable` (though blocking a non-local instance is still an edge case that is handled in `BlockAction`)
- `UnifiedUpdateQueue.refresh()` now behaves correctly with spinners
- The instance block action correctly says "unblock" if the instance is already blocked--this was reported in Matrix, can't find an issue if one was created

## Implementation Notes

- `InstanceStubResolutionPage` can have multiple targets, so takes in a closure to build the target navigation page from the fetched `Instance`


Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/ApiClient+Caches.swift
```diff
@@ -21,9 +21,7 @@ public protocol CacheIdentifiable {
 
 extension ApiClient {
     struct BaseCacheGroup {
-        var instance1: Instance1Cache = .init()
-        var instance2: Instance2Cache = .init()
-        var instance3: Instance3Cache = .init()
+        var instance: InstanceCache = .init()
```
Sjmarf, line 24: We should probably be cleaning this cache. Though, we weren't clearing the instance caches before this PR, either 🤔

Mlem/App/Actions/BlockAction.swift
```diff
@@ -16,40 +16,13 @@ struct BlockAction: Actions.Action {
...
-        var blockable: (any Blockable)? {
-            switch self {
-            case let .blockable(entity): entity
-            default: nil
-            }
-        }
-    }
-
-    let content: [Content]
+    let content: [any Blockable]
```
Sjmarf, line 20: Nice, this is much cleaner ✨

Mlem/App/Models/Account/Account.swift
```diff
@@ -42,9 +42,10 @@ enum AccountActivityState: Codable, Hashable {
     }
 }
 
-// Profile1Providing conformance
+// ProfileProviding conformance
 extension Account {
-    var blockedValue: Bool { false }
+    var blocked: any RealizedValueProviding<Bool> { RealizedValue(false) }
+    var displayName: String { name }
```
Sjmarf, line 48: Hmm, it's not ideal to be using `displayName` in this way. Maybe we should be storing the actual display name, and implement `var displayName` in `UserAccount` and `GuestAccount` rather than in this protocol. Happy to leave it as-is for now and make a separate issue though

Mlem/App/Models/Account/UserAccount.swift
```diff
@@ -120,14 +133,18 @@ class UserAccount: Account, CommunityOrPerson {
...
     
     var keychainId: String {
         getKeychainId(actorId: actorId)
     }
     
     @MainActor
-    func update(person: Person, instance: Instance3, software: SiteSoftware) {
+    func update(person: Person, software: SiteSoftware) {
         var shouldSave = false
         if avatar != person.avatar {
```
Sjmarf, line 149: We should replicate this logic for the other attributes we're now storing in `UserAccount` (description, banner, etc).  This ensures that the model gets updated correctly if the user edits their profile.

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Instance/Instance+Conformance.swift
```diff
@@ -0,0 +1,68 @@
...
+                }
+            }
+        }
+    }
+}
+
+// MARK: Sharable
+
+public extension Instance {
+    var allResolvableUrls: [URL] { [url()] }
```
Sjmarf: Hmm, making `Instance` conform to `Sharable` means that it also must conform to `Resolvable`, which is a requirement of `Sharable`. However, instances are _not_ resolvable using the API. Maybe we could remove `Resolvable` as a requirement of `Sharable`?

Mlem/App/Views/Shared/Search/Results/InstanceListRow.swift
```diff
@@ -64,7 +62,7 @@ struct InstanceListRow<Content2: View>: View {
         .padding(.vertical, 6)
         .background(.themedSecondaryGroupedBackground, in: .rect(cornerRadius: Constants.main.standardSpacing))
         .contentShape(.contextMenuPreview, .rect(cornerRadius: Constants.main.standardSpacing))
-        .contextMenu(instance: instanceStub)
+        .contextMenu(instance: instance)
```
Sjmarf, line 62: Instances in the instance list no longer have a context menu, for me.

Mlem/App/Views/Shared/Search/Results/InstanceListRow.swift
Sjmarf, line 1: The "recently searched" section in Search doesn't appear to be working for instances.

Sjmarf (commented):
Nice work! 😄 A few comments

Mlem/App/Views/Shared/Search/Results/InstanceListRow.swift
EricBAndrews: Nice catch! This required some moderate changes to fix:

- `InstanceSummary` and `Instance` now conform to `InstanceActionProviding`, which is now used in the relevant actions
- Added some conditional logic to check whether we have an `Instance` or an `InstanceSummary` in relevant cases
- `Sharable` is no longer `ContentIdentifiable` or `Resolvable`; it is simply `ActorIdentifiable` and `Hashable`. This allows `Instance` and `InstanceSummary` to conform to `Sharable` without requiring noop conformances, and doesn't actually affect how `Sharable` is used elsewhere

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Instance/Instance+Conformance.swift
EricBAndrews: See https://github.com/mlemgroup/mlem/pull/2682#discussion_r3105409330

Mlem/App/Models/Account/Account.swift
EricBAndrews: Opened https://github.com/mlemgroup/mlem/issues/2695

Mlem/App/Actions/BlockAction.swift
Sjmarf, line 1: Unblocking an instance from the blocklist seems to have no effect. When I reload the app it comes back

Mlem/App/Actions/ContextMenu+Instance.swift
```diff
@@ -34,11 +35,42 @@ extension View {
...
+
+public protocol InstanceActionProviding: Sharable, Blockable {
+    var instanceStub: InstanceStub { get }
+}
+
+extension Instance: InstanceActionProviding {
+    public var instanceStub: InstanceStub { .init(api: api, actorId: actorId) }
+}
+
+extension InstanceSummary: @retroactive Sharable {}
```
Sjmarf, line 68: I didn't know about `@retroactive` 😮 That's super useful

Mlem/App/Views/Shared/Search/Results/InstanceListRow.swift
Sjmarf: Nice 💯

Mlem/App/Actions/BlockAction.swift
EricBAndrews: Whoops, was passing the wrong id 🤦

Sjmarf (approved):
LGTM 🚀🚀

PR #2690, opened 2026-04-13
Fix "block" being shown for blocked communities rather than "unblock"

### LINKED ISSUE #2592: Make unblocking community more clear
If you already have a community blocked, clicking Block -> Community on a post unblocks it. However, it isn't very clear that this is what's happening:

https://github.com/user-attachments/assets/a1f3d051-2799-4f86-ac88-8e8031fda610

Sjmarf:
Previously, the popup would show "block community" and "block user" even when one was already blocked. Now, it shows "unblock community" / "unblock user" instead:

image

Closes #2592

### DIFF
--- a/Mlem/App/Actions/BlockAction.swift
+++ b/Mlem/App/Actions/BlockAction.swift
@@ -226,7 +226,7 @@ extension BlockAction {
                 callback: callback
             )
         }
-        environment.popupModel?.showPopup(message: "Community or User?", actions)
+        environment.popupModel?.showPopup(message: "Community or user?", actions)
     }
 
     @MainActor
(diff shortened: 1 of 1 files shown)


Mlem/App/Actions/BlockAction.swift
```diff
@@ -212,11 +212,21 @@ extension BlockAction {
...
+                contentType: item.blockable is Person ? .personOnly: .communityOnly
+            )
+            return .init(
+                title: label.title,
+                isDestructive: label.isDestructive,
+                callback: callback
+            )
         }
-        environment.popupModel?.showPopup(message: "Block...", actions)
+        environment.popupModel?.showPopup(message: "Community or user?", actions)
```
EricBAndrews: *nit*

This should match the order in which the options are presented:
```suggestion
        environment.popupModel?.showPopup(message: "User or community?", actions)
```

EricBAndrews (approved):
LGTM! 🔥

PR #2705, opened 2026-04-23
Improve Search Refresh Logic

### LINKED ISSUE #2686: Search term needs to be manually adjusted when switching search tabs
### Requirements

### Description

If you enter a search term on one search tab of Mlem, start the search in that category, and then switch to another category, the search term must be manually altered before it is applied to the new tab.

https://github.com/user-attachments/assets/0e630a83-8be7-408f-aa7e-7b072af6c25e

### Mlem Version

Mlem 2.4.1 (753)

EricBAndrews:
## Issues

- closes #2686 

## Description

Updates the search refresh logic. Search now tracks the last executed query per tab, and, upon switching tabs, refreshes if the current query does not match the last executed one for that tab.


Sjmarf:
Nice, this is much better UX.

I found a bug - repro:

- Restart app
- Search for something in the "communities" tab
- Switch to "posts" tab
- Posts load forever

Also, when the user switches tabs and it reloads, I think we should show a loading symbol while it's loading, rather than the old results.

Sjmarf (changes requested):
.

Mlem/App/Views/Shared/Search/SearchView+Views.swift
```diff
@@ -45,59 +45,79 @@ extension SearchView {
...
+                    }
                 }
-                EndOfFeedView(loadingState: .done, viewType: .hobbit)
             }
         case .posts:
             Group {
                 if postLoader.loadingState == .idle, postLoader.items.isEmpty {
                     searchPlaceholder
+                } else if query != lastExecutedQuery[.posts] {
+                    ProgressView()
```
Sjmarf: _nit_ IMO the progress views need more top padding

Sjmarf (approved):
LGTM 💯 One minor nit, optional

Issue #2251, opened 2025-09-04
Feature `canvas.fediverse.events` prominently at the top of the "Search" tab when the event is active


Sjmarf:
It's essentially r/place. Runs every July

https://canvas.fediverse.events

I imagine we could just show a fullscreen `WebView` with the website when the user taps on it. Maybe not an in-app safari view, to avoid all the browser UI.

grahhnt:
I introduced a fediverse.events api to query for upcoming events (with additional metadata) and automatic login (if the user's account is under control of the app)

* [fediverse.events-api docs (sc07.dev)](https://sc07.dev/fediverse.events/fediverse.events-api/-/wikis/home)
* [how auth works (sc07.dev)](https://sc07.dev/fediverse.events/fediverse.events-api/-/wikis/authentication)

Sjmarf:
Thanks :)

PR #2707, opened 2026-04-24
Refactor `SearchHomeView`

Sjmarf:
No functional changes

The main point of this was to create a reusable `SearchHomeListView`, which basically mimics a `Form` but is more flexible. I will be using this to add a section for events in #2251


EricBAndrews (approved):
Nice work! This is cleaner 😄

Issue #2710, opened 2026-04-26
Automatic authentication to canvas


Sjmarf:
See #2251

Issue #2709, opened 2026-04-26
Show countdown at top of community when event is active


Sjmarf:
See #2251

PR #2711, opened 2026-04-26
Initial events implementation

### LINKED ISSUE #2251: Feature `canvas.fediverse.events` prominently at the top of the "Search" tab when the event is active

Sjmarf:
Closes #2251

Events are now shown at the bottom of the "search" tab. You can switch to the QC backend under Settings -> Developer to test the feature.

When an event is tapped, it opens the related community, if one is available. If there is no community link associated with the event, it opens the event link directly. It is assumed that there will be a pinned post in the community explaining how to open the event, while it is active.

Tapping the ellipsis menu in the events section allows you to turn off the events feature. Once disabled, you can turn it on again in Settings -> General.

# Future plans

In future, I'd like to add some sort of countdown widget directly within the community view, to show when the event will start. There could be a big "open event" button in the community while it's ongoing. I've opened an issue for this: #2709

It's also possible for Mlem to handle authenticating to the event website with the user's Lemmy account, which I'll also implement soon: #2710


EricBAndrews (approved):
LGTM! 🔥

PR #2716, opened 2026-04-29
Compact comment readout improvements

### LINKED ISSUE #2683: Saved comments lack a bookmark icon
### Requirements

### Description

While saved posts have a green bookmark indicator, saved comments don't have that indicator.

https://github.com/user-attachments/assets/eee2cc61-bd27-4fb4-a7a8-c732ca24a185

### Mlem Version

Mlem 2.4.1 (753)

EricBAndrews:
## Issues

- closes #2683 

## Description

- Compact comments now use a static set of readouts rather than the configured interaction bar readouts
- Added a toggle to show comment downvotes separately (same behavior as posts)
- Collapsed comments now indicate upvote, downvote, and saved status. This follows the same behavior as the compact/tiled saved indicator, where a readout is only shown if the user has voted/saved.

### DIFF
--- a/Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/CompactPostView.swift
+++ b/Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/CompactPostView.swift
@@ -23,13 +23,11 @@ struct CompactPostView: View {
     let post: Post
     var requireConsistentHeight: Bool = false
     
-    var readouts: [PostBarConfiguration.ReadoutType?] {
-        let saved: PostBarConfiguration.ReadoutType? = post.saved.value ?? false ? .saved : nil
-        if showDownvotesCompact {
-            return [.created, .upvote, .downvote, .comment, saved]
-        } else {
-            return [.created, .score, .comment, saved]
-        }
+    var readouts: [PostBarConfiguration.ReadoutType] {
+        var readouts: [PostBarConfiguration.ReadoutType] = [.created]
+        readouts.append(contentsOf: showDownvotesCompact ? [.upvote, .downvote, .comment] : [.score, .comment])
+        readouts.appendIfPresent(post.saved.value ?? false ? .saved : nil)
+        return readouts
     }
     
     var blurred: Bool {
--- a/Mlem/App/Views/Root/Tabs/Settings/CommentSettingsView.swift
+++ b/Mlem/App/Views/Root/Tabs/Settings/CommentSettingsView.swift
@@ -12,6 +12,7 @@ struct CommentSettingsView: View {
     @Setting(\.comment_gestures_tapToCollapse) var tapCommentsToCollapse
     @Setting(\.comment_maxDepth) var maxCommentDepth
     @Setting(\.comment_jumpButton) var jumpButton
+    @Setting(\.comment_showDownvotesCompact) var showDownvotesCompact
     @Setting(\.interactionBar_comment) var commentInteractionBar
 
     var body: some View {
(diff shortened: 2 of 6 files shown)


Mlem/App/Views/Root/Tabs/Settings/CommentSettingsView.swift
```diff
@@ -43,6 +44,8 @@ struct CommentSettingsView: View {
                     icon: .settings.commentDepth,
                     destination: .settings(.commentMaximumDepth)
                 )
+                
+                Toggle("Show Downvotes Separately", icon: .lemmy.votes, isOn: $showDownvotesCompact)
```
Sjmarf, line 49: This setting should be shown conditionally, to match the behaviour in `PostSettingsView`


```suggestion
        if compactComments {
                Toggle("Show Downvotes Separately", icon: .lemmy.votes, isOn: $showDownvotesCompact)
        }
```

Mlem/App/Views/Shared/CommentView.swift
```diff
@@ -35,6 +36,15 @@ struct CommentView<EmbeddedContent: View>: View {
...
     let depthOffset: Int
     
+    var compactReadouts: [CommentBarConfiguration.ReadoutType?] {
+        let saved: CommentBarConfiguration.ReadoutType? = comment.saved.value ?? false ? .saved : nil
+        if showDownvotesCompact {
+            return [.created, .upvote, .downvote, .comment, saved]
+        } else {
+            return [.created, .score, .comment, saved]
+        }
+    }
```
Sjmarf: IMO the return type shouldn't be optional here. Having it non-optional avoids the added complexity in `InfoStackView.init`.

The following isn't particularly _elegant_, but it's readable:

```suggestion
    var compactReadouts: [CommentBarConfiguration.ReadoutType] {
        var result: [CommentBarConfiguration.ReadoutType] = [.created]
        if showDownvotesCompact {
            result += [.upvote, .downvote]
        } else {
            result += [.score]
        }
        result.append(.comment)
        if let saved = comment.saved.value, saved {
            result.append(.saved)
        }
        return result
    }
```

Sjmarf (changes requested):

Mlem/App/Views/Shared/CommentView.swift
EricBAndrews: Hmm, good point. I've extended `Array` to have `.appendIfPresent(Element?)` to make this logic a little prettier, and applied the same change to compact post and the post info stack init.

Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/CompactPostView.swift
```diff
@@ -23,13 +23,11 @@ struct CompactPostView: View {
...
-        let saved: PostBarConfiguration.ReadoutType? = post.saved.value ?? false ? .saved : nil
-        if showDownvotesCompact {
-            return [.created, .upvote, .downvote, .comment, saved]
-        } else {
-            return [.created, .score, .comment, saved]
-        }
+    var readouts: [PostBarConfiguration.ReadoutType] {
+        var readouts: [PostBarConfiguration.ReadoutType] = [.created]
+        readouts.append(contentsOf: showDownvotesCompact ? [.upvote, .downvote, .comment] : [.score, .comment])
+        readouts.appendIfPresent(post.saved.value ?? false ? .saved : nil)
```
Sjmarf, line 29: _nit_

Could maybe do something like `post.saved.value.map { .saved }`

Sjmarf (approved):
LGTM 👍

Mlem/App/Views/Root/Tabs/Feeds/Feed Posts/CompactPostView.swift
EricBAndrews: It still needs a ternary because the icon is only supposed to render when the comment is actually saved, so it would be the IMO equally unwieldy `post.saved.value.map { $0 ? .saved : nil }`

Issue #2728, opened 2026-05-09
Comment text is empty in the modlog on PieFed


Sjmarf:
Pending a fix on PieFed's side https://codeberg.org/rimu/pyfedi/issues/1795

PR #2729, opened 2026-05-09
PieFed modlog

### LINKED ISSUE #2622: PieFed modlog

Sjmarf:
Closes #2622

PieFed has slightly different behaviour regarding purged content. If a post is removed and then later purged, PieFed still returns an entry for the removed post. Lemmy doesn't do this.

There's some new UI to handle this case:

IMG_2453

Tapping the question mark opens this sheet:

IMG_2454

The same applies to comments, communities and users.

# Known issues

All comments in the modlog have an empty string for the body. The API is returning empty strings. I've asked the PieFed devs what's going on here, and have opened an issue to track this (#2728). If they confirm that this is intended, I'll add a better UI.


EricBAndrews (approved):
LGTM! 🔥

PR #2736, opened 2026-05-15
Always show controls on tap for videos

Sjmarf:
Previously, the video scrubbing bar would be inaccessible when the "show controls" setting is set to "never". Now, "never" behaves the same as "when I tap" for videos.


EricBAndrews (approved):
LGTM!

This might be a tiny bit unintuitive since "never" now means "never for images," but I can't think of a nicer way to handle it. Maybe some explanatory text?

PR #2741, opened 2026-05-18
Fix `CommentTreeTracker` bugs

### LINKED ISSUE #2739: New reply isn't inserted properly in some comment trees
https://lemmy.ml/post/47475102

Sjmarf:
There were two issues that affected `CommentPage` (the one with the "view all button"), specifically in cases where the focused comment had a depth of > 2.

In such cases, the intent is to load the parent of the focused comment, and all of its children. However, it was actually loading _all_ of that comment's parents, which was unintended.

Also, the check for determining whether a comment should be considered "top-level" in the current view incorrectly counted all children as being top-level, which messed up the tree structure. This caused "more replies" buttons to show where they shouldn't have done.

Closes #2739


EricBAndrews (approved):
LGTM! 🚀

PR #2756, opened 2026-05-29
Customisable user and instance swipe actions

### LINKED ISSUE #2617: Community, User, and Instance list gesture customization

Sjmarf:
Closes #2617

Settings -> Users
Settings -> Instances

---

The root settings page is getting rather long; I think we should find a better way to arrange this in future. Maybe all of the content types ("Posts", "Comments", etc) could go under a new settings page called "Content"?


EricBAndrews (approved):
LGTM! 🔥

PR #2759, opened 2026-05-29
Inbox lock

Sjmarf:
Added a lock to the `ApiClient` methods that affect the inbox count.

In theory, this should prevent very occasional notification count desyncs caused by multiple requests running at the same time. 

Additionally, this work may become necessary for #2710. Authenticating to Canvas may require sending and receiving DMs with a bot account, and we'll want to do this seamlessly in the background without the user's inbox count updating.

Using actors allows for nested locks acquisition. For instance, this won't cause a deadlock:

```swift
try await inboxLock.withLock {
    try await inboxLock.withLock {
        // ...
    }
}
```


EricBAndrews (approved):
LGTM! 💯

PR #2762, opened 2026-06-01
Separate error handling system for Lemmy and PieFed

### LINKED ISSUE #2563: Posts pinned to instance aren't shown as pinned on PieFed
Possibly missing API support - there's only a single `sticky` boolean. It's also possible that `sticky` is used for both community and instance pins. Not sure

### LINKED ISSUE #2599: Logging into PieFed with incorrect credentials causes "something went wrong" error
We should show "Password is incorrect" instead.

https://lemmy.ml/comment/24058118

Sjmarf:
`PieFedConnection` was incorrectly using the Lemmy error response schema. It now uses its own PieFed schema.

Fixes #2599

Also fixed instance-pinned posts not showing as pinned on PieFed (closes #2563)


EricBAndrews (approved):
LGTM! 🚀 

This is much nicer than the combined system :)

PR #2764, opened 2026-06-03
Shared media state

### LINKED ISSUE #1153: Un-blurred NSFW image becomes blurred again when opening in expanded view

EricBAndrews:
## Issues

- closes #1153 

## Description

Media control states are now tied to the media they control, so multiple instances of that media will use the same control state across the app. This is implemented by a global `MediaTracker`, which maps URLs to `MediaControlState`s.

## Implementation Notes

- `MediaTracker` stores weak references to the control states, and periodically removes any control states that have become nil from its map. This is done with a `WeakMediaControlState` rather than an `NSMapTable` because it needs to use URLs, a value type, as the keys.
- Multiple URLs can point to the same control state (e.g., in the case of a proxy bypass, media created with either the base or the bypassed URL should point to the same control state)
- To prevent multiple copies of the same animated media from playing over each other, `MediaControlState` stores a `mediaLockId`; this is unique per `MediaView`, the view acquires the lock when appearing, and animation is disabled if the media does not have the lock.


Mlem/App/Globals/Definitions/MediaTracker.swift
```diff
@@ -0,0 +1,51 @@
...
+    }
+}
+
+@Observable
+public class MediaTracker {
+
+    private var controlStates: [URL: WeakMediaControlState] = .init()
+    
+    private var lastCleaned: Date = .init()
+    private let cleanInterval: TimeInterval = .init(60)
```
Sjmarf: _nit_


```suggestion
    private let cleanInterval: TimeInterval = 60
```

Mlem/App/Globals/Definitions/MediaTracker.swift
```diff
@@ -0,0 +1,51 @@
...
+    public func controlState(for url: URL?, create: () -> MediaControlState) -> MediaControlState {
+        defer {
+            if Date().timeIntervalSince(lastCleaned) > cleanInterval {
+                for key in controlStates.keys where controlStates[key]?.value == nil {
+                    controlStates.removeValue(forKey: key)
+                }
+            }
+        }
+        
+        guard let url else { return create() }
```
Sjmarf, line 41: We're not storing the created `MediaControlState` in the dictionary here. Is that intended?

Mlem/App/Views/Root/ContentView.swift
```diff
@@ -35,6 +35,7 @@ struct ContentView: View {
     var appState: AppState { .main }
     var tabReselectTracker: TabReselectTracker { .main }
     var navigationModel: NavigationModel { .main }
+    var mediaTracker: MediaTracker { .main }
```
Sjmarf, line 38: Could we remove `.main` and instead initialise the `MediaTracker` here?

```suggestion
    @State var mediaTracker: MediaTracker = .init()
```

Personally I'd like to move away from having `.main` (where possible), and instead _require_ accessing our models through the environment. This makes writing SwiftUI previews (and testing, if we ever add that) much easier IMO.

If this isn't easily possible I'm also happy to leave it as-is

Sjmarf (commented):
Looks good! A few minor notes

Mlem/App/Globals/Definitions/MediaTracker.swift
Sjmarf, line 1: The blur state syncs correctly with the image shown in the reply sheet, but there's no "tap to unblur" option in the reply sheet. Maybe worth adding?

Mlem/App/Globals/Definitions/MediaTracker.swift
EricBAndrews: Yes, I've added a comment explaining this

Mlem/App/Views/Root/ContentView.swift
EricBAndrews: This is unfortunately rather awkward, as we have several places where we need to get the control state in initializers (e.g., when invoking `ImageViewer` or from Markdown).

I've added an `@available` check to discourage using `.main` unless absolutely necessary.

Sjmarf: Cool, that works 👍

Mlem/App/Globals/Definitions/MediaTracker.swift
EricBAndrews: Ooh, good point. I've made several changes to improve this behavior:
- `MediaView.largeImage` now sets `controlState.blurred` according to `shouldBlur`; this restores the existing behavior where images are unblurred when the reply sheet is opened
- `MediaView.largeImage` now takes in a set of overlays to apply in addition to the defaults; `.nsfw` is supplied according to the post's nsfw status in `CommentEditorView`
- `WebsitePreviewView` now also tracks `nsfw` state, and applies overlays appropriately

Sjmarf:
> MediaView.largeImage now sets controlState.blurred according to shouldBlur; this restores the existing behavior where images are unblurred when the reply sheet is opened

IMO we should keep the image blurred when the reply sheet is opened. I think in 99% of cases, the user will unblur the image before opening the reply sheet, anyway. I'm wary of unblurring images without explicit permission

Sjmarf (changes requested):

EricBAndrews:
Makes sense to me 👍

Sjmarf (approved):
LGTM 💯

PR #2766, opened 2026-06-05
Fix pinch to zoom breaking when device rotates

### LINKED ISSUE #2575: Image viewer handles device rotation weirdly
https://lemmy.ml/post/42719746

EricBAndrews:
## Issues

- closes #2575 

## Description

I feel like there should be a way to do this within the embedded UIKit framework, but the orientation APIs are _really_ bad.


Mlem/App/Utility/Extensions/Views/View Modifiers/View+DeviceRotation.swift
```diff
@@ -0,0 +1,28 @@
...
+//  Mlem
+//
+//  Created by Eric Andrews on 2026-06-04.
+//
+
+// Taken directly from https://www.hackingwithswift.com/quick-start/swiftui/how-to-detect-device-rotation
+
+import SwiftUI
+
+struct DeviceRotationViewModifier: ViewModifier {
```
Sjmarf: _nit_


```suggestion
private struct DeviceRotationViewModifier: ViewModifier {
```

Mlem/Packages/ComponentViews/Sources/ComponentViews/View+DeviceRotation.swift
Sjmarf, line 1: _nit_ Arguably this should go in `ComponentViews`

Mlem/App/Views/Shared/Images/Helpers/ZoomRecognizer/ZoomRecognizerCoordinator+Logic.swift
```diff
@@ -268,7 +268,8 @@ extension ZoomRecognizerCoordinator {
             return
         }
         
-        if bounds == nil, view.bounds != .zero {
+        if view.bounds != .zero && (bounds == nil || deviceOrientation != lastBoundsDeviceOrientation) {
+            lastBoundsDeviceOrientation = deviceOrientation
             bounds = .init(width: view.bounds.width, height: view.bounds.height)
```
Sjmarf: If the image is already zoomed in when the device is rotated, the old bounds still apply. We might want to call `initializeBounds` in `.onRotate`, or something like that?

Sjmarf (changes requested):
Nice work! Some comments

Mlem/App/Views/Shared/Images/Helpers/ZoomRecognizer/ZoomRecognizerCoordinator+Logic.swift
EricBAndrews: `initializeBounds` is triggered at the start of any gesture that depends on the bounds, so in that case this will trigger after rotation when the user begins a gesture.

Mlem/Packages/ComponentViews/Sources/ComponentViews/View+DeviceRotation.swift
EricBAndrews: Fair enough 👍

Mlem/App/Views/Shared/Images/Helpers/ZoomRecognizer/ZoomRecognizerCoordinator+Logic.swift
Sjmarf: If I zoom in, rotate the phone, then pan the image, it goes out of bounds and cannot be brought back. Here's a video:

https://github.com/user-attachments/assets/1726c9f4-679f-40cf-83b5-908a7f3fb7ab

If I rotate the phone when the image is _not_ already zoomed in, then the bounds prevent the image moving outside of the viewport.

EricBAndrews: Ah, I see. Nice catch! This actually requires calling `resetToBounds`--`initializeBounds` just sets the values for use in computations, but doesn't modify the view.

`updateUIView` now checks the view size to detect rotation, and handles both reinitializing the bounds on rotation and resetting the view to bounds if needed. This removes the `.onRotate` dependency and the bucket brigade `deviceRotation` :)

Sjmarf: Nice, that's a clean implementation :)

Sjmarf (approved):

Sjmarf:
LGTM 🔥🔥

PR #2767, opened 2026-06-06
Show warning when version is unsupported

### LINKED ISSUE #113: Explicitly check and "range" the supported Lemmy versions
While clamping the versions of lemmy should also be guided by a policy we should setup at a later time, we should also explicitly, in the app, look at the version number the server gives us, and check if we support it.

Sjmarf:
Closes #113

I will shortly be removing support for some of the older PieFed versions, so I thought now would be a good time to do this. The easiest way to test this is to change the minimum version in `Mlem/App/Utility/Extensions/Content Models/SiteSoftwareType+Extensions.swift`.

When an instance is unsupported, a warning is shown in the account switcher:

IMG_2520

When you open the account, this warning is shown:

IMG_2519

The warning will only be shown once for that account. If the instance is upgraded, and later falls outside of the support window for a _second_ time, the warning will be shown again. For testing purposes, you can use the "Reset Version Warning" button in Developer settings to see the warning a second time.

A similar warning is shown in the log in sheet, if the user chooses an unsupported instance:

IMG_2517

### DIFF
--- a/Mlem/App/Views/Shared/ListRow/AccountListRowBody.swift
+++ b/Mlem/App/Views/Shared/ListRow/AccountListRowBody.swift
@@ -80,7 +80,7 @@ struct AccountListRowBody: View {
 
         // The unreadCount check ensures that we are not showing stale data.
         // Once the unread count has been fetched, the version will have been updated.
-        if let software = account.siteSoftware, !software.isSupported, unreadCount != nil {
+        if let software = account.siteSoftware, !software.isSupported, responseTime != nil {
             var str = AttributedString(localized: "Unsupported")
             str.foregroundColor = ThemedColor.themedWarning.resolve(with: palette)
             output.append(str)
--- a/Mlem/App/Views/Shared/ListRow/AccountListRowBody.swift
+++ b/Mlem/App/Views/Shared/ListRow/AccountListRowBody.swift
@@ -66,18 +66,18 @@ struct AccountListRowBody: View {
         }
     }
     
-    var captionText: String? {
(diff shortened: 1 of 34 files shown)


Mlem/App/Views/Shared/ListRow/AccountListRowBody.swift
```diff
@@ -66,30 +68,47 @@ struct AccountListRowBody: View {
...
+                output.append(.init("@\(account.api.host)"))
             }
         }
-        if complications.contains(.lastUsed), let timeText {
-            if (account as? GuestAccount)?.isSaved ?? true {
-                output.append(timeText)
-            } else {
-                output.append(.init(localized: "Temporary"))
+
+        // The unreadCount check ensures that we are not showing stale data.
```
EricBAndrews: Not clear what the unreadCount check refers to here

EricBAndrews (approved):
Nice job, works great! 🚀

Mlem/App/Views/Shared/ListRow/AccountListRowBody.swift
Sjmarf: Good spot, this should say `responseTime` 👍

Because we store the account's instance version in a JSON file, we need to be careful that we're not still showing "unsupported" after the instance has upgraded, because of our cached data. The `responseTime != nil` check prevents this. Shortly before the `responseTime` is made available, Mlem updates the cached account version. This means it'll only show "unsupported" when we've fetched the account version recently.

PR #2768, opened 2026-06-06
More context menu customisation

### LINKED ISSUE #1642: Adopt full-size action sheet for posts/comments as we add more actions

Sjmarf:
- Context menus for communities, users and instances are now customisable
- You can now open the context menu editor from settings (See the "Posts" page, etc)

Closes #1642

I've removed some of the actions from the default context menus, but I'm not sure this is the right choice. Does this sacrifice usability? Would appreciate your opinion. One option is to put _all_ of the options in the default context menu, and hiding the "more" button by default (#2770). That would keep the same default behavior we have now, while still making it customisable.


EricBAndrews (approved):
LGTM! 💯 

Re: how many options to include, I think the set you've picked is a reasonable default. Personally I almost never use these menus so I don't have a strong opinion here--let's put it out to the TF and see if anybody says anything.

One minor note, happy to leave out of scope as it affects all of these menus: selecting "More..." while the keyboard is open causes the options sheet to be blocked by the keyboard, which can make the UX in the search tab clunky.

Sjmarf:
Good spot, have opened an issue: https://github.com/mlemgroup/mlem/issues/2773

Yeah sounds good, let's see what the TF thinks 👍

PR #2777, opened 2026-06-11
"Toggle notifications" action on PieFed

### LINKED ISSUE #1547: Subscribe to be notified of replies to a post/comment
Reddit allowed you to "subscribe" to a post/comment and be notified when someone replies to it. Lemmy doesn't have this yet. We could potentially provide this service if we had an Mlem backend server (#1313).

(Maybe to avoid confusion it would be better to call this "Watch" rather than "Subscribe")

### LINKED ISSUE #2615: Mute reply notifications for individual comments
This is a feature on PieFed and Lemmy 1.0

https://lemmy.ml/post/43641495

Sjmarf:
On PieFed, posts and comments now have a "toggle notifications" action.

Posts and comments with notifications enabled will send all replies to your inbox. By default, notifications are enabled for your own posts and comments.

For your own posts and comments, an icon is shown next to the ellipsis menu if you have disabled notifications.

For posts and comments that are _not_ yours, an icon is shown if you have enabled notifications.

LMK what you think of this way of displaying notification status. Happy to change it.

Closes #1547, closes #2615


EricBAndrews (approved):
LGTM! 🚀 

Re: notifications position, I think the only other viable place is the info stack, but I think the corner makes more sense--top right is very strongly associated with notifications. It is an incredibly privileged location, so we should be open to reconsidering what goes there if something more important comes along, but this design works for me for now 👍

Sjmarf:
Yeah that's a good point, we can totally move it in future 👍

Issue #2771, opened 2026-06-07
Lemmy 1.0: Use cursor pagination in search


Sjmarf:
Community and person search is currently broken past the first page. We need to switch to cursor pagination. See also #2559

Issue #2559, opened 2026-02-07
Lemmy 1.0: Use cursor pagination in Modlog


Sjmarf:
At the moment, the modlog loads but only the first page will be shown. We need to switch to using the paging system. This should be relatively easy.

PR #2780, opened 2026-06-12
`PageCursor` system

Sjmarf:
Lemmy has two paging systems:

- `page: Int`
- `cursor: String`

Previously, we hid this complexity inside `Fetcher`, which would choose which system to use. In this PR, I've moved the responsibility from `Fetcher` to `InstanceConnection`.

# Technical details

This is accomplished using a new `PageCursor` type. When you call a paged method of `InstanceConnection`, you pass in a `PageCursor`. Then, it returns a new `PageCursor` as part of the response. 

Internally, `PageCursor` stores either the page number or cursor string:

``` swift
public struct PageCursor: Hashable {
    internal enum CursorType: Hashable {
        case pageNumber(Int)
        case cursorString(String)
    }

    private let cursorType: CursorType

    // ...
}
```

Sorry for the large PR; there wasn't really a way to do this in smaller pieces :(

# Benefits and drawbacks

The downside: we can't start loading a feed at an arbitrary page number. We _must_ start at the first page, get the cursor for the second page, fetch the second page, and so on.

Lemmy 1.0 does away with the page number system entirely, which means we wouldn't have been able to take advantage of arbitrary starting locations anyway.

The benefits are:

- Fewer parameters on `InstanceConnection` methods
- Reduces complexity in some areas
- Makes it easier to switch to the cursor system in Lemmy 1.0 (#2559, #2771)

# Functional changes 

The only place we were using random access paging was in the logic for hiding read items when the inbox refreshes. Previously, hiding read items could be done without refreshing, but it now requires a refresh. I think this is an acceptable trade-off.


EricBAndrews (approved):
LGTM! 🔥

PR #2782, opened 2026-06-14
Fetch all instances for search

EricBAndrews:
Removes the default filters when fetching instances from the backend.


Sjmarf (approved):
LGTM 👍

PR #2783, opened 2026-06-15
Block list fixes

### LINKED ISSUE #2179: Lemmy 1.0: Viewing the instance block list is broken

### LINKED ISSUE #2486: View instance blocklist on PieFed
This is supported by the API, but is inconvenient to implement for us. 

Lemmy gives us all of the instance's information (avatar, version, etc), which we use to construct an `Instance1`. The block list UI relies on having an array of `Instance1`.

PieFed does /not/ provide this - it only provides the instance name and ID. This does not allow us to create an `Instance1`. We'd have to do some rewriting to support this.

Sjmarf:
- Fixed the block list being empty on Lemmy 1.0 (closes #2179)
- Enabled "Instances" tab in the block list on PieFed (closes #2486)

On both Lemmy 1.0 and PieFed, the block list endpoint does not return full instance information. To make support possible, I've changed the UI to only show the domain name:

IMG_2533

Previously, it looked like this:

IMG_2534


EricBAndrews (changes requested):
Nice work!

Re: reduced data payloads, can we get the `InstanceSummary` from `MlemStats` and show the full instance row?

Sjmarf:
Good idea, implemented 👍 It falls back to the simpler behavior if one of the instances can't be found in `MlemStats` for whatever reason.

EricBAndrews (approved):
LGTM! 💯

PR #2793, opened 2026-06-18
Translate comments

### LINKED ISSUE #928: Translate posts/comments
Add a 'translate' action to posts/comments. Apple's Translate API is private, so we'd have to use someone else's. Maybe Google MLKit? https://developers.google.com/ml-kit/language/translation/ios

Sjmarf:
There's a "translate" action in the action sheet, which is also configurable as a swipe action.

Closes #928. Translating posts is not implemented yet (I've opened #2792)

If the translation is supported, you get this:

https://github.com/user-attachments/assets/5de0e381-fc7f-4ee9-bcc6-2a52c941bf64

If the translation is unsupported, you get a toast:

Screenshot 2026-06-18 at 17 13 14

If the translation is supported, but you haven't downloaded the model for that language, you get this sheet:

Screenshot 2026-06-18 at 17 12 37

# Testing

Remember that you need to set your discussion languages to include other languages if you want them to appear in feeds and search.

Some useful examples:

Translation works: https://sh.itjust.works/comment/18282648
Translation produces toast: https://sh.itjust.works/comment/23705395
Translation prompts for download: https://sh.itjust.works/comment/24697922


EricBAndrews (changes requested):
Nice work, this feature is super cool!

I've noticed a couple bugs:
- Language install prompt only triggers from the ellipsis/context menu; it does not activate from the swipe action or the extended menu options
- The translation status icon is a little taller than the rest of that top section, which pushes the comment content down (easy to see when toggling one-line comments)
- Some comments (e.g., https://lemmy.world/comment/24363820) are being detected as English even though they are _very_ German--I'm not sure if this is a bug on our side or just a failing in the translation engine itself

On a design note, the translation icon and the notification icon both occupying the top right starts to crowd in on the username. I'm not entirely sure how to resolve this--moving notifications to the interaction bar would be consistent with other server-side statuses, but could be awkward if the user doesn't have notifications in the bar. I don't love adding text, but "Translated from Language. Show Original" in caption font might be a good resolution, and would add the context of the original language.

Sjmarf:
## Bugs

> Language install prompt only triggers from the ellipsis/context menu; it does not activate from the swipe action or the extended menu options

Can't repro for swipe actions; does it happen every time?

Fixed for the action sheet

> The translation status icon is a little taller than the rest of that top section, which pushes the comment content down (easy to see when toggling one-line comments)

Good spot, fixed 👍 

> Some comments (e.g., https://lemmy.world/comment/24363820) are being detected as English even though they are very German--I'm not sure if this is a bug on our side or just a failing in the translation engine itself

This is because the author has explicitly set the language ID to English. Mlem uses the language ID if available; otherwise, it falls back to using the language detector.

We can't just use the language detector all of the time, because it's not very accurate on short comments, especially if they contain slang.

We could try the language detector if translation using the language ID is unsupported, but this wouldn't work for cases where it _is_ supported.

Not sure what the best option is here. Maybe it's best to just leave it, and implement a way to manually specify the language in future.

## Translation indicator

> On a design note, the translation icon and the notification icon both occupying the top right starts to crowd in on the username.

Agreed, it's not ideal. However, it's very unlikely that both icons will be showing at once.

> I'm not entirely sure how to resolve this--moving notifications to the interaction bar would be consistent with other server-side statuses, but could be awkward if the user doesn't have notifications in the bar. 

We could change the save and notification status widgets to be completely hidden when inactive, rather than showing an outlined version. Then, we could add them both to the bar.

>  I don't love adding text, but "Translated from Language. Show Original" in caption font might be a good resolution, and would add the context of the original language.

Implemented, what do you think?

I have mixed feelings about this. With European languages, the translated text tends to take up the same number of lines as the original text. This makes the animation play nicely. With an extra line, the comment has to get larger every time. It _does_ show more information though, so maybe we should go with function > form.

I also removed the slight bounce from the transition, I think it looks better without it.

EricBAndrews:
I like the Translated From... text--given that there's already an animation on the text itself, the box height increase isn't obtrusive. Do we also need the translation icon in the top right if we have that text?

I can't repro the popup bug any more. That one might have just been an Xcode build quirk, it's working correctly for me now.

Re: language detection, we could fall back on the language detector if the language ID matches the system language; presumably users will not try to translate from their own language to their own language, so if we see that then it's probably mis-tagged.

Sjmarf:
> Do we also need the translation icon in the top right if we have that text?

Removed 👍 

> Re: language detection, we could fall back on the language detector if the language ID matches the system language; presumably users will not try to translate from their own language to their own language, so if we see that then it's probably mis-tagged.

Good idea, added.

EricBAndrews (approved):
LGTM! 💯

PR #2795, opened 2026-06-18
Fix occasional 404 error spam

### LINKED ISSUE #2788: Random chance of 404 error
Seems this is a recent regression. Probably error handling in the software multiplexer

Sjmarf:
Introduced recently, I'm guessing as a result of #2762. Can't repro right now so can't actually check if its fixed. Closes #2788


EricBAndrews (approved):
Nice, seems like a credible fix 👍

PR #2797, opened 2026-06-19
Better `UrlRequest` handling

### LINKED ISSUE #2070: Use a custom Nuke ImagePipelineDelegate to apply the User-Agent header to image requests

Sjmarf:
Removed `mlemUrlRequest`. The logic has been moved to the Nuke image pipeline instead. This means that `Rest` no longer needs to be a direct dependency of `Mlem`.

No functional changes.

Closes #2070 

## Testing

Tested that:

- `.gifv` media still loads (tested with https://64.media.tumblr.com/6caad9ad2481f0ea7fea725c59914bda/60a745b393eca1a8-13/s2048x3072/5ab7988dcc68b68a623a2361cd44404b8cd915e2.gifv)
- Lemmy.ml still works (they check the `User-Agent`)
- Image uploading still works on lemmy.ml


EricBAndrews (approved):
LGTM! 🚀

PR #2804, opened 2026-06-23
`Handle` system

Sjmarf:
Added two new structs to `MlemMiddleware`:

- `PersonHandle`: represents a `"@user@instance.com"` string
- `CommunityHandle`: represents a `"!comm@instance.com"` string

You can resolve a community or person using these types.

This is necessary because converting a handle to a URL isn't guaranteed to work for softwares other than Lemmy or PieFed. For instance, Mbin uses `/m/` rather than `/c/` for communities. Keeping the data in handle form - rather than converting to URL - allows us to support those cases.

I think this work may be necessary for the upcoming Canvas auth work (#2710), which is why it's happening now.

### DIFF
--- a/Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Community/CommunityStub.swift
+++ b/Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Community/CommunityStub.swift
@@ -10,7 +10,7 @@ import Foundation
 public struct CommunityStub: Hashable {
     public var api: ApiClient
 
-    private enum Content: Hashable {
+    private enum Reference: Hashable {
         case url(URL)
         case handle(CommunityHandle)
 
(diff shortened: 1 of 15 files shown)


Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Community/CommunityStub.swift
```diff
@@ -7,36 +7,56 @@
...
+
+        var baseUrl: URL {
+            switch self {
+            case let .url(url): url.removingPathComponents()
+            case let .handle(handle): handle.baseUrl()
+            }
+        }
     }
+
+    private let content: Content
```
EricBAndrews: *nit*

`content: Content` might not be the best name to use here--outside of views, that is used throughout the rest of the codebase to indicate a wrapper around multiple conceptually distinct entities, where here we are wrapping multiple representations of the same basic entity. Maybe `reference: Reference` makes more sense?

EricBAndrews (approved):
LGTM!

One minor nit

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Community/CommunityStub.swift
Sjmarf: Good point, fixed 👍

PR #2818, opened 2026-07-01
Maybe fix image upload error

Sjmarf:
The one Zedestrian was talking about on Matrix. Couldn't repro. I suspect it's something to do with using Lemmy for upload when it should be using PieFed.


EricBAndrews (approved):
LGTM 👍

PR #2820, opened 2026-07-03
Edit button in block list

### LINKED ISSUE #2760: Swipe actions should be overriden in blocklist

Sjmarf:
~The block list now has a single trailing swipe action: "unblock". It ignores the user's swipe action configuration (unless they've got swipe actions completely off, in which case it's disabled)~

The block list now has an "edit" button

Closes #2760


Sjmarf:
Switched from swipe actions to an "edit" button, as discussed

EricBAndrews (changes requested):
Nice work, the UI is very smooth.

The unblock button in the edit menu, however, doesn't actually unblock the instance--tested on sh.itjust.works and piefed.social.

Sjmarf:
Good catch, fixed :+1:

EricBAndrews (approved):
LGTM! 💯

PR #2826, opened 2026-07-03
Fix `NavigationLayer.replace` stuttering

### LINKED ISSUE #2533: NavigationLayer.replace does not animate
This was introduced in #2499.

Sjmarf:
Closes #2533. I couldn't repro it after this change, but can't be 100% sure.

Previously, we fed the`navigationDestination` with the `NavigationPage`:

``` swift
.navigationDestination(for: NavigationPage.self) {
    $0.view()
        .environment(\.isRootView, false)
}
```

I fixed the issue by introducing a new wrapper type, `NavigationFrame`, which is an `@Observable` class that wraps a `NavigationPage`, and has a unique `hashValue`.

``` swift
@Observable
class NavigationFrame: Hashable {
    var page: NavigationPage

    init(page: NavigationPage) {
        self.page = page
    }

    func hash(into hasher: inout Hasher) {
        hasher.combine(ObjectIdentifier(self))
    }

    public static func == (lhs: NavigationFrame, rhs: NavigationFrame) -> Bool {
        lhs === rhs
    }
}
```

`navigationDestination` now uses that wrapper type:

``` swift
.navigationDestination(for: NavigationFrame.self) {
    NavigationFrameView(frame: $0)
        .environment(\.isRootView, false)
}
```

When `NavigationLayer.replace` is called, it updates the `NavigationPage` _inside_ of the top frame, rather than the whole frame.

This shifts responsibility for view updates from the `navigationDestination` down into `NavigationFrameView`, which is a simple wrapper view.

``` swift
struct NavigationFrameView: View {
    let frame: NavigationFrame

    var body: some View {
        frame.page.view()
    }
}
```

One convenient consequence of this change is that `NavigationPage` no longer needs to be `Hashable`, so I've removed that conformance. This cleans up the `HashWrapper` mess.


EricBAndrews (approved):
LGTM! 🔥 

This is a clean fix 😁

PR #2827, opened 2026-07-04
Fix too many parents being shown for comment

### LINKED ISSUE #2800: "Show parent" doesn't work the first time

Sjmarf:
Closes #2800

I thought I had got this right in #2741, but it seems I was wrong. Example comment: 

https://lemmy.ml/comment/26240544

When you open a comment page, it should now always shown exactly one parent, and the "show parent" button should show parents one at a time.


EricBAndrews (approved):
LGTM! 🚀

PR #2829, opened 2026-07-04
Fix 2.4 -> 2.5 Settings decoding error

EricBAndrews:
When installing 2.5 after configuring settings with 2.4, InboxItemType decoder would throw an error instead of falling through the LegacyInboxItemType


Sjmarf (approved):
Nice 👍

PR #2830, opened 2026-07-05
Improve editor performance

### LINKED ISSUE #2727: Lag when writing a long post
### Requirements

### Description

As I typically keep my comments and posts short, I haven't encountered this problem before, but I noticed in writing a relatively long post that the app started to lag while I typed:

https://github.com/user-attachments/assets/51a291cf-b97f-4510-a414-2618d5d9222a

The recording itself didn't seem to quite capture the lag perfectly, but the choppiness of the keyboard inputs in the video is indicative of how it was; letters also take a moment to appear after being tapped on the keyboard. [Here's](https://sopuli.xyz/post/45351429) the post in question for reference.

### Mlem Version

Mlem 2.4.1 (775)

Sjmarf:
Closes #2727

Not perfect, but better than it was.

This is as good as it will get, I think. We _must_ recalculate the height of the view each time the character is typed, because we need to expand the height of the text editor within the editor sheet. This is, necessarily, an O(n) operation.

### DIFF
--- a/Mlem/App/Views/Pages/Editors/CommentEditor/CommentEditorView+Logic.swift
+++ b/Mlem/App/Views/Pages/Editors/CommentEditor/CommentEditorView+Logic.swift
@@ -99,14 +99,19 @@ extension CommentEditorView {
     }
     
     func checkSlurFilter(text: String) {
-        do {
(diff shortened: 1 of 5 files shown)


EricBAndrews (approved):
LGTM! 🔥

PR #2835, opened 2026-07-06
Fix comment animation

Sjmarf:
In #2827 I broke the comment loading animation. It has this "slide-down" appearance. I've fixed this


EricBAndrews (approved):
LGTM! 🚀

PR #2851, opened 2026-07-09
Silence network errors

### LINKED ISSUE #2839: Error toast for timeout error should be hidden

Sjmarf:
Closes #2839

If the app is moved to the background while a request is ongoing, these errors can appear. I kept getting these popups when returning to the app.

I've silenced those errors. They still appear in the error log and `ErrorView`.

The downside is that they won't appear in cases where we _did_ get a legitimate timeout, but I think it's worth that trade-off.


EricBAndrews (approved):
LGTM! 🚀 

I agree this is better overall

PR #2862, opened 2026-07-12
2.5

EricBAndrews:
Bump project version to 2.5.0


Sjmarf (approved):
LGTM 👍

Sjmarf:
We used just "2.4" last time - are we including ".0" intentionally?

EricBAndrews:
Good point--I don't think there's any difference either way but I've followed that pattern for consistency

PR #2866, opened 2026-07-17
Fix subscription list navigation in windowed mode on iPad

### LINKED ISSUE #2484: Cannot tap on feed in subscription list when app is windowed on iPad
Reported by @sunshinehot on Matrix

Sjmarf:
Closes #2484

We're now using `NavigationSplitView` on iPhone, too. This simplifies some code and may become necessary if Apple releases a foldable

Also added a nice transition when switching pages in split-view mode on iPad.


EricBAndrews (approved):
LGTM! 💯

PR #2872, opened 2026-07-19
"Mark Read on Visit" setting for inbox

### LINKED ISSUE #720: Mark read when reply is visited?

Sjmarf:
Closes #720

Settings -> Inbox

Enabled by default. Is this the right move?


EricBAndrews (changes requested):
Works great!

IMO this doesn't need to be configurable--posts already mark read on visit, it's honestly a little weird that inbox items haven't up to this point.

Sjmarf:
Good point, made it non-configurable 👍

EricBAndrews (approved):
LGTM! 💯

Issue #2290, opened 2025-09-15
Add "crosspost count" info stack widget


Sjmarf:
Pending API support from both Lemmy and PieFed. Requested here https://lemmy.ml/post/36180426

Sjmarf:
There's API support for this in PieFed 1.2.

PR #2873, opened 2026-07-19
Simplify `InteractionBarConfiguration` generics

Sjmarf:
No functional changes; refactor only.

Previously, `InteractionBarConfiguration` had three associated types:

``` swift
associatedtype ActionType: ActionTypeProviding
associatedtype CounterType: CounterTypeProviding
associatedtype ReadoutType: ReadoutTypeProviding
```

In this PR, I've made `CounterType` and `ReadoutType` into concrete types, rather than being associated types. Now, there is only one associated type.

``` swift
associatedtype ActionType: ActionTypeProviding
// CounterType and ReadoutType are now concrete
```

This is only possible because all three interaction bar configurations (post, comment, inbox) have identical options for counters and readouts.

# Downsides

It's now impossible to implement readouts or counters that are _specific_ to one of the three types (e.g. #2290).

# Motivation

This change will make it significantly easier to switch the interaction bar system over to the new action system, which I'll be doing soon.

Once the new action system is in, I'd like to make the readout/counter system more flexible again. I'm imagining a system similar to `ActionSeed`, to avoid the need for generics. But for now, I just want to get rid of the old action system ASAP lol


EricBAndrews (approved):
LGTM! 🚀

PR #2876, opened 2026-07-20
Infer image type from image data when file extension unavailable

### LINKED ISSUE #2875: Files downloaded from URLs without extension have no extension
https://sopuli.xyz/comment/24664561

Sjmarf:
Closes #2875

My original plan was to read the MIME type from the response, but we only get the `URLResponse` if the image was fetched directly. It's unavailable if the image was cached.


EricBAndrews (approved):
LGTM! 💯

PR #2886, opened 2026-07-25
More specific action labelling system

Sjmarf:
This is prep work for rewriting the interaction bar.

# The problem

In some cases, we want an action's label to describe a _transition between states_. For example:
```swift
// When unsaved
Label("Save", systemImage: "bookmark")
// When saved
Label("Unsave", systemImage: "bookmark.slash")
```

However, in the interaction bar, we want the label to describe the _current state_. For example:
```swift
// When unsaved
Label("Unsaved", systemImage: "bookmark")
// When saved
Label("Saved", systemImage: "bookmark.fill")
```

# The solution

Currently, `Action` only supports the "state transition" label and not the "current state" label. In this PR, you can now (optionally) specify two labels when creating an action's appearance:

```swift
static let saveAppearance: ActionAppearance = .init(
    currentStateLabel: .init("Unsaved", icon: .lemmy.saved.representingState(active: false)),
    stateTransitionLabel: .init("Save", icon: .lemmy.addSave),
    color: .themedSave
)
static let unsaveAppearance: ActionAppearance = .init(
    currentStateLabel: .init("Saved", icon: .lemmy.saved.representingState(active: true)),
    stateTransitionLabel: .init("Unsave", icon: .lemmy.removeSave),
    color: .themedSave,
    prominent: true
)
```

# Other changes

I renamed `ActionLabel` -> `ActionAppearance`. This frees up the term `ActionLabel` for use in the new struct. 

```swift
public struct ActionLabel {
    public let title: String
    public let icon: Icon
}
```

The old action system also had a struct called, `ActionAppearance`, which I've renamed to `LegacyActionAppearance`.

I added a `prominent: Bool` property to the new `ActionAppearance`, which will be used to control whether an interaction bar item appears "on". This is equivalent to `isOn` in the old system.

The icons for `VoteAction` and `SaveAction` have been changed. Previously, they reflected current state. They now reflect state transitions, in line with the other actions. `VoteAction` also no longer has squared icons.

<table>
  <tr>
    <td> Before  </td>
    <td> After  </td>
   </tr> 
   <tr>
      <td>
  </td>
      <td>
  </td>
  </tr>
</table>

### DIFF
--- a/Mlem.xcodeproj/project.pbxproj
+++ b/Mlem.xcodeproj/project.pbxproj
@@ -279,8 +279,8 @@
 		0397D4932C6CE87E002C6CDC /* PostEditorTargetView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0397D4922C6CE87E002C6CDC /* PostEditorTargetView.swift */; };
 		0397D49A2C6EA6EE002C6CDC /* InteractionBarEditorView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0397D4992C6EA6EE002C6CDC /* InteractionBarEditorView.swift */; };
 		0397D49C2C6EA73C002C6CDC /* InteractionBarConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0397D49B2C6EA73C002C6CDC /* InteractionBarConfiguration.swift */; };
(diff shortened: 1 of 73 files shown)


EricBAndrews (approved):
LGTM! 💯

Issue #2897, opened 2026-08-03
Use new action system for tile layout ellispsis menus


Sjmarf:
Both posts and comments should use the new system

Issue #2895, opened 2026-08-03
Haptics are delayed for certain interaction bar items


Sjmarf:
E.g. Pin, lock.

Related: #2893

Issue #2894, opened 2026-08-03
Certain toasts should not occur when invoked from interaction bar


Sjmarf:
When "toggle notifications" is an interaction bar item, we shouldn't show the toast.

However, we SHOULD show the toast in cases where the action removes the post from the feed, e.g. hiding or blocking.

Issue #2893, opened 2026-08-03
Interaction bar moderator actions should show `ProgressView` while completing


Sjmarf:
Blocked by #2840

PR #2898, opened 2026-08-03
New interaction bar system

Sjmarf:
Progress towards #1923

The interaction bar system now uses the new `Action`.

# Status of migration process

Still remaining:

- The ellipsis menu in tile posts still uses the old `Action`. #2897
- Counters still rely on `LegacyActionAppearance`. #2896

The interaction bar was the last hard bit; I should be able to get those two issues done pretty quickly. Then it'll be done! 🎉

# Regressions

Some minor regressions in functionality.

- Moderator actions no longer show `ProgressView` while the action is ongoing. I intend to fix this before 2.6 #2893
- Alternate mod mail layouts no longer have separate default configurations. I'm not sure this is worth fixing. The alternate layouts are off by default, anyway, and the fix could end up messy. Thoughts?

# Other functional changes

- There are now many more actions available as interaction bar items
- The "mark read" interaction bar item icon is now a closed envelope in both "on" and "off" states. I think this is less confusing
- Moderation actions are now greyed out rather than hidden when unavailable. This was accidental, but I don't mind it either way. Can raise an issue to fix it, if you'd prefer they disappear?
- The "Save" action now uses the `.success` haptic in the new action system, rather than `.lightSuccess` as it did previously. This makes it consistent with the old action system, which used `.success`.

# Known issues

- "Toggle notifications" shows a toast. It probably shouldn't.  #2894
- Delete button is green. Tricky to fix. Leaving this for now, under the assumption that no-one is putting the delete button in the interaction bar. #2903
- Haptics are delayed for certain actions (pin, lock) #2895
- Both "purge" and "purge user" use the same icon. Not sure it's worth fixing this.


EricBAndrews (changes requested):
Works well, this is nice and clean!

Some minor notes:
- Resolve report is an option for post and comment bars, unless I'm mistaken that action is only defined for reports
- There are a _lot_ of options in the interaction bar customizer. I think we should group them into sections, something like:
  - Common: Upvote, Downvote, Save, Reply, Share, Crosspost, Hide
  - Counters: current counters section
  - Moderation: User Modlog, Pin, Lock, View Votes, Remove, Ban, Purge, Purge User
  - Advanced: Select Text, Create Image, Details, Toggle Notifications, Report, Edit, Delete, Block, Copy Username, Send Message, Add NSFW Tag
- I'm OK removing the alternate mod layouts, though if users want them back I think we should remain open to that

Sjmarf:
> Resolve report is an option for post and comment bars, unless I'm mistaken that action is only defined for reports

Good spot that it's only defined for reports - it wasn't showing up in the interaction bar. I've changed `ResolveAction` to instead depend on `ReportableProviding`, and pull the `reportContext` from the environment. The action now appears and is usable in the mod mail. 

Because post/comment interaction bars and the mod mail interaction bars may share the same configuration, I think it's best to keep it as an option. Removing it would be more trouble than it's worth, I think.

Side note: in future it may be possible to show reports inline in post/comment feeds. The PieFed web frontend has this I believe, though the API doesn't.

> There are a lot of options in the interaction bar customizer. I think we should group them into sections

Good point. I've given them the same grouping that we have in the action sheet, for consistency and because we have that information already. 

IMG_2649

This isn't quite as well-organised as your proposal, but is easier to implement, and I think there's something to be said for using the same ordering as the action sheet. Can implement the order you mentioned if you disagree.

> I'm OK removing the alternate mod layouts, though if users want them back I think we should remain open to that

To be clear, the alternate layouts are still available, but they no longer have separate _default_ configurations.

EricBAndrews (approved):
LGTM! 💯

PR #2902, opened 2026-08-07
Use quick swipes in subscription list

### LINKED ISSUE #2671: Creating image from post fails to load thumbnail sometimes
### Requirements

### Description

When trying to create an image of this post, the thumbnail from the article is displayed. But when you save or share the image it is broken. https://slrpnk.net/post/35922874

### Mlem Version

_No response_

Sjmarf:
The subscription list now used `.quickSwipes` rather than iOS swipe actions. This also makes them customisable (closes #2671)

This is more consistent internally, at the expense of being less consistent with iOS in general. I think this is the better choice, though. If you disagree, we can go back to using regularr swipe actions and just make the iOS swipe actions customisable


EricBAndrews (approved):
LGTM!

What does the `originalMinListRowHeight`/`defaultMinListRowHeight` do?

Sjmarf:
@EricBAndrews `defaultMinListRowHeight` is a SwiftUI env value that controls the minimum height of each row. SwiftUI applies the equivalent of `.frame(minHeight: defaultMinListRowHeight)` to each item. Because our rows are smaller than this value, this causes some "padding" to be added at the top and bottom of each row. This needed to be removed so that the `.quickSwipes` can extend all the way up and down to the divider.

However, we then need to re-apply the _same_ minimum height inside of the `.quickSwipes` content view, otherwise the rows are too short. To accomplish this, I save the value of `defaultMinListRowHeight` _before_ setting it to 0, and put that original value in `originalMinListRowHeight`. Inside of the `.quickSwipes` content, `originalMinListRowHeight` is read and `.frame(minHeight: originalMinListRowHeight)` is applied

I could have just hard-coded the min height inside the `.quickSwipes` label, but I thought this was more robust in-case the padding is different on iPad or something.

PR #2903, opened 2026-08-07
Silence timeout error (again)

Sjmarf:
It can be nested inside of `ApiClientError.networking()`, which escaped the check


EricBAndrews (approved):
Nice 👍

PR #2910, opened 2026-08-11
Parallelize initial MultiFetcher fetch

### LINKED ISSUE #2808: Mod mail loads slowly
Upwards of 60 seconds on some accounts. On Voyager it's much faster on those same accounts. Have a look at what requests Voyager is making

EricBAndrews:
## Issues

- closes #2808 

## Description

`MultiFetcher.fetch` now triggers all empty child sources to load their first page in parallel. This should reduce the loading time of the first aggregate page from the sum of all source loading times to the maximum individual source loading time.

### DIFF
--- a/Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/FeedLoaders/Generics/MultiFetcher.swift
+++ b/Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/FeedLoaders/Generics/MultiFetcher.swift
@@ -27,6 +27,7 @@ class MultiFetcher<Item: FeedLoadable>: Fetcher<Item> {
                     do {
                         try await source.loadMoreItems()
                     } catch {
+                        self.log.error("\(error.localizedDescription)")
                         // noop - if the error recurs it will be caught by normal error handling below.
                         // prefetch is not required to succeed for loading to work
                     }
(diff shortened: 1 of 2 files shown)


Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/FeedLoaders/Generics/MultiFetcher.swift
```diff
@@ -20,6 +20,13 @@ class MultiFetcher<Item: FeedLoadable>: Fetcher<Item> {
     }
     
     override func fetch() async throws -> LoadingResponse<Item> {
+        // preemptively trigger loading on all child feed loaders if they have not loaded anything yet
+        for source in sources where source.fetcher.location == .start {
+            Task {
+                try await source.loadMoreItems()
```
Sjmarf: Do we want to do anything with this error? Maybe await all of the tasks and unwrap the errors?

EricBAndrews: I think errors here should be expressly ignored. If something goes wrong in the prefetch, it shouldn't interrupt loading--the call will just be retried when that source comes up in the merge. At that point, either:
1. The error recurs and is handled by the existing flow
2. The load succeeds, which suggests the error was a one-off

Either way, errors here should not interrupt loading, since this operation succeeding is not required for the loading to work

Sjmarf: Makes sense 👍

Sjmarf (approved):
LGTM 👍

Issue #1923, opened 2025-03-29
Better architecture for storing the interaction bar and swipe actions


Sjmarf:
Our current architecture has some issues. IMO we should rewrite large parts of it. Haven't put much thought into potential solutions just yet

PR #2913, opened 2026-08-12
Use new action system for tile posts

### LINKED ISSUE #2897: Use new action system for tile layout ellispsis menus

### LINKED ISSUE #2907: "Resolve" should not show in context menu
Visibility should be set to `.hidden`

Sjmarf:
Closes #2897. This means we can remove some more of the old action system

Also fixes #2907

Remaining issues are tracked as sub-tasks of #1923


EricBAndrews (approved):
LGTM! 💯

Issue #2918, opened 2026-08-12
Use new action system for `MessageView` swipe actions


Sjmarf:

Sjmarf:
Closed in #2924

PR #2924, opened 2026-08-13
Unified message

### LINKED ISSUE #2696: Unified Models: Message
### Requirements

### Description

_No response_

EricBAndrews:
## Issues

- closes #2696 
- progress towards #2918 

## Description

This PR replaces the old `Message1Providing` and `Message2Providing` with a unified `Message` model.

## Implementation Notes

`MessageView.quickSwipes` now uses the new action system. I have simply implemented the old functionality (just a single trailing swipe to toggle read) using the new system; modifying that functionality is OOS here


Mlem/App/Actions/EditAction.swift
```diff
@@ -68,15 +68,16 @@ extension EditAction {
...
-            if let message = message as? any Message2Providing {
-                if let editMessage = environment.editMessage {
-                    editMessage(message.message2)
+            if let editMessage = environment.editMessage {
+                editMessage(message)
+            } else {
+                if message.isOwnMessage, let recipient = message.recipient.value {
+                    environment.navigation?.push(.messageFeed(recipient, focusTextField: true, editing: message))
+                } else if !message.isOwnMessage, let creator = message.creator.value {
+                    environment.navigation?.push(.messageFeed(creator, focusTextField: true, editing: message))
```
Sjmarf: I don't think it's possible to edit others' messages, even as an admin


```suggestion
```

Mlem/App/Actions/ReplyAction.swift
```diff
@@ -77,7 +77,11 @@ extension ReplyAction {
         case let .comment(comment):
             navigation.openSheet(.createComment(.comment(comment), commentTreeTracker: environment.commentTreeTracker))
         case let .message(message):
-            navigation.push(.messageFeed(message.creator, focusTextField: true))
+            guard let creator = message.creator.value_ else {
+                assertionFailure("Reply called on Message1")
+                return
+            }
+            navigation.push(.messageFeed(creator, focusTextField: true))
         }
```
Sjmarf: Will this work for your own messages? Looks like it might open a DM with yourself

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Message/Message.swift
Sjmarf, line 1: Found a bug:

- Go to inbox
- Mark all as read
- Tap on one of **your own** messages, to open the message feed
- Go back to inbox
- The "mark all as read" button shows

We've had a similar bug in the past; iirc `read` is always `false` for your own messages. We should probably change it to `true` during conversion from `LemmyPrivateMessage` -> `Message1Snapshot`

Sjmarf (commented):
Nice work! 😄 A few comments

Mlem/App/Actions/EditAction.swift
EricBAndrews: Hmm, good point. This is technically the logic that was previously implemented, but there's no UI to hit this path. Removed.

Mlem/App/Actions/ReplyAction.swift
EricBAndrews: Nice catch, fixed

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/Message/Message.swift
EricBAndrews: Nice catch! I'm not actually sure why this started manifesting here--the root cause was actually that marking those messages read on navigation was sending the unread count below 0. I've added a bunch of assertions to catch this if it happens again, and made `updateRead` noop if the existing state matches the desired.

Sjmarf: Cool, looks good!

Sjmarf (approved):

PR #2926, opened 2026-08-13
Improve haptic response time

### LINKED ISSUE #2895: Haptics are delayed for certain interaction bar items

EricBAndrews:
## Issues

- closes #2895 

## Description

This PR unifies how haptics are handled within the new action system:
1. Haptics are now _always_ called optimistically; i.e., they fire before the call actually goes through. 
2. Calling haptics is now the action's responsibility, and are called in `execute` (or a helper called therein). I have started to remove haptics from the `toggle` functions that previously took feedback arguments, though removing them outright requires some legacy action code to be removed first

## Implementation Notes

There is still, regrettably, a slight delay in haptic feedback for any action that has a dialog. As far as I can tell this delay is imposed by the dialog itself, and this is as good as it can get without removing the dialog.


Mlem/App/Utility/Extensions/Content Models/Interactable/InteractableProviding+Toggles.swift
```diff
@@ -15,12 +15,9 @@ import Theming
...
+    var toggleVote: ((ScoringOperation) -> Void)? {
         if let updateVote, let votes = votes.value {
-            return { operation, feedback in
-                if feedback.contains(.haptic) {
-                    HapticManager.main.play(haptic: .lightSuccess, tier: .low)
-                }
+            return { operation in
                 updateVote(operation == votes.myVote ? .none : operation)
                 self.inboxItem?.updateRead(true)
             }
```
Sjmarf, line 23: OOS but this should probably be pushed down into MlemMiddleware. #2927

Sjmarf (approved):
Nice! 🙌

PR #2930, opened 2026-08-14
Fix code block highlighting breaking when moving to background

### LINKED ISSUE #2929: Syntax highlighting breaks when app moves to background
https://lemmy.ml/post/46635399

Sjmarf:
Closes #2929

See the [LemmyMarkdownUI PR](https://github.com/mlemgroup/LemmyMarkdownUI/pull/65).

Note that we're now pinning to a commit rather than a version number. I don't think it's worth bothering with versioning. Maybe we should move that package into the Mlem repo at some point?


EricBAndrews (approved):
LGTM! 💯 

Moving this into the Mlem repo makes sense to me 👍

PR #2935, opened 2026-08-16
bump version to 2.5.1

EricBAndrews:


Sjmarf (commented):
I believe we also need to bump the share extension version?

Sjmarf (approved):
LGTM

PR #2938, opened 2026-08-19
Fix image sharing regression

Sjmarf:
Fix regression introduced in #2876. When sharing an image, the file name would be /just/ the file extension, rather than the last path component as was intended


EricBAndrews (approved):
LGTM 👍

PR #2939, opened 2026-08-19
Add "mark read on visit" toggle

Sjmarf:


Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/UnreadCount.swift
```diff
@@ -38,12 +38,16 @@ public final class UnreadCount {
...
     var verifiedCount: Count = .init()
+
+    /// A delta to apply to verifiedCount for state-faking. For example,
+    /// if the state-faked count contains one less personal inbox item than
+    /// the verified count, this will be (personal: -1, moderation: 0).
     var unverifiedCount: Count = .init()
     
     public var personal: Int {
         assert(verifiedCount.personal >= 0, "Invalid verifiedCount.personal \(verifiedCount.personal)")
-        assert(unverifiedCount.personal >= 0, "Invalid unverifiedCount.personal \(unverifiedCount.personal)")
```
Sjmarf, line 46: This was going off when an item was marked as read. We don't need this assertion (see comment above)

EricBAndrews: Fair, though we have had problems in the past with negative counts so I don't want to entirely remove the guards. The critical thing is that the actual return stays positive, so I think this would capture the intended behavior without tripping false positives:

```
assert(verifiedCount.personal + unverifiedCount.personal >= 0, "Invalid personal count \(verifiedCount.personal + unverifiedCount.personal)")
```

EricBAndrews (changes requested):
LGTM!

One nit re: assertions

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/Content Models/UnreadCount.swift
Sjmarf: Nice, that's much better 👍

EricBAndrews (approved):
LGTM! 💯

PR #2945, opened 2026-08-26
Refresh inbox when mark all read used while unread only filter is on

### LINKED ISSUE #2941: Refresh inbox when "mark read" tapped if "unread only" filter is on

EricBAndrews:
## Issues

- closes #2941


Sjmarf (approved):
LGTM!

PR #2947, opened 2026-08-30
Create actions from notification/report for message swipes

### LINKED ISSUE #2925: `MessageView` should use same swipe action configuration as replies
At the moment, there's a single trailing "mark read" swipe action. It does not follow the inbox swipe configuration

aqylbermeshtech:
## Issues

- closes #2925

## Description

`MessageView` hardcoded a single trailing "mark read" swipe instead of reading the inbox swipe configuration, so private message rows ignored Settings → Inbox → Swipe Actions. It now uses the same `ReplyBarConfiguration` (`interactionBar_reply`) that `ReplyView` uses.

## Implementation Notes

Adds a `quickSwipes(notification:message:report:configuration:)` overload alongside the existing one in `ContextMenu+InboxNotification.swift`, plus a private `ActionSeed.createAction(notification:message:report:)` helper that resolves each seed against notification → report → message — the same order `contextMenu(notification:message:report:)` already uses in that file.

Seeds that don't apply to messages (`upvote`, `downvote`, `save`) return `nil` and drop out via `compactMap`, so only the applicable configured actions render.

With the default config, message rows now show **Reply** where they previously showed **Mark Read**. Agreed as the desired behavior in #2925.


EricBAndrews (approved):
LGTM! 💯

PR #2948, opened 2026-09-04
Update UITextView+Extensions.swift

### LINKED ISSUE #2942: "Insert spoiler" button should add a new line
If the cursor is at the end of a non-empty line and you tap the "spoiler" button, it adds the spoiler tag on the current line, which is invalid. It should make a new line for it instead in this case.

aqylbermeshtech:
## Issues

- closes #2942

## Description

Tapping "Insert spoiler" while the cursor sits at the end of a non-empty line inserted the tag onto that same line, producing invalid markup:

## Implementation Notes

The change is in `insertBlock(prefix:suffix:)` in `UITextView+Extensions.swift`, which also backs the code block button, so ``` fences get the same fix.

- Added an isAtStartOfLine(_:)`helper that checks the single character behind a position; the start of the document counts as the start of a line.
- When not at the start of a line, a \n is prepended -mirroring the existing atEnd check that appends one on the trailing side.
- The caret offset is bumped by one in that case, so the cursor still lands on the empty content line inside the block.

Unchanged when the cursor is already at the start of a line or document (no stray blank line), and with a non-empty selection the selected text still ends up inside the block.


Sjmarf (approved):
Nice! Works well 😄

PR #2950, opened 2026-09-04
Language picker

### LINKED ISSUE #1200: Language picker in post/comment composer
We should only show this option when the user's account has >1 languages tied to it. We could even [auto-detect the language used in the text](https://developer.apple.com/documentation/naturallanguage/identifying-the-language-in-text).

Sjmarf:
Added a language picker to the post and comment editors. The picker is only shown if you have more than one discussion language configured.

Closes #1200

Not available on PieFed.


Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/ApiClient+Instance.swift
```diff
@@ -17,6 +20,16 @@ public extension ApiClient {
         }.result
         return model
     }
+
+    func getLanguageId(language: Locale.Language?) async throws -> Int {
+        guard let language else { return 0 } // 0 is the ID for "undefined"
```
EricBAndrews: *nit*

This might be more readable/self-documenting if we extend `Int` to have a static `.unidentifiedLanguageId`

EricBAndrews (approved):
LGTM! 💯

Mlem/Packages/MlemMiddleware/Sources/MlemMiddleware/API Client/ApiClient+Instance.swift
Sjmarf: Good idea 👍

Issue #2952, opened 2026-09-07
Test image deletion on Lemmy 1.0


Sjmarf:
It looks like they might have changed how it works (see the new `LemmyDeleteImageParams` struct)

Issue #2409, opened 2025-11-05
Delete images on PieFed


Sjmarf:
On Lemmy we delete images that were uploaded in the post editor but not used; we should support this for PieFed too

PR #2953, opened 2026-09-07
Refactor image uploading

Sjmarf:
No functional changes.

Moved some of the image uploading complexity to `RestClient`. This will make some upcoming work (#2409, #2334, #2952) easier.

Also, we now include the correct MIME type for images. Previously, we always used `image/png` on PieFed (interesting that this didn't cause any issues) and `application/octet-stream` on Lemmy (causing the backend to infer the content type from the data). Now, the correct MIME type is used if available, and we fall back to `application/octet-stream` otherwise.

Tested:
- Uploading PNG from photos works on Lemmy & PieFed
- Uploading HEIC from photos works on Lemmy & PieFed
- Uploading from clipboard works on Lemmy & PieFed
- Works on lemmy.ml (which has `User-Agent` check)
- Lemmy "Entity too large" error decoded correctly
- PieFed image upload errors decoded correctly

Haven't tested on Lemmy 1.0; it was already broken anyway.


EricBAndrews (approved):
LGTM! 🚀

PR #2957, opened 2026-09-09
Mock read status when unauthenticated

EricBAndrews:
This is a small quality-of-life change for unauthenticated browsing. Posts now appear read when interacted with from an unauthenticated client. This is a slight lie, but I think makes using the app unauthenticated a little easier. Happy to workshop/discard/etc if there is design pushback.

Also prevents "Not logged in" errors from certain post interactions.


Sjmarf (approved):
Good idea, this is cool! 👌

PR #2960, opened 2026-09-10
Minor action system cleanup

### LINKED ISSUE #2917: Don't use old `upvoteAction` in `upvoteCounter` guard
This would allow us to remove `upvoteAction`

```swift
    func upvoteCounter(appState: AppState) -> Counter? {
        guard let votes = votes.value,
              let upvoteAction = upvoteAction(appState: appState, feedback: [.haptic]) else { return nil }
        return .init(
            value: votes.upvotes,
            leadingAction: .upvote,
            trailingAction: nil
        )
    }
```

The same applies for `downvoteAction`

### LINKED ISSUE #2919: Remove `Report.toggleResolved`
Calling haptics from inside model extensions is no longer idiomatic, and should be avoided

Sjmarf:
Closes #2919
Closes #2917

On instances with upvotes/downvotes disabled, the buttons now appear greyed out rather than disappearing.

Screenshot 2026-09-10 at 15 52 40


EricBAndrews (approved):
LGTM! 💯

PR #2964, opened 2026-09-13
Fix PieFed link handling when instance not in `MlemStats`

Sjmarf:
PieFed has the following link formats:

- https://piefed.social/c/politics/p/1385905/will-the-supreme-court-hand-government-contractors-blanket-immunity
- https://piefed.social/post/2376868#comment_12898985

These would not be handled when the instance was not in `MlemStats`. Today, `piefed.social` was not in `MlemStats`.


EricBAndrews (approved):
LGTM! 💯

PR #2972, opened 2026-09-17
Fix post editing failing on PieFed if post contains alt text

Sjmarf:
We need a better error handling system for this at some point. These errors don't show up in the log


EricBAndrews (approved):
LGTM! 🚀

